PR #10, opened 2018-10-28
Network approach to BP algorithm

elshize:


include/recursive_graph_bisection.hpp
```diff
@@ -18,6 +19,83 @@
 namespace ds2i {
 const Log2<4096> log2;
 
+#ifndef NDEBUG
+const static bool Debug = true;
+#else
+const static bool Debug = false;
+#endif
```
amallia: Really needed?

include/recursive_graph_bisection.hpp
```diff
@@ -18,6 +19,83 @@
...
+} GreaterOrEqual;
+
+template <typename T1, typename T2, typename BinaryOperator>
+DS2I_ALWAYSINLINE void expects(const T1 &v1, BinaryOperator op, const T2 &v2) {
+    if constexpr (Debug) {
+        if (not op(v1, v2)) {
+            op.error(std::cerr, v1, v2);
+            std::cerr << '\n';
+            std::abort();
+        }
```
amallia: An assert will produce the same result here.

include/recursive_graph_bisection.hpp
```diff
@@ -0,0 +1,408 @@
...
+#include "forward_index.hpp"
+#include "util/index_build_utils.hpp"
+#include "util/log.hpp"
+#include "util/progress.hpp"
+#include "util/single_init_vector.hpp"
+
+namespace ds2i {
+const Log2<4096> log2;
+
+#ifndef NDEBUG
```
elshize: This is somewhat temporary. I needed some assertion at hoc, didn't want to put it anywhere else, since I know you've looked into assertions and maybe already planned to do something else.

amallia (changes requested):

include/recursive_graph_bisection.hpp
elshize: See comment

include/recursive_graph_bisection.hpp
amallia: Oh ok, read only now. Just commented about that. Thanks for clarifying

include/recursive_graph_bisection.hpp
elshize: An assert will not print the values.

amallia: Shall we try https://github.com/amallia/better_assert ?

include/recursive_graph_bisection.hpp
elshize: Actually, let me remove all of this for now, we should add it separately globally somewhere. Just needed something to debug and started playing around.

src/recursive_graph_bisection.cpp
```diff
@@ -24,10 +42,13 @@ int main(int argc, char const *argv[]) {
     app.add_option("--store-fwdidx", output_fwd, "Output basename (forward index)");
     app.add_option("--fwdidx", input_fwd, "Use this forward index");
     app.add_option("-m,--min-len", min_len, "Minimum list threshold");
-    app.add_option("-d,--depth", depth, "Recursion depth");
+    auto optdepth = app.add_option("-d,--depth", depth, "Recursion depth");
     app.add_option("-t,--threads", threads, "Thread count");
     app.add_option("--prelim", prelim, "Precomputing limit");
+    auto optconf = app.add_option("--config", config_file, "Node configuration file");
     app.add_flag("--nogb", nogb, "No VarIntGB compression in forward index");
+    app.add_flag("-p,--print", print, "Print ordering to standard output");
```
amallia, line 71: Why does ordering need to be printed if a `.mapping` file is generated?

amallia (approved):
LGTM

src/recursive_graph_bisection.cpp
elshize: Frankly, I didn't notice (or remember) that it it was printing out the mapping. Which makes me think that maybe that should be a separate function called from the source file. Just something to think about.

Anyway, I think it still makes sense to print it out as text, might be handy for analysis or debugging. But I can remove it if it bothers you.

amallia: We can leave it. I don't mind. I was just wondering

include/recursive_graph_bisection.hpp
elshize: Sure, why not.

elshize:
How come `constexpr` fails now on Travis? Did it work before?

amallia:
I got rid of C++17, but feel free to re-add it in CMakeLists.txt

elshize:
Oh, so you never ran the BP on Travis before? I thought you had. Anyway, I'll deal with it.

amallia:
Yes, I did. I just removed all the `if constexpr` as I don't need it anymore.

elshize:
So where are these changes? Did you work on another branch? Coz nothing was on `master` and this branch had `constexpr`s before my changes.

amallia:
I will update the compiler in travis. It is using GCC-6, so there is not support for `if constexpr`. My bad!

https://github.com/pisa-engine/pisa/pull/11

amallia:
Updated to use GCC 7

elshize:
There seems to be a problem with `QMX` library that a test is added but no binary and `ctest` fails with "Not Run". We should fix that before merging this.

amallia:
Where is `qmxTest` coming from? I never created this test AFAIK. I will give it a look tonight.

EDIT: coming from here https://github.com/amallia/QMX/blob/1c5c7385de244589f04c30bfe9ec12de0c99c15e/test/qmxTest.cpp

amallia:
Re-running CI in order to squash and merge

amallia:
`C++17` is blocked by https://github.com/stxxl/stxxl/pull/94

elshize:
> `C++17` is blocked by [stxxl/stxxl#94](https://github.com/stxxl/stxxl/pull/94)

What do you use from `stxxl`? I guess we can for now use c++14 globally and only set c++17 wherever we need.

Another question: how come it compiles? I set c++17 in another branch for the entire project, and I found no issues.

amallia:
Good point @elshize! Added C++17 only for `recursive_graph_bisection` and set C++14 globally.

The problem is that Clang has implemented C++17 and it has removed `random_shuffle` and other stuff, according to the new standard. GCC 7 (used by Travis) does not have this problem, but it will show up soon.

elshize:
Oh, it's a clang-specific issue, got it.

PR #14, opened 2018-12-20
MaxScore TAAT and blocked accumulators

elshize:
Here's what should be done in this PR:

- [x] Exhaustive TAAT
  - [x] Lazy initialization
  - [x] Prefetching
  - [ ] ~SIMD?~
  - [x] Block-wise aggregation
- [x] TAAT MaxScore


elshize:
@amallia I didn't have as much time as I thought I would, so I only managed to start some coding.

Is `queries` program supposed to be used for benchmarking?

amallia:
From my understanding `nodiscard` is used to enforce the programmer to use the returned value of a function. For example, this can be useful when an error is returned and the error checking has to be enforced. In this context I believe `nodiscard` can be avoided.

include/query/algorithm/exhaustive_taat_query.hpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+} // namespace query
+
+template <typename Index, typename WandType, size_t accumulator_block_size = 0>
+struct exhaustive_taat_query {
+    exhaustive_taat_query(Index const &index, WandType const &wdata, uint64_t k)
+        : m_index(index), m_wdata(wdata), m_topk(k), m_accumulators(index.num_docs())
+    {
+        static_assert(accumulator_block_size >= 0, "must be non-negative");
```
amallia: Always true since it is size_t, right?

include/query/algorithm/exhaustive_taat_query.hpp
```diff
@@ -0,0 +1,138 @@
...
+            m_acc_block_count = 1;
+            m_accumulators_max.push_back(std::numeric_limits<float>::max());
+        }
+    }
+
+    uint64_t operator()(term_id_vec terms) {
+        auto cws = query::cursors_with_scores(m_index, m_wdata, terms);
+        return taat(std::move(cws.first), std::move(cws.second));
+    }
+    using score_function_type = std::function<float(uint64_t, uint64_t)>;
```
amallia: Although I agree the return type is always a float, not sure that it will take always the same input. Should we include this change in all the query algorithms? Maybe in a separate PR?

include/query/algorithm/exhaustive_taat_query.hpp
```diff
@@ -0,0 +1,138 @@
...
+                break;
+            }
+            uint64_t end = std::max(accumulator_block_size * (block + 1), m_accumulators.size());
+            for (uint64_t docid = accumulator_block_size * block; docid < end; ++docid) {
+                m_topk.insert(m_accumulators[docid], docid);
+            }
+        }
+    }
+
+    // TODO(michal): I think this should be eventually the `operator()`
```
amallia: Agreed

include/query/algorithm/exhaustive_taat_query.hpp
```diff
@@ -0,0 +1,138 @@
...
+    uint64_t taat(std::vector<Cursor> cursors, std::vector<score_function_type> score_functions) {
+        m_topk.clear();
+        if (cursors.empty()) {
+            return 0;
+        }
+        init();
+        for (uint32_t term = 0; term < cursors.size(); ++term) {
+            auto &cursor         = cursors[term];
+            auto &score_function = score_functions[term];
+            if constexpr (std::is_same_v<typename Cursor::enumerator_category,
```
amallia: This will not allow us to use encodings that are not block-based

include/query/algorithm/exhaustive_taat_query.hpp
```diff
@@ -0,0 +1,138 @@
...
+                                         ds2i::block_enumerator_tag>) {
+                while (cursor.docid() < m_accumulators.size()) {
+                    auto const &documents = cursor.document_buffer();
+                    auto const &freqs     = cursor.frequency_buffer();
+                    for (uint32_t idx = 0; idx < documents.size(); ++idx) {
+                        intrinsics::prefetch(&m_accumulators[documents[idx + 3]]);
+                        auto& accumulator = m_accumulators[documents[idx]];
```
amallia: Did you measure any improvements here? There shouldn’t I think

elshize:
> From my understanding `nodiscard` is used to enforce the programmer to use the returned value of a function. For example, this can be useful when an error is returned and the error checking has to be enforced. In this context I believe `nodiscard` can be avoided.

There's simply no reason for you calling a pure function or a const method and not use the result. E.g., if you use:
```cpp
docenum.document_buffer();
```
it makes no sense, either you forgot to use it or it simply should be removed. The most reasonable approach is to give a warning there, which is precisely what `nodiscard` will do. So:

- All pure free functions (not changing any state) should be `nodiscard` by default.
- All const methods should be `nodiscard` by default.

I would go even further:

- Any function whose return value in 99% of cases should be used, should be `nodiscard` by default. In the rare cases when you might not want to do that, there's always `[[maybe_unused]]` attribute.

Of course, there might be exceptions to the rules but I see no reason to avoid `nodiscard`. In fact, this might be a language default if not for backward-compatibility. For example, it is default in Rust.

include/query/algorithm/exhaustive_taat_query.hpp
elshize: Two things here:

1. As to the separate PR, note that this is all internal, doesn't really affect anything outside. I wanted to extract it to another function so we have logic separated. You can revert it now but it makes little sense. Exposing it as an interface should, however, be done in a separate PR, I agree.

2. As to argument types, my thinking is the following: this is always the function of a posting: (doc, freq), which is (`int`, `int`). The idea is for another function/method to return these functions. They might be, say, lambdas with captured index, which resolve the document size. It can be something completely different. The point is that by separating these two things, your processing algorithm cares not about what it must pass, it always passes the posting. Figuring out what to do with it falls onto index/scoring function, as it should.

include/query/algorithm/exhaustive_taat_query.hpp
elshize: I guess. I think I'm just not a big fan of unsigned integers for arithmetic. But I used it coz all the other sizes seem to be unsigned, and I left the assert.

include/query/algorithm/exhaustive_taat_query.hpp
elshize: Hence the `if constexpr` here. This distinction is for three reasons:

1. If we have blocks, it is more efficient to exhaustively iterate over a range instead of repeatedly calling `next()`.
2. I wanted to test out prefetching (not finished there yet).
3. I want at some point I wanted to figure out if there's something we can do with SIMD there.

The 'standard' way would be implemented in `else`.

include/query/algorithm/exhaustive_taat_query.hpp
elshize: Yeah, I haven't really found any yet.

elshize:
One thing to note here is that this is not finished. I'll push stuff sometimes to get it to the server. But I'm working slowly over Christmas so changes are slow to appear.

That doesn't mean you can't comment, feel free, just keep that in mind that this is still work in progress to test out some stuff.

include/query/algorithm/exhaustive_taat_query.hpp
elshize: Actually, it **should** be a signed integer. Just think about it. Someone could very well declare `exhaustive_taat_query<.., .., -1>` thinking it's "no blocks" or by accident. It will compile and define the block size to a huge number. That's just weird and unexpected. That's precisely why you shouldn't use unsigned integers unless you do bit manipulation. Sadly, someone thought a while ago that a container size is `size_t`, which is stupid.

elshize:
I'm wondering about a few things.

1. Why two of the tests don't find OpenMP but the rest does?
2. If we intend to use it, we should just require it in CMake. I don't see a reason not to, and optional dependencies complicate things that should be simple (like including a header file).

amallia:
> Why two of the tests don't find OpenMP but the rest does?

In this case we need to make OpenMP required and link to the target:
```
find_package(OpenMP REQUIRED)
target_link_libraries(target_name PRIVATE OpenMP::OpenMP_CXX)
```
> If we intend to use it, we should just require it in CMake. I don't see a reason not to, and optional dependencies complicate things that should be simple (like including a header file).

ATM is not required because it is only used by `stxxl` and it is optional there, since it works without OpenMP too.

src/queries.cpp
```diff
@@ -107,32 +107,32 @@ void perftest(const std::string &index_filename,
         logger() << "Query type: " << t << std::endl;
         std::function<uint64_t(term_id_vec)> query_fun;
         if (t == "and") {
-            query_fun = [&](term_id_vec query) { return and_query<false>()(index, query); };
+            query_fun = [&](term_id_vec query) { return and_query<IndexType, false>(index)(query); };
```
elshize: You don't need the lambda here, do you? `and_query(index)` is already a right type of function.

amallia:
Removed "block traversal" since it was wrong and fixing it does not make any difference in terms of performance.
BTW the problem was:
https://github.com/pisa-engine/pisa/blob/3672c4c4ded444464cfdaf724010cceda828dfe7/include/pisa/block_posting_list.hpp#L114

A block of values needs to be shifted by its universe.

elshize:
@amallia Should we merge this?

amallia:
I need to look at Maxscore and the blocked/lazy accumulators, but I already started.
We could create a separate branch containing simple accumulator and exhaustive taat and we merge it, in the meantime I look at the rest. What do you think?

elshize:
> I need to look at Maxscore and the blocked/lazy accumulators, but I already started.
> We could create a separate branch containing simple accumulator and exhaustive taat and we merge it, in the meantime I look at the rest. What do you think?

Nah, it's fine, just do your thing and let's merge once done.

amallia:
I have merged in a separate PR (https://github.com/pisa-engine/pisa/pull/73) the simple `ranked_or_taat` algorithm. 
There were also some general improvements to the query algorithms, so I thought it was worth to merge separately.

PR #16, opened 2018-12-23
Top-k heap performance (Issue #15)

### LINKED ISSUE #15: Improve topk_queue::insert() to require only one pass down the heap
In case when `topk_queue::size() == k`, we perform a pop and a push on the heap, which can be avoided.

Thanks to the property that the queue stays the same length, we can do a little trick.

1. We make sure that our internal array is at least `k + 1` long; or, in other words, the capacity is this much.
2. When pushing to the heap, we do the following:
  2.1. If we're at the threshold or above, we push/emplace back the current entry (even if heap full).
  2.2. If the size after push is at most `k`, then we simply do `std::push_heap`.
  2.3. If we overflow by one, we perform a `std::pop_heap` on the entire vector, and remove the last element.

### Explanation
The pop instruction in 2.3 will get rid of the top entry, replace it with the new one, and fix the heap. Only one heap traversal is necessary. This might not matter for `k = 10` but for `k = 1000` and beyond, it very well might. And I don't see any drawbacks of doing it this way. It should always be faster.

elshize:
Fixes #15

### DIFF
--- a/include/topk_queue.hpp
+++ b/include/topk_queue.hpp
@@ -30,13 +30,16 @@ struct topk_queue {
             return false;
         }
         m_q.emplace_back(score, docid);
-        if (m_q.size() < m_k) {
+        if (DS2I_UNLIKELY(m_q.size() < m_k)) {
             std::push_heap(m_q.begin(), m_q.end(), min_heap_order);
+            if(DS2I_UNLIKELY(m_q.size() == m_k)) {
+                m_threshold = m_q.front().first;
+            }        
         } else {
             std::pop_heap(m_q.begin(), m_q.end(), min_heap_order);
             m_q.pop_back();
+            m_threshold = m_q.front().first;
         }
-        m_threshold = m_q.size() == m_k ? m_q.front().first : 0.0;
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


include/topk_queue.hpp
```diff
@@ -1,59 +1,49 @@
...
-                                   return l.first > r.first;
-                               });
-                threshold = m_q.front().first;
-                return true;
-            }
+            std::pop_heap(m_q.begin(), m_q.end(), min_heap_order);
+            m_q.pop_back();
         }
-        return false;
+        m_threshold = m_q.size() == m_k ? m_q.front().first : 0.0;
```
amallia: I feel like we can avoid this check, since we are inserting there is something for sure. 
Shall we change it to `m_threshold = m_q.front().first`?

amallia (approved):
I think this makes a lot of sense. Thanks for this smart PR. 
We can merge

include/topk_queue.hpp
elshize: We can't do that here, because if k=10, and you inserted 9-th element, the actual threshold is still 0. But we can do it a little better: check in the `if` branch, and not check in the `else` branch. We should also make it "unlikely" to go into `if` branch.

I'll do it sometime today.

amallia: Yes, you are right! I have updated it. Please let me know if this is what you mean.

elshize:
Looks good

PR #25, opened 2018-12-29
Support thresholds for query performance/correctness tests

### LINKED ISSUE #23: Enable using threshold/threshold estimates for performance/correctness test
Threshold should be in a file, passed with an optional argument, and propagated as an optional vector. Decision whether or not to use threshold should be compile-time, so it doesn't slow down other types of queries.

elshize:
Fixes #23


elshize:
Stalled by #26

src/thresholds.cpp
```diff
@@ -0,0 +1,129 @@
...
+        mapper::map(wdata, md, mapper::map_flags::warmup);
+    }
+
+    wand_query<WandType> query_func(wdata, k);
+    for (auto const& query : queries) {
+        query_func(index, query);
+        auto  results   = query_func.topk();
+        float threshold = 0.0;
+        if (results.size() == k) {
+            auto min = std::min_element(
```
amallia: Do we need this?

test/test_ranked_queries.cpp
```diff
@@ -7,73 +7,80 @@
...
-
-            for (auto const& q: queries) {
-                or_q(index, q);
-                op_q(index, q);
-                REQUIRE(or_q.topk().size() == op_q.topk().size());
-                for (size_t i = 0; i < or_q.topk().size(); ++i) {
-                    REQUIRE(or_q.topk()[i].first == Approx(op_q.topk()[i].first).epsilon(0.1)); // tolerance is % relative
-                }
+        std::string t;
+        std::ifstream tin(DS2I_SOURCE_DIR "/test/test_data/thresholds");
```
elshize: The file is `top5_thresholds`!

src/thresholds.cpp
elshize: I guess we don't, I wasn't aware at the time. Will fix.

test/test_ranked_queries.cpp
```diff
@@ -7,73 +7,80 @@
...
+        while (std::getline(tin, t)) {
+            thresholds.push_back(std::stof(t));
```
amallia: Where do we use this? It would be good to test w/ and w/o thresholds

elshize: We don't use it yet because we don't have any algorithm that takes thresholds. But this way I can choose to use them in any test case (but don't have to).

amallia (approved):

amallia:
Okay, once the comments have been addressed it can be merged

Issue #29, opened 2019-01-02
Remove Boost.Test


amallia:
Rewrite existing tests and get rid of Boost.Test

amallia:
Addressed in https://github.com/pisa-engine/pisa/pull/30

Issue #28, opened 2019-01-02
Add Catch2


amallia:
Add Catch2 for the new testing framework. So we can write new tests already in it

amallia:
Addressed in https://github.com/pisa-engine/pisa/pull/30

PR #30, opened 2019-01-03
Catch2

amallia:
This should address both:
- https://github.com/pisa-engine/pisa/issues/28
- https://github.com/pisa-engine/pisa/issues/29


elshize (approved):
Looks good. One comment, though: I think one of the strengths of Catch2 over Boost is how you can write more intuitive code without test classes and `TEST_CASE_METHOD`. Just wanted to state that for the future so you don't miss out on some nice features. But no sense in refactoring the tests we already have, so I'm merging this.

PR #31, opened 2019-01-04
Revert top-k queue changes

amallia:
This reverts commit 892de4054eee8f7fbcfab2bd89823b81e034e04e, which causes queries.cpp to stuck for no apparent reason.

### DIFF
--- a/include/topk_queue.hpp
+++ b/include/topk_queue.hpp
@@ -1,52 +1,59 @@
 #pragma once
-
 namespace ds2i {
 
 struct topk_queue {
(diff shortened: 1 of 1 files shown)


elshize:
> for no reasons makes the query processing stuck

It might be good to figure out the actual reason (nothing happens without it, at the risk of sounding philosophical). I feel uneasy not understanding what was wrong.

How to reproduce this error?

amallia:
Fixed in https://github.com/pisa-engine/pisa/pull/33

Issue #26, opened 2018-12-29
Discrepancies in query result set sizes


elshize:
https://github.com/pisa-engine/pisa/blob/8177a472f770d50e09676d68fbe74a97c657985d/test/test_ranked_queries.cpp#L45

There seems to be something wrong with query results when I vary `k`. (Or I don't understand somethinig.)

To show what I mean, I change this to:
```cpp
            ranked_or_query<WandType> or_q(wdata, 1);

            for (auto const& q: queries) {
                or_q(index, q);
                op_q(index, q);
                //BOOST_REQUIRE_EQUAL(or_q.topk().size(), op_q.topk().size());
                std::cerr << or_q.topk().size() << '\n';
                for (size_t i = 0; i < or_q.topk().size(); ++i) {
                    BOOST_REQUIRE_CLOSE(or_q.topk()[i].first, op_q.topk()[i].first, 0.1); // tolerance is % relative
                }
            }
```

I printed out `ranked_or_query` result sizes for all queries with both `k = 1` and `k = 10`. In the former case, there are some non-zero sizes (nothing equal 10, though). In the latter, all are zeroes **when I expected some ones**.

Here's the output files: [k10.txt](https://github.com/pisa-engine/pisa/files/2716553/k10.txt) and [k1.txt](https://github.com/pisa-engine/pisa/files/2716554/k1.txt)

If I set `k = 1`, I should stil get the same top result as for `k = 10`, shouldn't I?

elshize:
I wrote and pushed (#33) test that IMO should be passing but doesn't. At least for me ~locally~ on `xor`, we can wait and see what happens on Travis.

elshize:
Fixed by #33

PR #33, opened 2019-01-04
Test showing the problem in #26

elshize:


elshize:
Fixes #26 and makes #31 obsolete.

PR #34, opened 2019-01-04
Removed boost as a system dependency

amallia:


amallia:
@elshize for review. Squash and merge if good. Thanks

elshize (approved):

PR #36, opened 2019-01-04
Added clang

amallia:


amallia:
@elshize this should pass now. Feel free to squash and merge once approved.

PR #37, opened 2019-01-04
Parsing

elshize:


include/binary_collection.hpp
```diff
@@ -138,24 +114,26 @@ namespace ds2i {
 
                 size_t n = 0;
                 size_t pos = m_pos;
-                while (!(n = m_collection->m_data[pos++])); // skip empty seqs
+                n = m_collection->m_data[pos++];
```
amallia: Let's add an assert here.

src/invert.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+int main(int argc, char **argv) {
+
+    std::string input_basename;
+    std::string output_filename;
+    size_t      threads = std::thread::hardware_concurrency();
+    //ptrdiff_t   batch_size = 100'000;
+
+    CLI::App app{"parse_collection - parse collection and store as forward index."};
```
amallia: Wrong copy paste

elshize:
Parsing tests are ready. Now just inverting test and will do some manual sanity checks for a larger collection. Plus some minor changes. So almost done.

src/parse_collection.cpp
```diff
@@ -0,0 +1,63 @@
...
+                      output_filename,
+                      [](std::istream &in) -> std::optional<Warc_Record> {
+                          Warc_Record record;
+                          if (read_warc_record(in, record)) {
+                              return record;
+                          }
+                          return std::nullopt;
+                      },
+                      [&](std::string const &term) -> std::string {
+                          return Porter2_Stemmer{}(tolower(term));
```
amallia: Do not bother here, just use:
```cpp
#include <boost/algorithm/string.hpp>

boost::to_lower(term);
boost::to_lower_copy(term)
```

amallia: Worst case:
```cpp
#include <algorithm>
#include <string> 

std::transform(term.begin(), term.end(), new_term.begin(), ::tolower);
```

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,298 @@
...
+                first_document += batch_size;
+                record_batch = std::vector<Record>();
+            }
+        }
+        batch_group.wait();
+        merge(output_file, static_cast<ptrdiff_t>(first_document), batch_number);
+    }
+};
+
+std::string tolower(std::string const &term) {
```
amallia: Let's remove this, so we increase code coverage...

amallia:
`snowball.hpp` kills code coverage. Lets use https://github.com/pisa-engine/Porter2

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+        std::getline(in, line);
+    }
+}
+
+} // namespace warc
+
+std::istream& read_warc_record(std::istream& in, Warc_Record& record)
+{
+    std::string version;
+    if (not warc::read_version(in, version)) {
```
amallia: We should add something like the following in order to skip empty lines before reaching the version
```
while (line.empty())
   line = read()
```

src/parse_collection.cpp
elshize: Let's go with the standard one.

Also, I actually found that now I don't even use the parsing function anywhere, I'll fix that, and redesign slightly to be more efficient than copying a string three times.

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+    std::string line;
+    std::getline(in, line);
+    while (not line.empty()) {
+        std::regex  field_pattern("^(.+):\\s+(.*)$");
+        std::smatch sm;
+        if (not std::regex_search(line, sm, field_pattern)) {
+            throw Warc_Format_Error(line, "could not parse field: ");
+            //std::cerr << "cound not parse field: " << line << std::endl;
+        } else {
+            fields[sm.str(1)] = sm.str(2);
```
amallia: Should we lower `sm.str(1)`? And maybe trim `sm.str(2)`?

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+} // namespace warc
+
+std::istream& read_warc_record(std::istream& in, Warc_Record& record)
+{
+    std::string version;
+    if (not warc::read_version(in, version)) {
+        return in;
+    }
+    record.version_ = std::move(version);
+    warc::read_fields(in, record.warc_fields_);
```
amallia: We should check `content-length` here. If `0` stop.

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+    std::string line;
+    if (record.type() != "warcinfo") {
+        std::getline(in, line);
+    }
+    warc::read_fields(in, record.http_fields_);
```
amallia: We should check `content-length` here. If `0` stop.
Note: there are two `content-length`: one in the warc header and another in the http header.

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+    if (record.type() != "warcinfo") {
+        try {
```
amallia: Should we enforce `record.type() == 'response'`?

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+            std::size_t length = std::stoi(record.content_length());
+            record.content_.resize(length);
+            in.read(&record.content_[0], length);
```
amallia: A this point we have read `warc` and `http` headers. Now, shouldnt we skip 2 lines and then read the content?
If so, instead of skipping 2 lines, it would be better to match `\n\s*\n`

include/parsing/warc.hpp
elshize: I think that according to the format specification, this shouldn't happen (there are strict rules about empty lines). But I also **think** this won't hurt so we might as well make sure we skip any empty lines at the beginning.

include/parsing/warc.hpp
elshize: We should lower `sm.str(1)`. I'll modify the regex to make sure there's no white spaces around `sm.str(2)`.

include/parsing/warc.hpp
elshize: I don't think we should but correct me if I'm wrong. Here's a excerpt from a WARC file:
```
WARC/0.18
WARC-Type: response
WARC-Target-URI: http://00000-nrt-realestate.homepagestartup.com/
WARC-Warcinfo-ID: 993d3969-9643-4934-b1c6-68d4dbe55b83
WARC-Date: 2009-03-65T08:43:19-0800
WARC-Record-ID: <urn:uuid:67f7cabd-146c-41cf-bd01-04f5fa7d5229>
WARC-TREC-ID: clueweb09-en0000-00-00000
Content-Type: application/http;msgtype=response
WARC-Identified-Payload-Type: 
Content-Length: 16558

HTTP/1.1 200 OK
Content-Type: text/html
Date: Tue, 13 Jan 2009 18:05:10 GMT
Pragma: no-cache
Cache-Control: no-cache, must-revalidate
X-Powered-By: PHP/4.4.8
Server: WebServerX
Connection: close
Last-Modified: Tue, 13 Jan 2009 18:05:10 GMT
Expires: Mon, 20 Dec 1998 01:00:00 GMT
Content-Length: 16254

<head> <meta http-equiv="Content-Language" content="en-gb">
```

The `read_fields` function for HTTP will read the `content-length` line with `getline`, and that's the last parsed field. But we actually stop when **we have read an empty line**. So the empty line below will have been read when we are finished.

Am I right here?

amallia: The line break between the http header and the content might be something like `\r\n`. So not sure if you read empty line or `\r`. (https://stackoverflow.com/questions/45956271/stdgetline-reads-carriage-return-r-into-the-string-how-to-avoid-that)

include/parsing/warc.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+std::istream& read_warc_record(std::istream& in, Warc_Record& record)
+{
+    std::string version;
+    if (not warc::read_version(in, version)) {
+        return in;
+    }
+    record.version_ = std::move(version);
+    warc::read_fields(in, record.warc_fields_);
+    std::string line;
```
amallia: We should check `content-length` here. If 0 stop.
Note: there are two `content-length`: one in the warc header and another in the http header.

include/parsing/warc.hpp
elshize: You mean it can be `Content-Length: 219\r\n\n<content>` or `Content-Length: 219\r\n<content>`?

amallia: `Content-Length: 219\r\n\r\n<content>`

- First one will read: `Content-Length: 219\r`
- Second one will read: `\r` which is != than `''`

elshize: I want to write unit tests for all those possibilities.

amallia: Sure. 
As a side note, this is also the reason why I proposed to use some `strip` mechanism.

elshize:
Seems to be working now on 3.7 mln collection. I'll check the entire Clueweb overnight.

amallia:
There is still a small issue with compilation.
When running `make -j`, since it runs in parallel, it reaches `parse_collection` before gumbo has been compiled. 
Usually this never happens, but in this case gumbo is handled differently (as if you install it in the build subfolder) and I am not too familiar with this technique.

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,288 @@
...
+
+    void run(Batch_Process bp, process_term_function_type process_term) const
+    {
+        auto basename = batch_file(bp.output_file, bp.batch_number);
+
+        std::ofstream os(basename);
+        std::ofstream title_os(basename + ".documents");
+        std::ofstream url_os(basename + ".urls");
+        std::ofstream term_os(basename + ".terms");
+        write_header(os, bp.records.size());
```
amallia, line 137: Where do we get rid of these files?

include/parsing/warc.hpp
```diff
@@ -0,0 +1,159 @@
...
+        auto &field_value = http_fields_.at("content-length");
+        try {
+            return std::stoi(field_value);
+        } catch (std::invalid_argument &error) {
+            throw Warc_Format_Error(field_value, "could not parse content length: ");
+        }
+    }
+    [[nodiscard]] auto content() -> std::string & { return content_; }
+    [[nodiscard]] auto content() const -> std::string const & { return content_; }
+    [[nodiscard]] auto url() const -> std::string const & {
```
amallia: Never called. Indeed `.url` file is empty

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,288 @@
...
+                        queue.try_pop(x);
+                    });
+                    ++batch_number;
+                    first_document += last_batch_size;
+                    break;
+                }
+            } catch (Warc_Format_Error &err) {
+                continue;
+            }
+            if (record->type() != "response") {
```
amallia: Not too generic. Once we plug another parse might not be `response` anymore.
How about a `is_valid()` function?

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,288 @@
...
+                first_document += batch_size;
+                record_batch = std::vector<Record>();
+            }
+        }
+        batch_group.wait();
+        merge(output_file, static_cast<ptrdiff_t>(first_document), batch_number);
+    }
+};
+
+class Plaintext_Record {
```
amallia, line 307: Should Plaintext_Record have url too?

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,288 @@
...
+   public:
+    Plaintext_Record() = default;
+    Plaintext_Record(std::string trecid, std::string content)
+        : m_trecid(std::move(trecid)), m_content(std::move(content)) {}
+    [[nodiscard]] auto content() -> std::string & { return m_content; }
+    [[nodiscard]] auto content() const -> std::string const & { return m_content; }
+    [[nodiscard]] auto trecid() -> std::string & { return m_trecid; }
+    [[nodiscard]] auto trecid() const -> std::string const & { return m_trecid; }
+    [[nodiscard]] auto type() const noexcept -> std::string { return "response"; }
```
amallia: I would remove this

include/forward_index_builder.hpp
```diff
@@ -0,0 +1,288 @@
...
+            std::map<std::string, std::unordered_map<std::ptrdiff_t, std::ptrdiff_t>>;
+        std::ofstream title_os(basename + ".documents");
+        std::ofstream url_os(basename + ".urls");
+        std::ofstream term_os(basename + ".terms");
+
+        logger() << "Merging titles\n";
+        for (auto batch : enumerate(batch_count)) {
+            std::ifstream title_is(batch_file(basename, batch) + ".documents");
+            title_os << title_is.rdbuf();
+        }
```
amallia, line 180: Does it mean we need to be able to load all the `.documents` batches in memory? 
Fine for now, but we should think about a k-way merge solution in the next release. `Stxxl` (which is already a dependency) will make all this very easy!

include/forward_index_builder.hpp
elshize: We don't. We will. Same goes for inverting.

include/parsing/warc.hpp
elshize: We will need them though. We should rather add this when parsing collection. That's because we'll need those URLs if we want to do URL reordering.

include/forward_index_builder.hpp
elshize: Agreed.

include/forward_index_builder.hpp
elshize: Most likely yes. My plaintext doesn't have one though. This format was initially a workaround, so now we should agree on a specific format. I would add `url` but always return an empty string **for now**, and open a separate issue for that: to agree on a format and fix it here. I don't want to spend to much time on this right now, and don't want to change format just yet so my already parsed files can work in case I still need to use them.

include/forward_index_builder.hpp
elshize: Yeah, if we write a `valid` function, this won't be needed.

include/forward_index_builder.hpp
elshize: No, you load only one at a time, and redirect to another file. Note that `title_is` has a local scope, it will be closed after each iteration. And `title_os` is `ofstream`.

include/forward_index_builder.hpp
amallia: All this makes a lot of sense. Thanks.
Let's keep in mind that `Plaintext_Record` is still very important in order to parse collections that dont come in `warc` or `trec` format, like for example a wikipidia dump. 
So please do not remove `Plaintext_Record`

include/forward_index_builder.hpp
elshize: Same goes for URLs. What we **have to store in memory** are terms at this point. Because we have an in-memory map. That could be maybe improved at a later time.

elshize:
> There is still a small issue with compilation.
> When running `make -j`, since it runs in parallel, it reaches `parse_collection` before gumbo has been compiled.
> Usually this never happens, but in this case gumbo is handled differently (as if you install it in the build subfolder) and I am not too familiar with this technique.

This is very strange. I'd think cmake should figure out the order in which targets can be compiled. Right now, gumbo is more or less added as another library within a project. And it's set to be a dependency for some executables. This is not so unusual.

Anyway, I'll look into that and see what can be done.

include/forward_index_builder.hpp
elshize: Right, but that's why we will need to be careful about defining what we expect of a plaintext format. Anyway, let's do that later.

elshize:
We should start thinking about closing this one and moving any improvements to separate issues.

There are some things to be done yet before we can merge it:

- [ ] ~Integration test for parsing WARC~ **will open a new issue, see that for details**
- [x] Removing batches for `invert`
- [x] We should be more careful accessing WARC fields, I use `at` right now that will throw if a field isn't there, that should be changed.
- [ ] ~Figure out what's the problem with gumbo not building before its dependents.~ Let's do it in a separate PR.

Anything else?

include/parsing/warc.hpp
```diff
@@ -0,0 +1,160 @@
...
+        return warc_fields_.at("warc-target-uri");
+    }
+    [[nodiscard]] auto trecid() const -> std::string const & {
+        return warc_fields_.at("warc-trec-id");
```
amallia: I realized this is non standard. When I tied to parse a CC-NEWS collection it was not able to parse any documents.

```
WARC/1.0
WARC-Record-ID: <urn:uuid:4e7b712c-f7fd-4e31-81ad-b6b8b9a190c8>
Content-Length: 38840
WARC-Date: 2016-12-01T13:28:43Z
WARC-Type: response
WARC-Target-URI: http://www.banker.bg/finansov-dnevnik/read/parlamentut-prie-biudjetite-na-vsichki-ministerstva?utm_source=rss&utm_medium=click&utm_campaign=rss
Content-Type: application/http; msgtype=response
WARC-Payload-Digest: sha1:NKWIH3RIPHRCTJ7F7VM45KZFUR5DCXMW
WARC-Block-Digest: sha1:EOKDPEU3LQ7XESEPNRQO34YDAEDADIL6
```

For now I propose to ignore. If we use a `CC-NEWS` collection we will need to pre-process the data anyway (for other reasons I am not explaining here), so we can just add a `warc-trec-id` during that phase. 

I just wanted to document it here.

elshize: Just open a new issue and copy it there.

It makes sense that TREC ID is not a WARC standard. But that begs some questions:

- Is there a "CC-NEWS" ID by which you can identify a specific document?
- For that matter, are there any evaluation queries we could use, say, to test precision?

elshize: #55

elshize:
I added a few issues: #59, #60, and #61. Let's aim to close this one and work on WARC there.

amallia:
Moved WARC parser to separate repo: https://github.com/pisa-engine/warcpp

amallia (approved):

Issue #4, opened 2018-09-26
Add progress bar to index construction task #9


amallia:
Add progress bar to the verify index task

amallia:
Addressed in https://github.com/pisa-engine/pisa/pull/39

PR #39, opened 2019-01-04
Replaced progress logger with progress bar

amallia:
To address: https://github.com/pisa-engine/pisa/issues/4


elshize:
It looks good but didn't you say there was a problem with the progress bar? I vaguely remember something.

amallia:
Yes the problem was when we used an external thread to print. I removed the external thread and now it prints only on update.

elshize:
Right, now I remember. We should probably improve that at some point.

elshize (approved):

Issue #47, opened 2019-01-08
Replace boost::optional with std::optional


amallia:
https://github.com/pisa-engine/pisa/blob/3bfbdec0da399dcb278d61d496ae348b999644fe/src/create_freq_index.cpp

https://github.com/pisa-engine/pisa/blob/61261d3b9c2515168be0bb420559227a53cb0468/src/profile_queries.cpp

https://github.com/pisa-engine/pisa/blob/61261d3b9c2515168be0bb420559227a53cb0468/src/queries.cpp

PR #68, opened 2019-01-17
Moved to std optional

amallia:
Addresses https://github.com/pisa-engine/pisa/issues/47


amallia:
@elshize check and merge. Just removed Boost/Optional from everywhere

elshize:
The reason why you're now using `*optional` instead of `optional.value()` is because of apple clang thing, is that correct?

amallia:
> The reason why you're now using *optional instead of optional.value() is because of apple clang thing, is that correct?

Yes, unless we want to drop support of early version of `AppleClang`

elshize:
No, it's fine.

Issue #40, opened 2019-01-06
Use Catch2 CMake functions to discover tests


elshize:
We can enable Catch2 helpers with:
```
include(CTest)
include(Catch)
```
and then add tests with
```
catch_discover_tests(executable_name)
```

It's nicer because it exposes all individual test cases to `ctest` instead of each file only.

**NOTE:** This automatically worked with `find_package(Catch2)`, not sure if you need to do some extra include when you use a subdirectory, but I'm sure it's very easy either way.

elshize:
Another note: this would allow us to filter test by tags and we could tag integration and unit test to make it easier and faster to run only unit tests.

PR #69, opened 2019-01-17
Expose Catch2 test cases to ctest

elshize:
Solves #40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: require
 language: cpp
-dist: trusty
+dist: xenial
 
 branches:
   only:
@@ -8,9 +8,8 @@ branches:
 
 common_sources: &all_sources
   - ubuntu-toolchain-r-test
-  - llvm-toolchain-trusty
-  - llvm-toolchain-trusty-5.0
-  - llvm-toolchain-trusty-6.0
+  - llvm-toolchain-xenial
+  - llvm-toolchain-xenial-6.0
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,16 +1,15 @@
 sudo: require
 language: cpp
-dist: trusty
+dist: xenial
```
amallia: Just for curiosity, what is the advantage of using xenial?

I think we should move the dist inside the matrix, under the os.
```
matrix:
  include:
    - os: linux
      dist: trusty
    - os: osx
      osx_image: xcode7.2
```

test/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY  ${CMAKE_BINARY_DIR}/test)
 
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../external/Catch2/contrib)
```
amallia: I think we can avoid this. Let me check how we can do it better.

.travis.yml
elshize: Well, first of all, I wanted to see if it would work because I had a problem with cmake 3.9. So I wanted to run it with 3.10 and install it but then I saw they had xenial and thought I'd try that first.

But I haven't figured out what was exactly wrong for 3.9 (I think it should work in principle). So we should think about what to do with this. The easiest would be to just bump cmake required version to 3.10. I'm not convinced what the right thing to do here is.

CMakeLists.txt
```diff
@@ -95,7 +93,7 @@ target_include_directories(pisa INTERFACE external)
 
 add_subdirectory(src)
 
-if (BUILD_TESTING AND NOT_SUBPROJECT)
+if (PISA_ENABLE_TESTING)
```
elshize: `BUILD_TESTING` is a standard way of turning off test for CTest: https://cmake.org/cmake/help/v3.13/module/CTest.html

amallia: Yes I am aware, but why do we need it?
`BUILD_TESTING` is always `ON`. If you include pisa as a subproject you will eventually set `PISA_ENABLE_TESTING` to `OFF`. How can `BUILD_TESTING` help?

elshize: `BUILD_TESTING` is supposed to override any other options. So I would change it to:
```
if (PISA_ENABLE_TESTING AND BUILD_TESTING)
```

amallia (approved):

PR #70, opened 2019-01-17
Support parallel tests by using generated temp directories

### LINKED ISSUE #38: Tests fail when run in parallel
This one is low-priority but when I run, e.g.,  `ctest -j8`, one of the tests usually fails with `Bus error`. I suspect several tests write to the same file or something.

But as I said, this one is low priority.

elshize:
Fixes #38

### DIFF
--- a/test/temporary_directory.hpp
+++ b/test/temporary_directory.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "boost/filesystem.hpp"
+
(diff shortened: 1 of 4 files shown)


elshize:
@amallia I use a random temporary file to write to, this way there are no conflicts when test are run in parallel. Tiny change. Check and merge if okay.

amallia:
Awesome! Thanks

PR #74, opened 2019-01-21
Added wand data range-based

amallia:
Wand data (max-score and block-max-score) for docid range -based layout


include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include "wand_utils.hpp"
+
+namespace pisa {
+
+template <size_t range_size = 128, size_t min_list_lenght = 1024, typename Scorer = bm25>
+class wand_data_range {
+   public:
+
+    template<typename List>
+    std::vector<float> compute(List &list, std::function<float(uint64_t)> norm_lens) const {
```
elshize: From the signature, it's unclear what this does. Let's call it `compute_block_max_scores` maybe?

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+        std::vector<float> block_max_scores(m_blocks_num, 0.0f);
+        for (int i = 0; i < list.size(); ++i) {
```
elshize: Just an idea for the future: since our lists are not compliant with standard iterators, therefore we can't use range-loop, maybe we should implement a `for_each_posting` to do this?
```cpp
for_each_posting(list, [&](auto docid, auto freq) {
            float score = Scorer::doc_term_weight(freq, norm_lens(docid));
            size_t pos = docid/range_size;
            float &bm = block_max_scores[pos];
            bm = std::max(bm, score);
});
```

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+        uint64_t                              block_start;
+        mapper::mappable_vector<float> const &m_block_max_term_weight;
+    };
+
+    enumerator get_enum(uint32_t i) const {
+        return enumerator(
+            m_blocks_start[i], m_block_max_term_weight);
+    }
+
+    static std::vector<bool> compute_live_blocks(std::vector<enumerator> &enums, float threasold, uint64_t begin, uint64_t end) {
```
elshize: Typo in `threshold`

include/pisa/wand_data_range.hpp
elshize: See, you're using `norm_lens` as I used scorers (in this case one at a time). Same idea but here you unnecessarily transfer some details to `compute`, which `compute` shouldn't care about. In this case, why not just pass a `std::function<float(int, int)>` and you get much cleaner, and easier to understand, code:
```cpp
float& current_max = block_max_score[docid / range_size];
current_max = std::max(current_max, score(docid, freq);
```
Put it together with `for_each_posting` (below), and you get a beautiful little function, that's a work of art:
```cpp
auto compute_block_max_scores = [](auto& list, auto score) {
    std::vector<float> block_max_scores(m_blocks_num, 0.0f);
    for_each_posting(list, [&](auto docid, auto freq) {
        float& current_max = block_max_score[docid / range_size];
        current_max = std::max(current_max, score(docid, freq);
    }
    return block_max_scores;
};
```

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+        return block_max_scores;
+    }
+
+    class builder {
+       public:
+        builder(partition_type                type,
+                binary_freq_collection const &coll,
+                global_parameters const &     params) {
+            (void)params;
```
elshize: Use `[[maybe_unused]] global_parameters const &     params) {` instead.

elshize: I think as a rule of thumb, you should use initializers, and I see no reason not to do it here:
```cpp
: blocks_num(ceil_div(coll.num_docs(), range_size)),
  type(type),
  total_elements(0),
  blocks_start{0},
  block_max_term_weight{}
{}
```

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+            auto posting_lists = std::distance(coll.begin(), coll.end());
+            spdlog::info("Storing max weight for each list and for each block...");
+            spdlog::info("Range size: {}. Number of docs: {}. Blocks per posting list: {}. Posting lists: {}."
+                , range_size, coll.num_docs(), blocks_num, posting_lists);
+            total_elements = 0;
+            blocks_start.push_back(0);
+        }
+
+        float add_sequence(binary_freq_collection::sequence const &seq,
+                           binary_freq_collection const &          coll,
```
elshize: Seems to be unused.

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+            blocks_num = ceil_div(coll.num_docs(), range_size);
+            auto posting_lists = std::distance(coll.begin(), coll.end());
+            spdlog::info("Storing max weight for each list and for each block...");
+            spdlog::info("Range size: {}. Number of docs: {}. Blocks per posting list: {}. Posting lists: {}."
+                , range_size, coll.num_docs(), blocks_num, posting_lists);
+            total_elements = 0;
+            blocks_start.push_back(0);
+        }
+
+        float add_sequence(binary_freq_collection::sequence const &seq,
```
elshize: Can we be a bit more clear here, like `float add_term_sequence(term_sequence)` or something that clarifies what we pass here. Confusing at the first glance for someone who reads for the first time (like me lol).

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+                           binary_freq_collection const &          coll,
+                           std::vector<float> const &              norm_lens) {
```
elshize: First of, it's inconsistent with passing a function in `compute`; I think there's an advantage in consistency here. Plus, same comment as before about passing score function instead.

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+            float max_score =0.0f;
+
+            std::vector<float> b_max(blocks_num, 0.0f);
+            for (auto i = 0; i < seq.docs.size(); ++i) {
```
elshize: Can we implement a light-weight enumerator (we gotta talk about this name though) for `binary_freq_collection`. Then, this code is just calling `compute`, which makes sense, because that's what we're doing, so we kill two birds with one stone: cleaner intent, and reuse of code.

elshize: Also, seems like we only use it if `seq.docs.size() >= min_list_lenght`, so we should do it within that block.

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,155 @@
...
+            }
+            return max_score;
+        }
+
+        void build(wand_data_range &wdata) {
+            wdata.m_blocks_num = blocks_num;
+            wdata.m_blocks_start.steal(blocks_start);
+            wdata.m_block_max_term_weight.steal(block_max_term_weight);
+            spdlog::info("number of elements / number of blocks: {}",
+                         static_cast<float>(total_elements) / static_cast<float>(wdata.m_block_max_term_weight.size()));
```
elshize: Tip: The second cast is superfluous, removing it can clean up the code slightly.

include/pisa/wand_data_range.hpp
elshize: `begin` and `end` of what? Unclear from context of signature. How about `begin_doc` or even `std::pair<int, int> document_range`?

src/create_wand_data.cpp
```diff
@@ -20,12 +21,15 @@ int main(int argc, const char **argv) {
     std::string output_filename;
     bool        variable_block = false;
     bool        compress       = false;
+    bool        range          = false;
 
     CLI::App app{"create_wand_data - a tool for creating additional data for query processing."};
     app.add_option("-c,--collection", input_basename, "Collection basename")->required();
     app.add_option("-o,--output", output_filename, "Output filename")->required();
     app.add_flag("--variable-block", variable_block, "Variable length blocks");
```
elshize: We're not supporting variable blocks for doc ranges right now, are we? Maybe we should define variable block and doc range as mutually exclusive?

elshize (changes requested):

include/pisa/wand_data_range.hpp
```diff
@@ -32,40 +32,41 @@ class wand_data_range {
 
     class builder {
        public:
-        builder(partition_type                type,
+        builder(partition_type type,
```
elshize, line 40: The `[[maybe_unused]]` attribute is supposed to tell the compiler that this argument might not be used, and therefore it won't generate a warning.

The same thing was achieved before with `(void)params;`. Now, this should be removed.

src/create_wand_data.cpp
amallia: yes!

include/pisa/wand_data_range.hpp
YuhongZhang98: Michal suggests that we may want to move these outside of the class.

include/pisa/wand_data_range.hpp
YuhongZhang98: It is called in wand_data.hpp?

include/pisa/wand_data_range.hpp
YuhongZhang98: Antonio says we're using a consistent function name here. If we change it we need to change every wand data file. Do we wanna do that?

include/pisa/wand_data_range.hpp
YuhongZhang98: Same as previous reason, Antonio suggests to leave it like this here

test/test_wand_data.cpp
```diff
@@ -0,0 +1,108 @@
...
+    builder.build(index);
+
+    SECTION( "Compute at run time" ) {
+        size_t i =0;
+        for (auto const &seq: collection) {
+            auto list = index[i];
+            if(seq.docs.size() < 1024) {
+                auto max = wdata_range.max_term_weight(i);
+                auto &w   = wdata_range.get_block_wand();
+                const mapper::mappable_vector<float> bm = w.compute_block_max_scores(list, Score_Function<Scorer, wdata_range>);
```
amallia: This does not even compile @YuhongZhang98 
Please fix it ASAP

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "pisa_config.hpp"
+#include "index_types.hpp"
+#include "query/queries.hpp"
+#include "wand_data_range.hpp"
+#include "scorer/score_function.hpp"
+
+using namespace pisa;
+
+TEST_CASE("wand_data_range") {
+    typedef wand_data_range<64, 1024, bm25> WandTypeRange;
```
elshize: Please use `using WandTypeRange = wand_data_range<64, 1024, bm25>;` instead.

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+    typedef wand_data<bm25, WandTypeRange> WandType;
+    using Scorer = bm25;
+
+    binary_freq_collection collection(PISA_SOURCE_DIR "/test/test_data/test_collection");
+    binary_collection      document_sizes(PISA_SOURCE_DIR "/test/test_data/test_collection.sizes");
+    WandType          wdata_range(document_sizes.begin()->begin(),
+                              collection.num_docs(),
+                              collection);
+    SECTION( "Precomputed block-max scores" ) {
+        size_t i = 0;
```
elshize: Nope, we're not optometrists, we don't deal with `i`s.

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+        for (auto const &seq: collection) {
```
elshize, line 30: What you want to do here: include `range/v3/view/enumerate.hpp` and use:
```cpp
for (auto &&[term_id, seq] : ranges::view::enumerate(collection))
```

[range-v3](https://github.com/ericniebler/range-v3) is an implementation of ranges that will come with C++20 standard. Here's a nice presentation if you're interested to find out more https://www.youtube.com/watch?v=mFUXNMfaciE&t=533s

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+            if(seq.docs.size() >= 1024) {
+                auto max = wdata_range.max_term_weight(i);
+                auto w   = wdata_range.getenum(i);
+                for (auto j = 0; j < seq.docs.size(); ++j) {
```
elshize: Nope. `j` means nothing to me. Remember we talked about what collection is and how to iterate over it? If you obfuscate variables like this, it's difficult for anyone to understand what you're doing, especially someone new to the code.

Another thing is that you're not actually using `j` anyway. You do it only to move the iterators. Which is not a great way to do it. Why not define iterators and increment them? Or better yet, use ranges:

```cpp
for (auto&& [docid, freq] : ranges::view::zip(seq.docs, seq.freqs)) {...}
```

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+                    auto docid = *(seq.docs.begin() + j);
+                    auto freq = *(seq.freqs.begin() + j);
+                    float score = Scorer::doc_term_weight(freq, wdata_range.norm_len(docid));
+                    w.next_geq(docid);
+                    CHECKED_ELSE(w.score() >= score) {
```
elshize: This is cool, I wasn't aware of this syntax. :1st_place_medal:

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+                        FAIL("Term: " << i << " docid: " << docid << ", block docid: " << w.docid());
+                    }
+                    REQUIRE(w.score() <= max);
+                }
+            }
+            i+=1;
+        }
+    }
+    typedef opt_index index_type;
```
elshize: `using index_type = opt_index;`

test/test_wand_data.cpp
```diff
@@ -0,0 +1,109 @@
...
+    index_type::builder builder(collection.num_docs(), params);
+    for (auto const &plist : collection) {
+        uint64_t freqs_sum =
+            std::accumulate(plist.freqs.begin(), plist.freqs.end(), uint64_t(0));
+        builder.add_posting_list(
+            plist.docs.size(), plist.docs.begin(), plist.freqs.begin(), freqs_sum);
+    }
+    builder.build(index);
+
+    SECTION( "Compute at run time" ) {
```
elshize: Check same things as above.

src/create_wand_data.cpp
elshize: @YuhongZhang98 this needs to be done. You can read how to do it here: https://github.com/CLIUtils/CLI11/

Basic idea is simple: `add_option` returns a pointer to option that you later pass to a `excludes` method of another option; something like this:
```cpp
auto opt1 = app.add_option(...);
auto opt2 = app.add_option(...)->excludes(opt1);
opt1->excludes(opt2);
```

elshize (changes requested):

elshize:
@YuhongZhang98 You gotta pay more attention to style. In fact, we have defined code style rules in `.clang-format` file, which is used by clang-format program to automatically format code. This doesn't necessarily have to be followed 100% but exceptions are very few. Your editor should support some kind of clang-format plugin to make it easy. If it doesn't, you're most likely using a bad editor. Switch to `vim`. Now that I think about it, switch to vim anyway, coz it's awesome!

include/pisa/wand_data_range.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+        float add_sequence(binary_freq_collection::sequence const &term_seq,
+                           binary_freq_collection const &coll,
+                           std::vector<float> const &norm_lens,
+                           [[maybe_unused]] float lambda)
+        {
+            float max_score = 0.0f;
+
+            std::vector<float> b_max(blocks_num, 0.0f);
+            for (auto i = 0; i < term_seq.docs.size(); ++i) {
```
amallia, line 68: @elshize should we change this one too?

src/create_wand_data.cpp
amallia: Apparently only one `excludes` is enough https://github.com/CLIUtils/CLI11/pull/93/files

amallia:
Everything is in place! We can merge once Travis finishes.

PR #78, opened 2019-01-25
Add fmt and timer

elshize:


amallia:
Timer seems to be in the wrong folder.

include/timer.hpp
```diff
@@ -0,0 +1,107 @@
+#pragma once
+
+#include <chrono>
+#include <functional>
+#include <string>
+
+#include <fmt/format.h>
```
amallia: Can we use quotes instead of angle brackets?

elshize:
> Timer seems to be in the wrong folder.

Oops.

amallia (approved):

include/timer.hpp
elshize: We can. But that's something I've been meaning to ask you: why do you use quotes for libraries? It's like saying "it is somewhere here" instead of "it is on a system include path", which I think is where libraries should be.

Issue #81, opened 2019-01-26
Support using string terms with forward index


elshize:
For example, if we pass `--terms <file>` or something of the sort, we assume that input is terms to be later translated to IDs.

While we're at it, a program to map terms to IDs in a file would be useful as well.

- [ ] ~Lexicon and lexicon tools #89~ will not do it for now
- [ ] Support lexicon for `queries`, `thresholds`, etc. with `std::unordered_map`

PR #89, opened 2019-01-28
Lexicon_View and tools to build and lookup

elshize:
Part of #81


amallia:
I just had a very quick look at it. Shouldn't the lexicon be generated from the "parsing module" directly?

elshize:
> I just had a very quick look at it. Shouldn't the lexicon be generated from the "parsing module" directly?

Not sure what that means, can you elaborate?

amallia:
Closing this one. Feel free to re-open it later.

PR #92, opened 2019-01-29
Add range-v3 and remove `enumerate` with `view::iota`

### LINKED ISSUE #90: Add range-v3 dependency
[range-v3](https://github.com/ericniebler/range-v3) should be added as dependency in cmake.

**Motivation**
We've talked about using this in a big way but I think we should start with just adding it and using it from that point forward improve our code base. **Simple** examples include:
```cpp
for (auto [idx, value] : enumerate(std::vector<int>{5, 5, 2, 1})) {}
for (auto [left, right] : zip(...) {}
```

**Caution**
We should exercise caution when using these in performance-sensitive places. I'm not saying it's inefficient, but we should make sure it is before merging such a thing.

elshize:
Fixes #90


amallia (approved):

amallia:
LGTM

PR #97, opened 2019-01-30
Terms with map

### LINKED ISSUE #81: Support using string terms with forward index

elshize:
Fixes #81


amallia (approved):
LGTM, but please remove `target_include_directories(queries PRIVATE ${CMAKE_SOURCE_DIR})` and `target_include_directories(thresholds PRIVATE ${CMAKE_SOURCE_DIR})`

PR #106, opened 2019-02-15
Sharding forward index

elshize:


elshize:
@amallia Ready for review.

include/pisa/invert.hpp
```diff
@@ -383,7 +286,8 @@ void merge_batches(std::string const &output_basename, uint32_t batch_count, uin
 
     std::ofstream dos(output_basename + ".docs");
     std::ofstream fos(output_basename + ".freqs");
-    write_sequence(dos, gsl::make_span<uint32_t const>(&term_count, 1));
+    auto document_count = static_cast<uint32_t>(document_sizes.size());
```
elshize, line 285: Note: this was a bug, wrong value was stored before.

elshize:
This is mostly ready for a review. I included docs. Note that I mention a script that doesn't exist yet. I'll push that soon and add usage to docs.

elshize:
Just for posterity's sake: partitioning CW09B into 123 shards on `moa` took about 45 min using 8 threads. Can't do much better, the bottleneck is I/O, but this is reasonable I think.

include/pisa/vector.hpp
```diff
@@ -0,0 +1,164 @@
...
+    reference at(K key) { return std::vector<V, Allocator>::at(static_cast<size_type>(key)); }
+    const_reference at(K key) const
+    {
+        return std::vector<V, Allocator>::at(static_cast<size_type>(key));
+    }
+
+    std::vector<V> const &as_vector() const { return *this; }
+
+    //auto entries()
+    //{
```
amallia: Please remove

include/pisa/vector.hpp
```diff
@@ -0,0 +1,164 @@
...
+namespace pisa {
+
+/// A (potentially) type-safe vector.
+///
+/// It derives from `std::vector<V>` and works essentially like one.
+/// The difference is that you define a key type as well, the outcome
+/// being that if you use a strong type key, you can differentiate between
+/// `Vector<IndexType_1, V>` and `Vector<IndexType_2, V>`.
+template <typename K, typename V = K, typename Allocator = std::allocator<V>>
+class Vector : protected std::vector<V> {
```
amallia: Do we need this? This feel like we are going to create one for each container in STL...
Idk, let me think about it, but you know me....

elshize: We absolutely do! Of course, we don't have to use it everywhere, but in some parts when we want to make sure we're type safe, this is great.

For instance, we have different mappings: documents to shards, shards to documents, terms to strings, etc., in the same chunk of code, it's very easy to lose track of what is what. Once you have type safety, you can't make an easy mistake. `std::vector` doesn't give you this.

As for the other containers, `vector` is a specific case. The reason is that here we're targeting the key that is not part of a vector type. But if we use, say, `map` it's already good because the key type is a template param.

So this vector is essentially used when we have a strongly typed integer keys. In other words, we model a total function `f: [0, n) -> V`. Alternatively, you could consider it an optimization of a map for all consecutive values between 0 and n (in `K`) but I think it's clearer to call it a vector (because that's what it in fact is, just with a strongly typed key).

Now, there might be other STL containers that need something like this, but only if we actually need to use them this way. Which we probably don't. Examples: `list` isn't even random access so it doesn't fit our use case of mapping; similarly queues are not used like that; or stacks... So that leaves the lonely vector.

elshize: I've renamed it `VecMap` for a more descriptive name.

elshize:
@amallia Can we try merging this to avoid continuous updates of this branch? Most of this is rather independent of anything else anyway...

amallia:
> @amallia Can we try merging this to avoid continuous updates of this branch? Most of this is rather independent of anything else anyway...

Totally, I would like to get also @JMMackenzie opinion of this PR. 
Also, @JMMackenzie what do you think of the `type_safe` file? This is broader decision in terms of design we want to follow in the future

JMMackenzie:
Looks great to me. I have had a look at the type safe stuff, pretty neat. Would the plan be to use this everywhere? 

PS, thoughts on this: https://github.com/foonathan/type_safe

Perhaps we could use as a dep instead, unless I'm missing something.

elshize:
Definitely not everywhere, but in as many places as it makes sense. We'll have to establish what that means.

We are also aware of this library, I've used it before and it's very good. But that's another dependency so I just used this for now. But I'd like to introduce the lib at some point in another PR if everyone's in board.

elshize:
@amallia can we merge then?

PR #102, opened 2019-02-05
Posting ranges and scored adaptor

elshize:

PR #108, opened 2019-02-17
Range query

amallia:
Heavily based on previous work from @elshize from https://github.com/pisa-engine/pisa/pull/102

### DIFF
--- a/include/pisa/cursor/block_max_scored_cursor.hpp
+++ b/include/pisa/cursor/block_max_scored_cursor.hpp
@@ -1,12 +1,12 @@
 #pragma once
 
 #include <vector>
-#include "scorer/bm25.hpp"
 #include "query/queries.hpp"
+#include "scorer/bm25.hpp"
 
 namespace pisa {
 
-template <typename Index, typename WandType>
+template <typename Index, typename WandType, typename Scorer>
 struct block_max_scored_cursor {
     typedef typename Index::document_enumerator     enum_type;
     typedef typename WandType::wand_data_enumerator wdata_enum;
@@ -14,15 +14,17 @@ struct block_max_scored_cursor {
     enum_type  docs_enum;
     wdata_enum w;
     float      q_weight;
+    Scorer     scorer;
     float      max_weight;
 };
 
 template <typename Index, typename WandType, typename scorer_type = bm25>
 [[nodiscard]] auto make_block_max_scored_cursors(Index const &index, WandType const &wdata,
-                                           term_id_vec terms) {
+                                                 term_id_vec terms) {
     auto query_term_freqs = query_freqs(terms);
+    using Scorer = Score_Function<scorer_type, WandType>;
 
-    std::vector<block_max_scored_cursor<Index, WandType>> cursors;
+    std::vector<block_max_scored_cursor<Index, WandType, Scorer>> cursors;
     cursors.reserve(query_term_freqs.size());
 
     for (auto term : query_term_freqs) {
@@ -30,8 +32,8 @@ template <typename Index, typename WandType, typename scorer_type = bm25>
         auto w_enum   = wdata.getenum(term.first);
         auto q_weight = scorer_type::query_term_weight(term.second, list.size(), index.num_docs());
         auto max_weight = q_weight * wdata.max_term_weight(term.first);
-        cursors.push_back(
-            block_max_scored_cursor<Index, WandType>{std::move(list), w_enum, q_weight, max_weight});
+        cursors.push_back(block_max_scored_cursor<Index, WandType, Scorer>{
+            std::move(list), w_enum, q_weight, {q_weight, wdata}, max_weight});
     }
     return cursors;
 }
--- a/include/pisa/cursor/block_max_scored_cursor.hpp
+++ b/include/pisa/cursor/block_max_scored_cursor.hpp
@@ -8,8 +8,8 @@ namespace pisa {
 
 template <typename Index, typename WandType, typename Scorer>
 struct block_max_scored_cursor {
-    typedef typename Index::document_enumerator     enum_type;
-    typedef typename WandType::wand_data_enumerator wdata_enum;
+    using enum_type  = typename Index::document_enumerator;
+    using wdata_enum = typename WandType::wand_data_enumerator;
 
     enum_type  docs_enum;
     wdata_enum w;
@@ -18,23 +18,25 @@ struct block_max_scored_cursor {
     float      max_weight;
 };
 
-template <typename Index, typename WandType, typename scorer_type = bm25>
+template <typename Index, typename WandType>
 [[nodiscard]] auto make_block_max_scored_cursors(Index const &index, WandType const &wdata,
                                                  term_id_vec terms) {
     auto query_term_freqs = query_freqs(terms);
-    using Scorer = Score_Function<scorer_type, WandType>;
+    using scorer_type     = bm25;
+    using Scorer          = Score_Function<scorer_type, WandType>;
 
     std::vector<block_max_scored_cursor<Index, WandType, Scorer>> cursors;
     cursors.reserve(query_term_freqs.size());
-
-    for (auto term : query_term_freqs) {
-        auto list     = index[term.first];
-        auto w_enum   = wdata.getenum(term.first);
-        auto q_weight = scorer_type::query_term_weight(term.second, list.size(), index.num_docs());
-        auto max_weight = q_weight * wdata.max_term_weight(term.first);
-        cursors.push_back(block_max_scored_cursor<Index, WandType, Scorer>{
-            std::move(list), w_enum, q_weight, {q_weight, wdata}, max_weight});
-    }
+    std::transform(query_term_freqs.begin(), query_term_freqs.end(), std::back_inserter(cursors),
+                   [&](auto &&term) {
+                       auto list       = index[term.first];
+                       auto w_enum     = wdata.getenum(term.first);
+                       auto q_weight   = scorer_type::query_term_weight(term.second, list.size(),
+                                                                      index.num_docs());
+                       auto max_weight = q_weight * wdata.max_term_weight(term.first);
+                       return block_max_scored_cursor<Index, WandType, Scorer>{
+                           std::move(list), w_enum, q_weight, {q_weight, wdata}, max_weight};
+                   });
     return cursors;
 }
 
--- a/include/pisa/cursor/block_max_scored_cursor.hpp
+++ b/include/pisa/cursor/block_max_scored_cursor.hpp
@@ -1,0 +1,37 @@
+#pragma once
+
+#include <vector>
+
+namespace pisa {
+
+template <typename Index, typename WandType>
+struct block_max_scored_cursor {
+    typedef typename Index::document_enumerator     enum_type;
+    typedef typename WandType::wand_data_enumerator wdata_enum;
+
+    enum_type  docs_enum;
+    wdata_enum w;
+    float      q_weight;
+    float      max_weight;
+};
+
+template <typename Index, typename WandType, typename scorer_type = bm25>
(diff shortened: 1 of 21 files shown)


amallia:
@elshize please review. It is just a first step, but it includes interesting changes. 
I have verified performance and correctness. 

After merging it, I will continue cleaning up the algorithms doing something similar to what you have done. I still prefer to address it in a separate PR.

elshize:
Why exactly do you want to do this in many PR's? This means more testing for efficiency, more code reviews, and more coding. We're just doing what we've already done. This is a lot of code, going through it again and again and again is not only annoying, but also wasteful.

Didn't you say you already got good performance? After all that work you did, do you really wanna scratch that and do it over?

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "scorer/bm25.hpp"
+
+namespace pisa {
+
+template <typename Index, typename WandType, typename Scorer>
+struct block_max_scored_cursor {
+    typedef typename Index::document_enumerator     enum_type;
+    typedef typename WandType::wand_data_enumerator wdata_enum;
+
+    enum_type  docs_enum;
```
elshize, line 14: How about adding `next()`, `next_geq()`, `docid()` and `freq()` that will call `docs_enum.next()`, etc? If you add `always_inline`, then you're surely getting the same level of indirection as now, only with better encapsulation and a unified interface across all cursors (remember that a non-scored cursor is called directly whereas scored indirectly, yet both are called `cursor`, which suggests similar behavior).

The same goes for `score()`, just make a member function that will do `ordered_cursors[i]->w.score() * ordered_cursors[i]->q_weight` behind the scenes. The only differences are in where this piece of code and how clear it is from the point of the algorithm. Also, less code repetition, fewer bugs.

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <vector>
+#include "query/queries.hpp"
+#include "scorer/bm25.hpp"
+
+namespace pisa {
+
+template <typename Index, typename WandType, typename Scorer>
+struct block_max_scored_cursor {
+    typedef typename Index::document_enumerator     enum_type;
```
elshize: We should use `using enum_type = ` syntax in any new code:
1. It's more intuitive, works as any other thing in the language, i.e., whatever you define is on the left.
2. It's more generic, literally. It works with templates, so whenever we need templated typedefs, we have to use it, which introduces unnecessary inconsistencies with other `typedef`s.

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -0,0 +1,41 @@
...
+    typedef typename WandType::wand_data_enumerator wdata_enum;
+
+    enum_type  docs_enum;
+    wdata_enum w;
+    float      q_weight;
+    Scorer     scorer;
+    float      max_weight;
+};
+
+template <typename Index, typename WandType, typename scorer_type = bm25>
```
elshize: Defining `scorer_type = bm25` as the last param is a bit problematic. Here's why:

Let's say you want to construct `bm25` cursors, you type this:
```cpp
make_block_max_scored_cursors(index, wdata, terms);
```
Swell. Now, with a non-default scorer:
```cpp
make_block_max_scored_cursors<Index, Wand, QLikelihood>(index, wdata, terms);
```
You need to define all types, which means you might as well just use the constructor and define everything explicitly.

I see two possible designs:

1. No default scorer: because why would BM25 be more default than anything else? Then you can simply move the type param to the front and always get cursors this way:
```cpp
make_block_max_scored_cursors<bm25>(index, wdata, terms);
```
2. You use scorer "tags". (Note that this can be used with or without defaults.)
```cpp
struct BM25_Tag {
    auto operator()(auto index, auto term, auto term_weight) { /* impl */ };
} bm25;
```
I used `auto` instead of templates to be more concise, you get the idea. Then, the signature would be:
```cpp
template <typename Index, typename WandType, typename ScorerTag>
[[nodiscard]] auto make_block_max_scored_cursors(Index const &index, WandType const &wdata,
                                                 term_id_vec terms, ScorerTag tag) {
    ...
    auto score_function = tag(index, term, term_weight);
    ...
}
```
And then you call it like this:
```cpp
make_block_max_scored_cursors(index, wdata, terms, bm25);
```
Awesome!

include/pisa/cursor/cursor.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <vector>
+#include "query/queries.hpp"
+
+namespace pisa {
+
+template <typename Index>
+[[nodiscard]] auto make_cursors(Index const &index, term_id_vec terms) {
+    auto query_term_freqs = query_freqs(terms);
+    typedef typename Index::document_enumerator cursor;
```
elshize: `using`

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -0,0 +1,41 @@
...
+[[nodiscard]] auto make_block_max_scored_cursors(Index const &index, WandType const &wdata,
+                                                 term_id_vec terms) {
+    auto query_term_freqs = query_freqs(terms);
+    using Scorer = Score_Function<scorer_type, WandType>;
+
+    std::vector<block_max_scored_cursor<Index, WandType, Scorer>> cursors;
+    cursors.reserve(query_term_freqs.size());
+
+    for (auto term : query_term_freqs) {
```
elshize: This looks like a perfect place to use ranges:
```cpp
auto cursors = ranges::view::transform(query_terms_freqs, [](auto &&term) {
    ...
    return cursor_type{...};
}) | ranges::to_vector;
```

include/pisa/cursor/cursor.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    std::vector<cursor> cursors;
+    cursors.reserve(query_term_freqs.size());
```
elshize: This place is even better for ranges.

include/pisa/cursor/max_scored_cursor.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <vector>
+#include "scorer/bm25.hpp"
+#include "query/queries.hpp"
+
+namespace pisa {
+
+template <typename Index, typename Scorer>
```
elshize, line 9: See all comments about `block_max_scored_cursor`

include/pisa/cursor/scored_cursor.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <vector>
+#include "scorer/bm25.hpp"
+#include "query/queries.hpp"
+#include "scorer/score_function.hpp"
+
+namespace pisa {
+
+template <typename Index, typename Scorer>
+struct scored_cursor {
```
elshize, line 11: See all comments about `block_max_scored_cursor`.

include/pisa/query/algorithm/and_query.hpp
```diff
@@ -5,60 +5,59 @@
...
                     i         = 0;
                     break;
                 }
             }
 
-            if (i == enums.size()) {
+            if (i == ordered_cursors.size()) {
                 results += 1;
 
                 if (with_freqs) {
```
elshize: `if constexpr (with_freqs)`

include/pisa/query/algorithm/and_query.hpp
```diff
@@ -5,60 +5,59 @@
...
+template <bool with_freqs>
 struct and_query {
 
-    and_query(Index const &index) : m_index(index) {}
+    and_query(uint64_t max_docid) : m_max_docid(max_docid) {}
 
-    uint64_t operator()(term_id_vec terms) const {
-        if (terms.empty())
+    template<typename Cursor>
+    uint64_t operator()(std::vector<Cursor> &&cursors) const {
```
elshize: So if we use `&&` on a collection, we could as well do:
```cpp
uint64_t operator()(CursorRange &&cursors) const
{
    using cursor_type = CursorRange::value_type;
    ...
}
```

include/pisa/query/algorithm/or_query.hpp
```diff
@@ -5,44 +5,36 @@
...
-        while (cur_doc < m_index.num_docs()) {
+        while (cur_doc < m_max_docid) {
             results += 1;
-            uint64_t next_doc = m_index.num_docs();
-            for (size_t i = 0; i < enums.size(); ++i) {
-                if (enums[i].docid() == cur_doc) {
+            uint64_t next_doc = m_max_docid;
+            for (size_t i = 0; i < cursors.size(); ++i) {
+                if (cursors[i].docid() == cur_doc) {
                     if (with_freqs) {
```
elshize: `if constexpr`

include/pisa/query/algorithm/ranked_or_taat_query.hpp
```diff
@@ -9,47 +9,38 @@
...
         }
-        m_accumulators.init();
-        for (uint32_t term = 0; term < cursors.size(); ++term) {
-            auto cursor = cursors[term];
-            const auto score = score_functions[term];
-            for (; cursor.docid() < m_accumulators.size(); cursor.next()) {
-                m_accumulators.accumulate(cursor.docid(), score(cursor.docid(), cursor.freq()));
+        m_accumulator.init();
+
+        for(auto&& cursor : cursors) {
```
elshize: Format the entire loop.

include/pisa/query/algorithm/wand_query.hpp
```diff
@@ -98,14 +75,14 @@ struct wand_query {
             } else {
                 // no match, move farthest list up to the pivot
                 uint64_t next_list = pivot;
-                for (; ordered_enums[next_list]->docs_enum.docid() == pivot_id; --next_list)
+                for (; ordered_cursors[next_list]->docs_enum.docid() == pivot_id; --next_list)
                     ;
```
elshize: A hanging semicolon like that is giving me a heart attack. Just `for (...) {}`

src/profile_queries.cpp
```diff
@@ -86,17 +90,32 @@ void profile(const std::string index_filename,
 
     for (auto const& t: query_types) {
         spdlog::info("Query type: {}", t);
+        std::function<uint64_t(term_id_vec)> query_fun;
         if (t == "and") {
-            op_profile(and_query<typename add_profiling<IndexType>::type, false>(index), queries);
+            query_fun = [&](term_id_vec terms){
+                and_query<false> and_q(index.num_docs());
+                remove_duplicate_terms(terms);
```
elshize: Won't `make_cursors` already remove duplicate terms?

test/test_ranked_queries.cpp
```diff
@@ -7,19 +7,22 @@
...
+#include "cursor/scored_cursor.hpp"
+#include "cursor/max_scored_cursor.hpp"
+#include "cursor/block_max_scored_cursor.hpp"
 
 namespace pisa { namespace test {
 
     struct index_initialization {
 
-        typedef single_index index_type;
+        typedef single_index IndexType;
```
elshize: `using`

elshize (changes requested):

include/pisa/cursor/block_max_scored_cursor.hpp
amallia: I will address this in the next PR. I am taking a note.

include/pisa/cursor/block_max_scored_cursor.hpp
amallia: It will be done once we introduce other scorers

elshize (approved):
Looks good.

PR #122, opened 2019-02-28
Cli lambda

JMMackenzie:
Added a small change which allows lambda to be set via command line arguments instead of directly out of the config file (avoids recompilation if searching for the right block size on a new collection).

Also added a small bit of text to the documentation.


amallia (approved):
Thank you

PR #123, opened 2019-02-28
Query algo

YuhongZhang98:


include/pisa/query/algorithm/range_query.hpp
```diff
@@ -0,0 +1,52 @@
...
+        m_topk.clear();
+        if (cursors.empty()) {
+            return 0;
+        }
+
+        size_t end = m_range_size;
+        for (; end <= m_max_docid; end += m_range_size) {
+            processRange(cursors, end);
+        }
+        if (end < m_max_docid + m_range_size) {
```
amallia: Tip: `if` statements are bad for efficiency! if you can avoid them, you should do it.
If you want to know more, look for `if branch misprediction`.

That said, here we can avoid this check.
```cpp
// Loop excluding always the very last range 
for (size_t end = m_range_size; end + m_range_size <= m_max_docid; end += m_range_size) {
   processRange(cursors, end);
}
// Always process the last one
processRange(cursors, m_max_docid);
```

include/pisa/query/algorithm/wand_query.hpp
```diff
@@ -17,7 +17,7 @@ struct wand_query {
 
     template<typename CursorRange>
     uint64_t operator()(CursorRange &&cursors) {
-        using Cursor = typename CursorRange::value_type;
+        using Cursor = typename std::decay_t<CursorRange>::value_type;
```
amallia, line 20: Can we change this for all the other algorithms?

test/test_ranked_queries.cpp
```diff
@@ -136,6 +136,16 @@ TEST_CASE_METHOD(pisa::test::index_initialization, "ranked_or_taat_lazy")
     test_against_or(query_fun);
 }
 
+TEST_CASE_METHOD(pisa::test::index_initialization, "range_query")
+{
+    auto query_fun = [&](pisa::term_id_vec terms){
+        pisa::range_query<pisa::wand_query> wand_q(10, index.num_docs(), 128);
```
amallia: Can you add a test for each of the algorithms we have?
So one for `block_max_wand`, another one  for `maxscore` etc.

amallia (changes requested):

include/pisa/query/algorithm/range_query.hpp
elshize: Not sure if that's gonna have much impact here, but Antonio's way is definitely cleaner. And a good tip in general.

include/pisa/query/algorithm/wand_query.hpp
elshize: We should. We must.

test/test_ranked_queries.cpp
elshize: We could do something smart in here (I'm writing off the top of my head, please consult documentation if decide to do it this way):
```cpp
TEST_CASE_METHOD(pisa::test::index_initialization, "range_query")
{
    // normally, we wouldn't erase for efficiency reasons, but here only testing correctness
    // BTW. in the coming versions of Catch2, there will be a template generator, so no erasure will be necessary
    using fn_type = std::function<uint64_t(decltype(make_block_max_scored_cursors(index, wdata, pisa::term_id_vec{})))>;
    auto query_type = GENERATE(fn_type, pisa::wand_query, pisa::maxscore_query, etc.);
    CAPTURE(query_type);
    auto query_fun = [&](pisa::term_id_vec terms){
        pisa::range_query<fn_type> q(10, index.num_docs(), 128);
        q(make_block_max_scored_cursors(index, wdata, terms));
        return q;
    };
    test_against_or(query_fun);
}
```

But if you're having problems with this, just copy/paste and modify.

elshize (commented):
One additional note: we use underscores in function names throughout the codebase, like `process_range`, so I think we should stick to that.

We're less consistent with class naming (I'm guilty of this myself), which we probably should agree on at some point and also stick to.

test/test_ranked_queries.cpp
amallia: Yes, this is much nicer. Let's do it this way pls.

elshize:
Changes look good, try adding that test then.

elshize:
I took a look at `GENERATOR` for that test, and it turns out it's more tricky for two reasons:
1. A query algorithm is not an `std::function` because it also has `topk()` member function that we use.
2. We create new algorithms in range algorithm so can't type-erase.

As much as 1 could be fixed with `boost.te` library, the second is much more tricky. And given that we will change interface to return top-k (no member function), and will probably use one algorithm object instead of creating new ones (when max docid will go to operator)---I don't see much sense in trying to be smart here.

Therefore, I added individual tests for each algorithm (and fixed BMW which was looping indefinitely).

Also, used clang-format on the header.

@amallia please take a look.

elshize (approved):
LKTM: @amallia?

amallia (approved):

PR #124, opened 2019-02-28
Pass max_docid in call operator

amallia:


include/pisa/query/algorithm/and_query.hpp
```diff
@@ -9,13 +9,19 @@ namespace pisa {
 template <bool with_freqs>
 struct and_query {
 
-    and_query(uint64_t max_docid) : m_max_docid(max_docid) {}
-
     template<typename CursorRange>
-    uint64_t operator()(CursorRange &&cursors) const {
+    auto operator()(CursorRange &&cursors, uint32_t max_docid) const {
```
elshize, line 13: It kinda bothers me we can't use a default value there if we want to process the entire range. But more importantly, what happens when you pass a value larger than document count? Won't it just loop forever?

So we got to a situation when you might not care about any range related stuff (just want to simply run entire thing) yet still must provide a value that can easily cause infinite loops. And there's no way to check if `max_docid` is valid within the function.

We could try hacking it but making cursors go to `std::numeric_limits<uint32_t>::max()` would fix the whole thing:
- can omit the argument, we can deduce max if not provided,
- never loops infinitely -- a huge gain,
- it's more semantically sound: why setting a max ID larger than any possible not work in principle? All your IDs are smaller, so I'd expect them to be just processed, and others ignored (coz don't exist).

How difficult would it be to make that little change?

amallia: It is easy and I also think we have to do it. Let's do it is a separate PR as this one is to improve the query algorithms API.

elshize: Ok. Can we do the rest of the algorithms then?

elshize (approved):
Looks good. One thing that you might consider once we make max_docid optional argument is to move the accumulator to the second argument.

PR #133, opened 2019-03-04
Fix #128

### LINKED ISSUE #128: binary_collection move invalidates iterator
**Describe the bug**
Moving `binary_collection` invalidates any iterator pointing to its elements. This is because we store a pointer to it in the iterator structure. The fix is to store data pointer and size instead.

amallia:
I clang-format'd it. So it looks worst than it actually is...

### DIFF
--- a/include/pisa/binary_collection.hpp
+++ b/include/pisa/binary_collection.hpp
@@ -11,131 +11,129 @@
 
 #include "util/util.hpp"
 
-#if defined (__unix__) || (defined (__APPLE__) && defined (__MACH__))
+#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
 #include <sys/mman.h>
 #endif
 
 namespace pisa {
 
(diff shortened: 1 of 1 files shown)


include/pisa/binary_collection.hpp
```diff
@@ -11,131 +11,129 @@
...
+            // file might be truncated
+            n          = std::min(n, size_t(m_data_size - pos));
+            auto begin = &m_data[pos];
+
+            m_next_pos = pos + n;
+            m_cur_seq  = S(begin, begin + n);
+        }
+
+        const pointer m_data;
+        size_t        m_data_size = 0, m_pos = 0, m_next_pos = 0;
```
elshize, line 127: I would avoid this in the future, it's inconsistent where to look for variable names: if we have a set of lines, then let's stick to it.

elshize (approved):
Looks good.

PR #135, opened 2019-03-04
Better CLI

### LINKED ISSUE #93: Easier use of standard command line options

Some cmd options will repeat throughout many tools, e.g., stem/nostem or index type. I suggest implementing some shortcut to defining these, so we don't have to repeatedly define a variable and then define the option. For one, it's annoying, and for two, any changes we want to make to the same option need to be manually propagated.

**Additional context**

I believe you can inherit from `App` to define custom bases, but that's limited (as inheritance is).

I've done something in `irkit` that we might reuse or adapt. For each commonly used option, you'd define something like this:
```cpp
struct k_opt {
    int k = 1000;

    template<class Args>
    void set(CLI::App& app, Args& args)
    {
        app.add_option("-k", args->k, "Number of documents to retrieve", true);
    }
};
```

Then, you'd use it like this:
```cpp
auto [app, args] = irk::cli::app(
        "Query inverted index",
        index_dir_opt{},
        nostem_opt{},
        id_range_opt{},
        score_function_opt{with_default<std::string>{"bm25"}},
        traversal_type_opt{with_default<Traversal_Type>{Traversal_Type::DAAT}},
        k_opt{},
        trec_run_opt{},
        trec_id_opt{},
        terms_pos{optional});
/* Potentially adding other custom options to app */
CLI11_PARSE(*app, argc, argv);
/* Parsed arguments are in args, e.g., args->index_dir */
```

elshize:
Fixes #93 
Takes care of several things.
1. Returning value where it's defined: cleaner code and more C++. At the same time, we have the same abilities to access underlying stuff and use it as we used to (to make sure we don't remove any features).
2. Defining some common stuff in a common header so not to duplicate code and have consistent options.
3. Providing easy way to do stuff like `check`, `required`, and `needs`.
4. Conversion to `bool` to check if option defined.
5. Explicit definition of default values.

### DIFF
--- a/src/cli.hpp
+++ b/src/cli.hpp
@@ -6,9 +6,17 @@
 
 #include <CLI/CLI.hpp>
 #include <CLI/Option.hpp>
+#include <CLI/Optional.hpp>
 #include <boost/filesystem.hpp>
 
 namespace pisa {
+
+template <typename T>
+struct is_optional : std::false_type {};
+
+template <typename T>
+struct is_optional<std::optional<T>> : std::true_type {
+};
 
 template <typename T>
 std::ostream &operator<<(std::ostream &os, std::optional<T> const &value)
@@ -53,12 +61,6 @@ class Option {
 
     explicit operator bool() const { return static_cast<bool>(*option_); }
 
-    template <class = std::enable_if_t<not std::is_same_v<T, bool>>>
-    operator T() const
-    {
-        return *value_;
-    }
-
     [[nodiscard]] auto opt() -> CLI::Option * { return option_; }
 
    private:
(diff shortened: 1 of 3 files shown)


elshize:
@amallia take a look and give me your opinion before I proceed any further.

amallia:
> @amallia take a look and give me your opinion before I proceed any further.

Interesting approach. Reviewing it now. Can we remove the `WIP` tag then?

elshize:
> > @amallia take a look and give me your opinion before I proceed any further.
> 
> Interesting approach. Reviewing it now. Can we remove the `WIP` tag then?

I was actually thinking of switching all the programs to the new method once it's reviewed and agreed upon. Hence `wip`.

JMMackenzie:
@elshize - Is this addressed in #313 and if yes, can we close off #93 too?

elshize:
Yes, this would be obsolete with #313.

amallia:
Replaced by https://github.com/pisa-engine/pisa/pull/313

PR #136, opened 2019-03-07
Support trecweb record format

### LINKED ISSUE #126: Gov2 parsing
**Summary**
We need to be able to parse GOV2 collection.

**Additional context**

We need to be able to use another format, say, `trec` in this function https://github.com/pisa-engine/pisa/blob/master/src/parse_collection.cpp#L15

**Format**
As far as I can tell (but needs to be verified) the record format is:
```
<DOC>
<DOCNO>TRECID</DOCNO>
<DOCHDR>...</DOCHDR>
HTML CONTENT
</DOC>
```
So it's simple enough, we might be able to get by without any XML parsing. We have to:
1. Extract TRECID as title
2. Extract URL from inside of `DOCHDR` (I think it's always first line)
3. And HTML CONTENT as content.

elshize:
Fixes #126

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,3 +61,6 @@
 [submodule "external/te"]
 	path = external/te
 	url = https://github.com/boost-experimental/te.git
+[submodule "external/trecpp"]
+	path = external/trecpp
+	url = git@github.com:pisa-engine/trecpp.git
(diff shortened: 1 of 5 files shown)


amallia:
LGTM

PR #140, opened 2019-03-08
Implement payload vector

### LINKED ISSUE #139: Lexicon revisited
**Summary**
Loading entire document and term vectors/hash maps is slow and annoying (it takes about a minute to load it!). We need a better solution.

This revisits #81

**Additional context**
For simplicity's sake, I suggest again a payload-based string map that can be memory-mapped:

```
[offset, offset, ..., offset] [payload, payload, ..., payload]
```

Offsets are `int`s and payloads are bytes (need not to be `string`s, could be anything we know how to store).

Now, our terms are always sorted, so a quick binary search will take care of mapping from term to ID. Documents are not sorted in general, but **we don't need to map strings to IDs here, only IDs to strings**. So it works.

elshize:
Fixes #139


amallia:
Once all the conflicts are solved, can you give it a quick manual test so we will merge it?
I would like to make sure that the tasks in the `src` folder actually work well with this. Can you also quickly check that it does not downgrade performance?

elshize:
Will do. Speaking of, we should have some tests for those binaries as well. Any good ideas?

amallia:
Integration tests with python :)

elshize:
> Integration tests with python :)

How about Rust?

elshize:
Ok, can you double check this after all that merging? It's ready to be merged.

On Robust04, it's actually ~40% faster. It seems like resolving term IDs is actually a quite large portion of the run :|

This is not a great collection to run efficiency tests. Is this the largest of collections or are there any larger in the potentially tested at the workshop?

.travis.yml
```diff
@@ -55,16 +56,16 @@ matrix:
...
+      env: MATRIX_EVAL="CC=clang-6.0 && CXX=clang++-6.0 && COVERAGE=Off"
     - os: linux
       dist: xenial
       compiler: clang
       addons:
           apt:
               sources: *all_sources
-              packages: ['clang-6.0', 'libstdc++-7-dev']
-      env: MATRIX_EVAL="CC=clang-6.0 && CXX=clang++-6.0 && COVERAGE=Off"
+              packages: ['libstdc++-7-dev']
```
amallia: I think it should be: `packages: ['clang-8.0', 'libstdc++-8-dev']`

.travis.yml
```diff
@@ -55,16 +56,16 @@ matrix:
       addons:
           apt:
               sources: *all_sources
-              packages: ['clang-5.0', 'libstdc++-7-dev']
```
amallia: So we dont support Clang 5 anymore?

.travis.yml
```diff
@@ -55,16 +56,16 @@ matrix:
...
-      env: MATRIX_EVAL="CC=clang-5.0 && CXX=clang++-5.0 && COVERAGE=Off"
+              packages: ['clang-6.0', 'libstdc++-7-dev']
+      env: MATRIX_EVAL="CC=clang-6.0 && CXX=clang++-6.0 && COVERAGE=Off"
```
amallia: Is there a Clang 7?

elshize: Yes, let me check 8 first though.

.travis.yml
elshize: I feel like if we just keep adding more instances, we'll have to wait forever for each build. So I suggest to remove it at least for now until we figure out how to speed up CI, e.g., until we get a server or something. What do you think?

.travis.yml
elshize: It's actually `clang-8` (I know, weird), and already fixed.

elshize:
@amallia take a look, fixed that one test and merged everything. Merge to master if looks okay.

amallia:
@elshize can you please reset `.travis.yml`? We can address that in a separate PR.

amallia:
Can we build the lexicon while parsing?

include/pisa/payload_vector.hpp
```diff
@@ -0,0 +1,346 @@
...
+    struct sizeofs<Head, Tail...> {
+        static const size_t value = sizeof(Head) + sizeofs<Tail...>::value;
+    };
+
+    template <typename T>
+    struct sizeofs<T> {
+        static const size_t value = sizeof(T);
+    };
+
+    struct test {
```
amallia, line 134: Should this be here?

amallia:
Overall this is a very nice PR. Clean and very well tested code. there are just a few comments I currently have.

include/pisa/payload_vector.hpp
elshize: Yes, I want to have a compile-time check for this, to make sure that these work as expected. It's in `detail` so I figure it's as good a place as any.

elshize:
> Can we build the lexicon while parsing?

I'm wondering whether we should create lexicons instead or in addition to text files... It seems a bit redundant if we keep both, especially since you can use `lexicon` tool to print entire list or parts of it on demand.

elshize:
@amallia I left the original text files, if we wat to get rid of them, let's do it in another PR. Check out my change, and let's merge soon.

PR #148, opened 2019-04-03
Fix issue with per-query timings

JMMackenzie:
I am pretty sure we need to divide through by the number of runs here (please double check this).


amallia (approved):

amallia:
Right! Thanks

@elshize pls check and merge

elshize (approved):
Good catch!

PR #149, opened 2019-04-04
Add heap score check to accumulation of top-k heap

JMMackenzie:
I spoke briefly to Michal regarding an issue I found where rarer terms would actually take longer to process in TaaT mode.

After some profiling, it became clear that when there aren't many accumulators touched (so, many tied scores), they were being pushed/popped in the heap. It seems that this is true for any tied score. 

This commit will fix this in the accumulators, but it's probably worth discussing if we should change the heap insert policy to require a "greater than current threshold" rather than a "greater than or equal to threshold" score for heap entry.

The line in question is here: https://github.com/pisa-engine/pisa/blob/49ebb8104d34bc71fe7611a02598ede3a7dffbb3/include/pisa/topk_queue.hpp#L24

### DIFF
--- a/include/pisa/accumulator/lazy_accumulator.hpp
+++ b/include/pisa/accumulator/lazy_accumulator.hpp
@@ -74,7 +74,7 @@ struct Lazy_Accumulator {
         for (auto const &block : m_accumulators) {
             int pos = 0;
             for (auto const &score : block.accumulators) {
-                if (block.counter(pos++) == m_counter) {
+                if (block.counter(pos++) == m_counter && topk.would_enter(score)) {
                     topk.insert(score, docid);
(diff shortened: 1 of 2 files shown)


amallia:
Is this one still giving improvements when compared to the more recent change?

JMMackenzie:
Reopening for examination based on the discussion in #158 - it seems that these checks definitely help. 

I can investigate further, but I am out of time for the next few days. We should also double check that the effectiveness is not impacted here.

elshize (approved):
Let's merge that. We don't need to concern ourselves with TAAT too much at this point, as we're not likely to use it for the workshop. Let's revisit this issue later when we have more time.

PR #154, opened 2019-04-09
Stopwords

elshize:
At least for GOV2, it doesn't seem to impact precision. If that's the case across all tested collections, we might wanna use it to speed up our processing (why go through long posting lists if not necessary?).

### DIFF
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -32,9 +32,10 @@ struct Query {
     std::vector<term_id_type> terms;
 };
 
-[[nodiscard]] inline auto parse_query(std::string const &query_string,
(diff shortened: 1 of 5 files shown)


amallia:
Is this removing stopwords only from queries? The index is still the same, isn’t it? If you remove the stopwords at index time the stats and consequently the BM25 model will also change slightly...

elshize:
True but the difference is negligible. There are 418 stopwords, the stats won't change much.

I'm not saying we shouldn't be able to remove them at indexing time (we should have that option) but removing them at query time is a more pragmatic approach.

amallia (approved):
Looks good, but let's merge it after https://github.com/pisa-engine/pisa/pull/156

amallia (approved):

Issue #150, opened 2019-04-04
Heap behavior for ties with threshold


JMMackenzie:
- If a document being added to the heap has a tie with the threshold of the heap, the current document will be added (and the document it was tied with will be thrown away).
- If we assume that two documents with the same score are equally good, it would be better to just leave the current document.
- I propose to modify the heap insert function to require that the incoming document score exceeds the threshold (rather than being greater-than-or-equal-to it).|

The line in question is here:  https://github.com/pisa-engine/pisa/blob/49ebb8104d34bc71fe7611a02598ede3a7dffbb3/include/pisa/topk_queue.hpp#L24 which we could update to `score <= m_threshold` to strengthen the heap inclusion check.

Also related to #149 which would not need to be merged if we opt to take this route. Opinions on this?

PR #158, opened 2019-04-10
Update the heap

JMMackenzie:
This will fix issue #150 and allow us to close off #149 without merging it (as this is the alternative solution).

The impact here is a more deterministic processing model, where a document can only make it into the top-k heap if its current score exceeds the heap threshold. This means that, given a query in which all document scores are tied, we would return the first k matching documents.

### DIFF
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -21,7 +21,7 @@ struct topk_queue {
     bool insert(float score) { return insert(score, 0); }
 
     bool insert(float score, uint64_t docid) {
-        if (PISA_UNLIKELY(score < m_threshold)) {
+        if (PISA_UNLIKELY(score <= m_threshold)) {
             return false;
         }
         m_q.emplace_back(score, docid);


amallia:
@JMMackenzie Can you take the timings again now and compare it with the previous one? Thanks

JMMackenzie:
@amallia Sure, just for the TaaT stuff?

amallia:
Correct. Thanks

JMMackenzie:
Sure, using k=10 on Gov2 with the 150 TREC topics 701-850, I get the following from the update:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140285, "q50": 131467, "q90": 189927, "q95": 201999}`

Compared to previously:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140994, "q50": 131512, "q90": 190844, "q95": 203007}`

Clearly, there is no real difference here.

Using k = 1000, we can see a little more of a difference.
Fixed:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140926, "q50": 132171, "q90": 190082, "q95": 202912}`

Previously:
`{"type": "opt", "query": "ranked_or_taat", "avg": 169428, "q50": 154762, "q90": 207383, "q95": 250648}`


Finally, showing a per-query example using some different length postings/combinations for k = 1000:

Queries are:
```
0: 1
1: 1000
2: 100000
3: 1000000
4: 1 1000
5: 1 100000
```
Note the improvements for query 2 and 3 (which is what we expect - as posting length decreases, we expect TaaT to do better).

Fixed:
```
qid	usec
0	261173
1	117068
2	102199
3	102877
4	275711
5	262361
```

Previously:
```
qid	usec
0	261900
1	129765
2	352133
3	664299
4	276163
5	262485
```

amallia:
The following is something you sent me previously. Is it still the case? I can see larger improvement there

PR 149 branch
```
Queries
10 100 1000 
10
100
1000
10000
1000000 2000000
```

```
With would_enter
0	205956
1	154851
2	104639
3	82788
4	69986
5	67581

VS

Master
0	236595
1	188309
2	138202
3	128899
4	244967
5	659882

```

JMMackenzie:
Interesting point, let me get back to you - possibly different experimental settings, I will double check it just to be sure!

EDIT: Note that the queries I used were different :-)

JMMackenzie:
OK, below I have also added the `would_enter` check on top of this PR, and I show the results from all three:

 k=10 on Gov2 with the 150 TREC topics 701-850:

This PR:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140285, "q50": 131467, "q90": 189927, "q95": 201999}`

Compared to previously:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140994, "q50": 131512, "q90": 190844, "q95": 203007}`

would_enter:
`{"type": "opt", "query": "ranked_or_taat", "avg": 107193, "q50": 98056, "q90": 156755, "q95": 169450}`

Using k = 1000, we can see a little more of a difference.
This PR:
`{"type": "opt", "query": "ranked_or_taat", "avg": 140926, "q50": 132171, "q90": 190082, "q95": 202912}`

Previously:
`{"type": "opt", "query": "ranked_or_taat", "avg": 169428, "q50": 154762, "q90": 207383, "q95": 250648}`

would_enter:
`{"type": "opt", "query": "ranked_or_taat", "avg": 109674, "q50": 99696, "q90": 158383, "q95": 170116}
`

Finally, showing a per-query example using some different length postings/combinations for k = 1000:

Queries are:
```
0: 1
1: 1000
2: 100000
3: 1000000
4: 1 1000
5: 1 100000
```
Note the improvements for query 2 and 3 (which is what we expect - as posting length decreases, we expect TaaT to do better).

This PR:
```
qid	usec
0	261173
1	117068
2	102199
3	102877
4	275711
5	262361
```

Previously:
```
qid	usec
0	261900
1	129765
2	352133
3	664299
4	276163
5	262485
```

would_enter check:
```
0	238909
1	82387
2	87401
3	109808
4	241500
5	230795
```

So, you are right - it looks like it's better to do the check as well. I'm not 100% sure why that's true at this case. It does save a call into the heap function, which avoids the `PISA_UNLIKELY()` check - perhaps we avoid some mispredictions on that? 

In any case, it's important to note that the test above was implemented on top of this PR. Maybe we should open the other PR and merge it too?

PR #161, opened 2019-04-10
Update invert.hpp

amallia:


elshize (approved):
Excellent!

elshize:
Well, I guess you need to fix that test, too. Seems like I was so sure it should be term count that I even wrote a test for it.

PR #165, opened 2019-04-13
Extract TREC topic

amallia:
Usage: 
`./bin/extract_topics -i [TREC query input file] -o Query output basename`

Example:
`./bin/extract_topics -i topics.robust04.301-450.601-700.txt -o robust04`

It produces 3 files:
`robust04.desc`  `robust04.narr`  `robust04.title`


include/pisa/util/string_util.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
elshize: Update `clang-format`

include/pisa/util/string_util.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <algorithm>
+#include <string>
+#include <utility>
+#include <locale>
+
+namespace pisa {
+namespace util {
+std::string ltrim(const std::string &&str) {
```
elshize: Why is the rvalue ref constant? You either want to move value (non-const) or copy, then use const ref.

I think we can either:
1. Take rvalue and return object.
2. Take const ref and return `string_view`. We can do that because the result will always point to a substring. Then, one can create a string from that view if a copy is desired. Otherwise, one can operate on a view.

There's also an option of implementing both interfaces if that makes sense.

include/pisa/util/string_util.hpp
```diff
@@ -0,0 +1,30 @@
...
+    auto it1     = std::find_if(trimmed.rbegin(), trimmed.rend(), [](char ch) {
+        return !std::isspace<char>(ch, std::locale::classic());
+    });
+    trimmed.erase(it1.base(), trimmed.end());
+    return trimmed;
+}
+std::string trim(const std::string &&str) { return ltrim(rtrim(std::forward<decltype(str)>(str))); }
+
+}  // namespace util
+}  // namespace pisa
```
elshize: We need unit tests for this.

include/pisa/util/string_util.hpp
elshize: This file is already in `util` dir so you may as well remove that `_util` from the file name.

I would prefer rather `algorithm/string.hpp` or something like that.

src/extract_topics.cpp
```diff
@@ -0,0 +1,124 @@
...
+    while (not is.eof()) {
+        if (is.peek() == std::istream::traits_type::eof() or pred(is.peek())) {
+            break;
+        }
+        os.put(is.get());
+    }
+    return os;
+}
+
+std::optional<Topic> next_topic(std::ifstream &is) {
```
elshize: Can we extract this to a header and write unit tests?

elshize (changes requested):

src/extract_topics.cpp
amallia: Yes! Will do.

elshize (approved):

PR #170, opened 2019-04-16
Select algorithm to use

amallia:
I am opening again this PR since I had to revert it in master as it was causing an issue.
Apparently this PR creates an issue, but I dont see how.

The final results set contain duplicate documents. If you see the error please let me know and I will fix it

### DIFF
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -1,22 +1,26 @@
 #include <iostream>
 
+#include <functional>
 #include "boost/algorithm/string/classification.hpp"
 #include "boost/algorithm/string/split.hpp"
 #include "boost/optional.hpp"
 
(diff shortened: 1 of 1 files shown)


JMMackenzie:
Was it definitely introduced in this commit? And is it occuring for all algorithms? 

I am just wondering because of the recent heap change that was made. Perhaps something unexpected is occuring there. Do tests pass without this commit?

amallia:
> Was it definitely introduced in this commit? And is it occuring for all algorithms?

Yes, I just checked and it is this specific commit.

> I am just wondering because of the recent heap change that was made. Perhaps something unexpected is occuring there. Do tests pass without this commit?

No, I don't think. The heap is not related to this problem.

amallia:
I am merging it because it seems the problem to be related to something else.

PR #171, opened 2019-04-16
Update default params for bmw25

amallia:
Recommended by Trotman et al. (SIGIR 2012 OSIR Workshop).
http://opensearchlab.otago.ac.nz/paper_4.pdf

### DIFF
--- a/include/pisa/scorer/bm25.hpp
+++ b/include/pisa/scorer/bm25.hpp
@@ -6,8 +6,8 @@
 namespace pisa {
 
     struct bm25 {
-        static constexpr float b = 0.5;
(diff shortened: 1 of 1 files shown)


JMMackenzie:
This will break current `wand` files that might be built using different parameters, right? Looks fine though, these parameters are commonly used and generally work well.

EDIT: If it is the case that changing the ranker will break the `wand` file, we should really add something to the file that defines which ranker it was built with, and then check at load-time whether it's correct. Otherwise, behavior will be undefined I think (skipping when you shouldn't or suboptimal skipping, for example).

PR #172, opened 2019-04-17
Update ranked_or_query.hpp

amallia:
The reason with `ranked_or` is failing with `evaluate_queries` is because we are not storing the document ID, so the result set consists of document ID 0 only.


JMMackenzie (approved):

elshize (approved):
So is this the only problem? All the other algorithms work fine now?

PR #177, opened 2019-04-18
Remove boost::te

elshize:
Since there were some problems with compiling `boost::te` on clang 8 and up (even their tests in their repo fail!), I took a look at it, and figured that we don't really need any smart stuff there. We already have different cases depending on a format, so we can simply move data to another object that is later operated on. This is what I'm doing.

This now compiles on clang 8.

But what should we do with travis? We should probably add 7 and 8 but that's another ~40 minutes of testing time. I think we should, at least for the time being, just skip 5 and 6. After all, we only test gcc 7 and 8, so why is clang any different? Plus we're testing whatever's on mac. I think that's enough.

### DIFF
--- a/src/parse_collection.cpp
+++ b/src/parse_collection.cpp
@@ -1,15 +1,14 @@
 #include <string>
 
 #include <CLI/CLI.hpp>
-#include <Porter2/Porter2.hpp>
 #include <KrovetzStemmer/KrovetzStemmer.hpp>
-#include <boost/te.hpp>
+#include <Porter2/Porter2.hpp>
+#include <boost/algorithm/string.hpp>
 #include <fmt/format.h>
 #include <spdlog/spdlog.h>
 #include <tbb/task_scheduler_init.h>
 #include <trecpp/trecpp.hpp>
 #include <warcpp/warcpp.hpp>
-#include <boost/algorithm/string.hpp>    
 
 #include "forward_index_builder.hpp"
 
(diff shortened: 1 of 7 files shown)


src/parse_collection.cpp
```diff
@@ -86,28 +88,29 @@ std::function<std::string(std::string &&)> term_processor(std::optional<std::str
 {
     if (not type) {
         return [](std::string &&term) -> std::string {
-            boost::algorithm::to_lower(term); 
-            return term;
+            boost::algorithm::to_lower(term);
+            return std::move(term);
```
amallia, line 92: You mean `std::forward` here?

amallia:
LGTM. There is only 1 comment that needs to be checked.

src/parse_collection.cpp
elshize: I don't think forward makes sense here because this is not a forwarding reference (no template) but simply an rvalue reference. What I'm now thinking is whether I need that nice or not. We could simplify this by passing by value actually. The idea was to force to move a value. If we pass by value, we could be copying by mistake if we don't move at call site.

What do you think?

amallia: Yes, you can pass by value and then move. In this case, when you call this function, you have to make sure you move when calling the lambda to avoid an extra copy though.

If you pass by forwarding reference the risk of using a move when returning is that you could pass an lvalue reference, which then gets moved (resulting in an extra copy). That’s why I wanted to use forward here.

amallia: Actually this is fine because it is not templated.

elshize: Yeah, I was gonna say.

PR #178, opened 2019-04-25
Implement tokenizer and use for query parsing.

elshize:
Additionally, parse abbreviations (n.s.a.) and possessives (dog's).

### DIFF
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -6,6 +6,7 @@
 #include <optional>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <unordered_map>
 
 #include <KrovetzStemmer/KrovetzStemmer.hpp>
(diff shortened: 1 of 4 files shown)


elshize:
So I've added two additional rules for parsing queries/documents:
1. Abbreviations as in U.S.A.
2. Possessives as in dog's collar.

These two are perfect candidates for simple query expansion, e.g., query:
> person's rights in U.S.A.
would be expanded to
```
[person, person's, rights, in, U.S.A, USA]
```

I opened an issue: #179

include/pisa/query/queries.hpp
```diff
@@ -47,22 +48,23 @@ struct Query {
         id = std::string(query_string.begin(), colon);
     }
     auto pos = colon == query_string.end() ? query_string.begin() : std::next(colon);
-    std::istringstream iline(std::string(pos, query_string.end()));
-    std::string term;
-    while (iline >> term) {
+    auto raw_query = std::string(pos, query_string.end());
```
amallia: Why not a string_view here?

elshize:
@amallia Take another look.

amallia (approved):

Issue #151, opened 2019-04-05
Add parallel query processing 


amallia:
Add a CLI argument to specify the number of threads to use to perform query processing.
The default should be `1`, but if a number higher than one is specified qp should be parallelized on a per query base.

JMMackenzie:
Did we have a particular framework for threading in mind? I'm not entirely up-to-date with how we are approaching all of the design details.

amallia:
`TBB` since it is already a dependency. 
This should be a pretty simple task, all we need is to do a parallel-for.

PR #181, opened 2019-04-26
Parallel evaluate

JMMackenzie:
First try at parallel evaluate. I will flag some potential issues or design considerations that we may change. This is at least a working prototype for now. Targets #151.

- Line 132 uses a default value (0) for the identifier if it can't use the one from the query object.
- Had to use stringstream to format output so that the call to `std::cout` wasn't garbled due to the threads.
- Added timer into the code which wraps the entire `parallel_for`

### DIFF
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -122,15 +122,19 @@ void evaluate_queries(const std::string &index_filename,
     auto source = std::make_shared<mio::mmap_source>(documents_filename.c_str());
     auto docmap = Payload_Vector<>::from(*source);
 
+    std::vector<std::vector<std::pair<float, uint64_t>>> raw_results(queries.size());
     auto start_batch = std::chrono::steady_clock::now();
-    std::vector<std::string> trec_results(queries.size());
-
     tbb::parallel_for (size_t(0), queries.size(), [&] (size_t query_idx) {
         auto qid = queries[query_idx].id;
-        auto results = query_fun(queries[query_idx].terms);
-        std::stringstream result_list;
+        raw_results[query_idx] = query_fun(queries[query_idx].terms);
+    });
+    auto end_batch = std::chrono::steady_clock::now();
+   
+    for (size_t query_idx = 0; query_idx < raw_results.size(); ++query_idx) { 
+        auto results = raw_results[query_idx];
+        auto qid = queries[query_idx].id;
         for (auto && [ rank, result ] : enumerate(results)) {
-            result_list << fmt::format("{}\t{}\t{}\t{}\t{}\t{}\n",
+            std::cout << fmt::format("{}\t{}\t{}\t{}\t{}\t{}\n",
                                      qid.value_or(std::to_string(query_idx)),
                                      iteration,
                                      docmap[result.second],
@@ -138,11 +142,6 @@ void evaluate_queries(const std::string &index_filename,
                                      result.first,
                                      run_id);
         }
-        trec_results[query_idx] = result_list.str();
-    });
-    auto end_batch = std::chrono::steady_clock::now();
-    for (auto const &result : trec_results) {
-        std::cout << result << std::endl;
     }
     auto end_print = std::chrono::steady_clock::now();
     double batch_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end_batch - start_batch).count();
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -129,7 +129,7 @@ void evaluate_queries(const std::string &index_filename,
         std::stringstream result_list;
         for (auto && [ rank, result ] : enumerate(results)) {
             result_list << fmt::format("{}\t{}\t{}\t{}\t{}\t{}\n",
-                                     queries[query_idx].id.value_or(std::to_string(0)),
+                                     qid.value_or(std::to_string(0)),
                                      iteration,
                                      docmap[result.second],
                                      rank,
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -12,6 +12,10 @@
 
 #include "mappable/mapper.hpp"
 
+#include <thread>
+#include "tbb/parallel_for.h"
+#include "tbb/task_scheduler_init.h"
+
 #include "cursor/block_max_scored_cursor.hpp"
 #include "cursor/max_scored_cursor.hpp"
 #include "cursor/scored_cursor.hpp"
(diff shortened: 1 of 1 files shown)


amallia:
Good job here! One minor comment is that the current implementation does not have any guarantees about the order of the output in terms of query id. 
I am wondering if this is a problem and what would be the impact.

Potentially we could use a vector of strings, resize it to the number of queries and then each thread accesses the vector at different indexes (based on the query). 
Finally, the vector is printed to std output.

JMMackenzie:
Right, I forgot to mention this. `trec_eval` doesn't mind, but it might be nice to have them ordered correctly.

src/evaluate_queries.cpp
```diff
@@ -118,18 +122,25 @@ void evaluate_queries(const std::string &index_filename,
     auto source = std::make_shared<mio::mmap_source>(documents_filename.c_str());
     auto docmap = Payload_Vector<>::from(*source);
 
-    for (auto const & [ qid, query ] : enumerate(queries)) {
-        auto results = query_fun(query.terms);
+    auto start_batch = std::chrono::steady_clock::now();
+    tbb::parallel_for (size_t(0), size_t(queries.size()), [&] (size_t query_idx) {
```
elshize: `queries.size()` is already of type `size_t`, no need for an explicit constructor.

src/evaluate_queries.cpp
```diff
@@ -118,18 +122,25 @@ void evaluate_queries(const std::string &index_filename,
...
+        auto qid = queries[query_idx].id;
+        auto results = query_fun(queries[query_idx].terms);
+        std::stringstream result_list;
         for (auto && [ rank, result ] : enumerate(results)) {
-            std::cout << fmt::format("{}\t{}\t{}\t{}\t{}\t{}\n",
-                                     query.id.value_or(std::to_string(qid)),
+            result_list << fmt::format("{}\t{}\t{}\t{}\t{}\t{}\n",
+                                     qid.value_or(std::to_string(0)),
```
elshize: Why not `qid.value_or(std::to_string(query_idx))`?

elshize (dismissed):
Let me know if this makes sense:

I was thinking of having a separate thread for writing results. The idea would work as follows:

1. First, we define a `tbb::concurrent_queue` with query results.
2. We pass it before the loop to a separate `std::thread` function that waits for results until all `n` queries are received, and then stops.
3. Run the loop as we do now, but simply pass the result of `query_fun` to the queue, and the other thread will do the printing.
4. After the loop, join on the writing thread to make sure we don't exit before printing is done.

The advantages:
1. We can keep results for one query together.
2. We avoid writing from many threads, which I think might be inefficient.

What do  you guys think?

elshize:
Some thoughts about parallel result reporting:

# Version 1: Vector

1. n query processing threads:
    - each processes one query at a time
    - accumulates the results in a vector when done
    - I/O mostly idle
2. Many I/O operations that take some amount of time T

# Version 2: Async Writing

1. One thread R for writing results.
    - waits for a result set from other threads
    - once received, writes it out
2. n query processing threads:
    - each processes one query at a time
    - sends asynchronously results to R
3. The main thread waits for R to finish

Hopefully, most of the time T will be done by the time queries are processed.
The point is that those are I/O operations, which can be parallelized on top of
CPU parallelization. That's why it makes sense to do it as we go instead of
at the end, when the CPU will be mostly idle through that.
Note that parallelization of writing at the end is troublesome not only because of
merging files but mostly because it won't really speed up writing.

# Version 3: Async Writing with Ordering

Before I go on, I want to note that I see nothing wrong with the fact that queries
are out of order. This is **natural** as in that's the order they finished
so I see nothing wrong in reporting them this way. Plus, it ultimately doesn't matter
for our purposes.

However, if you really want, e.g., for debugging or anything like that,
you can provide another implementation of thread R. The difference will be that
every time it receives a result set, it will first check if all others have finished
and write if so, or deter writing until all previous queries have finished, otherwise.

Here, for performance, it would be good if we ensured that we interleave queries between threads
instead of processing a contiguous range each.
But for debug, doesn't matter.

# Ending Note

If you **realy, realy want**, you can even implement this in a generic way so that you can
switch not only between V2 and V3, but also V1. But seems like an overkill to me.

# Motivating Example

Imagine you have a very long query log of, say, thousands of queries.
This will take a significant amount of time to write.
This will be especially significant for smaller collections: very fast query processing,
but writing will take the same time as for large collections.
So we have this disparity in how much writing takes between different experiments.
This is something I was hoping to avoid with Version 2.
I would personally implement both 2 and 3, and have the user choose (with default 2).

JMMackenzie:
I agree with what you're saying, indeed. However if the main concern is that the time to dump results is large with respect to query processing time, well, we don't really need to time the result dump part do we? Assuming we timed from when we started running queries until we had populated a vector of all results (but had not output them), seems fair to me, especially since we aren't timing the query loading or parsing stuff. I realize we are indeed timing the output part currently, but we could modify it.

Of course, the asynchronous writing approach is nice and elegant, and would be favourable in my opinion. I'm in agreement for not really bothering with V3 for now. I think if someone wants to debug, they'll just run a single thread anyway, so order would be preserved. Any thoughts?

elshize:
Sure, for our own time reporting, we're flexible, but I guess I'm still not sure how the measuring will be done.

I guess let's do the vector for now, and then do some measurements, and see how much the diff is between the program time and our internal time.

JMMackenzie:
Here's a second take. Now we only time the retrieval of the results, which are added directly to an ugly vector of "results" which are then output sequentially afterwards. Timings are output to capture both aspects, though the output part is clearly not optimized. Now we see expected speedups when increasing the threads. For example, on Gov2, time taken (without printing):
```
1 Thread: 1771ms
2 Threads: 914ms
4 Threads: 464ms
8 Threads: 255ms
16 Threads: 134ms
32 Threads: 131ms
```

We see a sublinear speedup which is expected from this simple parallelization approach. Perhaps we can merge this one if you're happy with it, and revisit with a thread pool later on? If so, let's make a new issue/feature request.

src/evaluate_queries.cpp
```diff
@@ -118,18 +122,32 @@ void evaluate_queries(const std::string &index_filename,
...
                                      rank,
                                      result.first,
                                      run_id);
         }
     }
+    auto end_print = std::chrono::steady_clock::now();
+    double batch_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end_batch - start_batch).count();
+    double batch_with_print_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end_print - start_batch).count();
+    spdlog::info("Time taken to process queries: {}ms", batch_ms);
+    spdlog::info("Time taken to process queries with printing: {}ms", batch_with_print_ms);
```
amallia, line 150: I would print this only in debug mode...

amallia (approved):

src/evaluate_queries.cpp
JMMackenzie: Both printouts? Or just the printing one? I assumed we wanted an output from this - let me know what you are thinking and I can update.

amallia: I would say just the second one

amallia: Actually, let's differ this to a separate PR eventually...

PR #182, opened 2019-04-26
Add weights to query terms

amallia:
Currently we have the possibility to specify the weight of a query term.

EDIT: Actually this PR is self contained. Let's see if it can get merged.


elshize:
Yeah, sounds good. Merging.

PR #183, opened 2019-04-29
Unify & Test CLI

elshize:
Here's a proposal, based on one executable, how we could approach testing our command line programs.

Here are some comments:
- For each program, we define a structure, e.g., `InvertSettings` that has all the parameters for the program's execution.
- This struct is parsed from CLI11 and passed to where the processing is taking place.
- Parsing is tested separately from processing.
- The code for parsing is in a separate file in directory `command` --- this is because we don't want to include too much when only testing parsing to one structure. There is no other logic in such structure. So think putting it out of the way in its own directory seems natural too.
- There is a template that should be used always in the actual executable file: it parses, and passes the settings to processing logic if successful.
- This template is implemented as `PISA_MAIN` macro, which takes name (say, `Invert`), and processing function (`invert::invert_forward_index`). This enforces consistency and makes it difficult to mess something up in the untested content. This way, the executable is only an include and invoking the macro.
- There's also some flexibility in that if at this point something is difficult to test (like `queries`) it can be skipped but still the argument parsing is tested, so it's better than just leaving it with no testing.
- The common options (right now they are `threads` and `batch_size` enforce proper member naming in the setting structures.


CMakeLists.txt
```diff
@@ -128,7 +128,10 @@ if(PISA_COMPILE_HEADERS)
   include(CheckCXXSourceCompiles)
   set(CMAKE_REQUIRED_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${CMAKE_CXX_STANDARD} -c")
   get_property(PISA_INTERFACE_INCLUDE_DIRECTORIES TARGET pisa PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
-  set(CMAKE_REQUIRED_INCLUDES "${CMAKE_SOURCE_DIR}/external/tbb/include" ${PISA_INCLUDE_DIRECTORIES})
+  set(CMAKE_REQUIRED_INCLUDES
+      "${CMAKE_SOURCE_DIR}/external/CLI11/include"
```
amallia, line 132: Why do you need this exactly?

elshize: Because of the `include/command/*.hpp` files that need CLI11. If you want and know how, you can remove this and filter these files. But I figured this was not worth our time.

include/pisa/command/cli.hpp
```diff
@@ -0,0 +1,37 @@
...
+        return app.add_option("-b,--batch-size",
+                              settings.batch_size,
+                              "Number of documents to process at a time",
+                              true);
+    }
+
+} // namespace options
+} // namespace pisa::cli
+
+#define PISA_MAIN(NAME, FUNCTION)                                                        \
```
amallia, line 27: I think we can avoid the macro here.
Let's use a function which takes 2 functions as parameters: `settings_name` and `function`.
Then we are supposed to call this function inside a `main`.

include/pisa/invert.hpp
```diff
@@ -11,16 +11,18 @@
...
-#include "tbb/task_group.h"
+#include <boost/filesystem.hpp>
+#include <gsl/span>
+#include <pstl/algorithm>
+#include <pstl/execution>
+#include <range/v3/view/iota.hpp>
+#include <spdlog/spdlog.h>
+#include <tbb/concurrent_queue.h>
+#include <tbb/task_group.h>
+#include <tbb/task_scheduler_init.h>
```
amallia, line 22: We did not agreed about the use of `<>` notation for includes yet. I would like to discuss advantages and disadvantages first.

include/pisa/invert.hpp
```diff
@@ -322,30 +324,26 @@ auto invert_range(gsl::span<gsl::span<Term_Id const>> documents,
     return index;
 }
 
-[[nodiscard]] auto build_batches(std::string const &input_basename,
-                                 std::string const &output_basename,
-                                 uint32_t           term_count,
-                                 size_t             batch_size,
-                                 size_t             threads) -> uint32_t
+[[nodiscard]] auto build_batches(InvertSettings const &settings) -> uint32_t
```
amallia, line 327: I think this should not have `InvertSettings ` as a dependency. I would keep the original signature..

include/pisa/command/invert.hpp
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <thread>
+#include <variant>
+
+#include <CLI/CLI.hpp>
```
amallia, line 6: We should not have a CLI dependency here in my opinion...

include/pisa/invert.hpp
```diff
@@ -403,19 +401,16 @@ void merge_batches(std::string const &output_basename, uint32_t batch_count, uin
     }
 }
 
-void invert_forward_index(std::string const &input_basename,
-                          std::string const &output_basename,
-                          uint32_t           term_count,
-                          size_t             batch_size,
-                          size_t             threads)
+void invert_forward_index(InvertSettings settings)
```
amallia, line 404: Same as above. 

My idea is that whatever stays in `include/pisa` should not be CLI related...

amallia (changes requested):
I added a bunch of comments. Let me know what you think

include/pisa/invert.hpp
elshize: Why?

If we don't, the testing does very little in terms of ensuring that we use the right parameters. We can have correct parsing, and correct function, but call it with wrong parameters.

include/pisa/command/invert.hpp
elshize: We could get away with this by moving Settings (see my other comment for that) to the other header, and `command/invert.hpp` or in that case even just `cli.hpp` file or similar, has the parsing function that is being tested. So no CLI dependency but we include the original header for testing parsing.

include/pisa/invert.hpp
elshize: See the two above comments.

include/pisa/invert.hpp
elshize: Discuss away. My point: these are not local files, as in, they are external libraries that just happen to be on the path. They are, in fact, included using `-I/-system` argument by `gcc`/`clang`. We can discuss later this week when I'm back home, if you prefer that.

PR #187, opened 2019-05-03
Block params

JMMackenzie:
Some potential issues:
- I am not particularly familiar with the `boost::variant` stuff, so please check that I am not abusing it.
- I was able to handle most of the command line argument interactions using the CLI `excludes`, but this does not work for supplying `--lambda` without `--variable-block` so I wrote an explicit case to catch that one.

Also updated the documentation. Please suggest changes and I will try to fix ASAP.

### DIFF
--- a/include/pisa/wand_data.hpp
+++ b/include/pisa/wand_data.hpp
@@ -4,87 +4,77 @@
 #include "spdlog/spdlog.h"
 
 #include "binary_freq_collection.hpp"
-#include "scorer/bm25.hpp"
 #include "mappable/mappable_vector.hpp"
+#include "scorer/bm25.hpp"
+#include "util/progress.hpp"
 #include "util/util.hpp"
 #include "wand_data_raw.hpp"
-#include "util/progress.hpp"
 
 class enumerator;
 namespace pisa {
 
-    template<typename Scorer=bm25, typename block_wand_type=wand_data_raw<bm25> >
-    class wand_data {
-    public:
+template <typename Scorer = bm25, typename block_wand_type = wand_data_raw<bm25>>
+class wand_data {
+   public:
+    using wand_data_enumerator = typename block_wand_type::enumerator;
 
-        using wand_data_enumerator = typename block_wand_type::enumerator;
+    wand_data() {}
 
-        wand_data() { }
+    template <typename LengthsIterator>
+    wand_data(LengthsIterator len_it,
+              uint64_t num_docs,
+              binary_freq_collection const &coll,
+              BlockSize block_size)
+    {
+        std::vector<float> norm_lens(num_docs);
+        std::vector<float> max_term_weight;
+        global_parameters params;
+        double lens_sum = 0;
+        spdlog::info("Reading sizes...");
 
-        template<typename LengthsIterator>
-        wand_data(LengthsIterator len_it, uint64_t num_docs,
-                      binary_freq_collection const &coll, partition_type type = partition_type::fixed_blocks, boost::variant<float, uint64_t> block_size = uint64_t(0)) {
-            std::vector<float> norm_lens(num_docs);
-            std::vector<float> max_term_weight;
-            global_parameters params;
-            double lens_sum = 0;
-            spdlog::info("Reading sizes...");
+        for (size_t i = 0; i < num_docs; ++i) {
+            float len = *len_it++;
+            norm_lens[i] = len;
+            lens_sum += len;
+        }
 
-            for (size_t i = 0; i < num_docs; ++i) {
-                float len = *len_it++;
-                norm_lens[i] = len;
-                lens_sum += len;
+        float avg_len = float(lens_sum / double(num_docs));
+        for (auto &norm_len : norm_lens) {
+            norm_len /= avg_len;
+        }
+
+        typename block_wand_type::builder builder(coll, params);
+        {
+            pisa::progress progress("Processing posting lists", coll.size());
+            for (auto const &seq : coll) {
+                auto v = builder.add_sequence(seq, coll, norm_lens, block_size);
+                max_term_weight.push_back(v);
+                progress.update(1);
             }
-
-            float avg_len = float(lens_sum / double(num_docs));
-            for(auto& norm_len: norm_lens) {
-                norm_len /= avg_len;
-            }
-
-            typename block_wand_type::builder builder(type, coll, params);
-            {
-                pisa::progress progress("Processing posting lists", coll.size());
-                for (auto const &seq: coll) {
-                    auto v = builder.add_sequence(seq, coll, norm_lens, block_size);
-                    max_term_weight.push_back(v);
-                    progress.update(1);
-                }
-            }
-            builder.build(m_block_wand);
-            m_norm_lens.steal(norm_lens);
-            m_max_term_weight.steal(max_term_weight);
-
         }
+        builder.build(m_block_wand);
+        m_norm_lens.steal(norm_lens);
+        m_max_term_weight.steal(max_term_weight);
+    }
 
-        float norm_len(uint64_t doc_id) const {
-            return m_norm_lens[doc_id];
-        }
+    float norm_len(uint64_t doc_id) const { return m_norm_lens[doc_id]; }
 
-        float max_term_weight(uint64_t list) const {
-            return m_max_term_weight[list];
-        }
+    float max_term_weight(uint64_t list) const { return m_max_term_weight[list]; }
 
+    wand_data_enumerator getenum(size_t i) const { return m_block_wand.get_enum(i); }
 
-        wand_data_enumerator getenum(size_t i) const {
-            return m_block_wand.get_enum(i);
-        }
+    const block_wand_type &get_block_wand() const { return m_block_wand; }
 
-        const block_wand_type& get_block_wand() const {
-            return m_block_wand;
-        }
+    template <typename Visitor>
+    void map(Visitor &visit)
+    {
+        visit(m_block_wand, "m_block_wand")(m_norm_lens, "m_norm_lens")(m_max_term_weight,
+                                                                        "m_max_term_weight");
+    }
 
-        template<typename Visitor>
-        void map(Visitor &visit) {
-            visit
-                    (m_block_wand, "m_block_wand")
-                    (m_norm_lens, "m_norm_lens")
-                    (m_max_term_weight, "m_max_term_weight");
-        }
-
-
-    private:
-        block_wand_type m_block_wand;
-        mapper::mappable_vector<float> m_norm_lens;
-        mapper::mappable_vector<float> m_max_term_weight;
-    };
-}
+   private:
+    block_wand_type m_block_wand;
+    mapper::mappable_vector<float> m_norm_lens;
+    mapper::mappable_vector<float> m_max_term_weight;
+};
+} // namespace pisa
--- a/include/pisa/wand_data.hpp
+++ b/include/pisa/wand_data.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "boost/variant.hpp"
 #include "spdlog/spdlog.h"
 
 #include "binary_freq_collection.hpp"
@@ -22,7 +23,7 @@ namespace pisa {
 
         template<typename LengthsIterator>
         wand_data(LengthsIterator len_it, uint64_t num_docs,
-                      binary_freq_collection const &coll, partition_type type = partition_type::fixed_blocks, const float lambda = 0.0f) {
+                      binary_freq_collection const &coll, partition_type type = partition_type::fixed_blocks, boost::variant<float, uint64_t> block_size = uint64_t(0)) {
             std::vector<float> norm_lens(num_docs);
             std::vector<float> max_term_weight;
             global_parameters params;
@@ -44,7 +45,7 @@ namespace pisa {
             {
                 pisa::progress progress("Processing posting lists", coll.size());
                 for (auto const &seq: coll) {
-                    auto v = builder.add_sequence(seq, coll, norm_lens, lambda);
+                    auto v = builder.add_sequence(seq, coll, norm_lens, block_size);
                     max_term_weight.push_back(v);
                     progress.update(1);
                 }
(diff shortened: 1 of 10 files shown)


JMMackenzie:
Not currently building due to the case where `--variable-block` is passed without `--lambda` - I was passing the variant in as a default `uint64_t` so when it gets to the variable block partition and tries to `get()` the float, it fails (since the `uint64_t` is the set type). 

Not sure what the clean way to fix this is, do you have suggestions?

JMMackenzie:
OK I have fixed the `test` issue, but I think this is quite ugly now (and I had to change the test cases to explicitly use the correct `variant` to avoid issues. Let me know what you think, we might want to refactor this.

include/pisa/wand_data.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
```
elshize, line 1: Let's `clang-format` these modified files, they're quite a mess.

include/pisa/wand_data.hpp
```diff
@@ -22,7 +23,7 @@ namespace pisa {
 
         template<typename LengthsIterator>
         wand_data(LengthsIterator len_it, uint64_t num_docs,
-                      binary_freq_collection const &coll, partition_type type = partition_type::fixed_blocks, const float lambda = 0.0f) {
+                      binary_freq_collection const &coll, partition_type type = partition_type::fixed_blocks, boost::variant<float, uint64_t> block_size = uint64_t(0)) {
```
elshize: In general, I'd like to use `std::variant` instead. But seeing as there are some problems with its support by AppleClang (lol, I know), we might wanna use boost for now. @amallia Are we ready to use `std::variant` properly or not yet?

elshize: Correct my if I'm wrong, but I think now that we have a variant type, `partition_type` is redundant, right?

include/pisa/wand_data_compressed.hpp
```diff
@@ -111,12 +112,12 @@ namespace {
...
             }
 
-            float add_sequence(binary_freq_collection::sequence const &seq, binary_freq_collection const &coll, std::vector<float> const & norm_lens, const float lambda = 0.0f){
+            float add_sequence(binary_freq_collection::sequence const &seq, binary_freq_collection const &coll, std::vector<float> const & norm_lens, boost::variant<float, uint64_t> block_size) {
 
                 if (seq.docs.size() > configuration::get().threshold_wand_list) {
 
-                    auto t = ((type == partition_type::fixed_blocks) ? static_block_partition(seq, norm_lens)
-                                                      : variable_block_partition(coll, seq, norm_lens, lambda));
+                    auto t = ((type == partition_type::fixed_blocks) ? static_block_partition(seq, norm_lens, boost::get<uint64_t>(block_size))
```
elshize: So typically, you would use a static visitor: https://www.boost.org/doc/libs/1_70_0/doc/html/variant/reference.html#variant.concepts.static-visitor

However, it is way too verbose for this, at least for inline use. Might as well do an `if` statement. If `type` not available, you can use `which` method that returns the index of the type that's in the variant. Not a great design though.

Not that for `std::variant`, it is very easy to write a wrapper function that turns this use case into:
```cpp
auto t = match(block_size,
    [](float alpha) { ... },
    [](uint64_t size) { ... })
```

You can see the example of that in on of the old commits in `warcpp`: https://github.com/pisa-engine/warcpp/blob/e8c53c09cb983ff6166360d46c514d8e36acb91a/include/warcpp/warcpp.hpp#L94 This **doesn't work** for our currently used AppleClang, though (blame Antonio).

It might be also possible for `boost::variant` but not sure how easy it is (the interface is slightly different). There is a lib for that: https://github.com/exclipy/inline_variant_visitor but it seems much longer compared to the `std::variant` version. And clearly, bringing another library for this is an overkill.

include/pisa/wand_utils.hpp
```diff
@@ -11,10 +11,13 @@ enum class partition_type { fixed_blocks, variable_blocks };
 template <typename Scorer = bm25>
 std::pair<std::vector<uint32_t>, std::vector<float>>
 static_block_partition(binary_freq_collection::sequence const &seq,
-                       std::vector<float> const &norm_lens) {
+                       std::vector<float> const &norm_lens,
+                       const uint64_t in_block_size = 0) {
   std::vector<uint32_t> block_docid;
   std::vector<float> block_max_term_weight;
   uint64_t block_size = configuration::get().block_size;
+  if (in_block_size > 0)
```
elshize: Missing braces.

include/pisa/wand_utils.hpp
```diff
@@ -11,10 +11,13 @@ enum class partition_type { fixed_blocks, variable_blocks };
 template <typename Scorer = bm25>
 std::pair<std::vector<uint32_t>, std::vector<float>>
 static_block_partition(binary_freq_collection::sequence const &seq,
-                       std::vector<float> const &norm_lens) {
+                       std::vector<float> const &norm_lens,
+                       const uint64_t in_block_size = 0) {
```
elshize: I think it would be more clear if we used `std::optional` here, explicitly indicating when a block size is defined and when it's not, instead of using a magic value 0:

```cpp
std::optional<uint64_t> const = std::nullopt) {
```

src/create_wand_data.cpp
```diff
@@ -9,30 +10,32 @@
...
+    bool compress = false;
+    bool range = false;
 
     CLI::App app{"create_wand_data - a tool for creating additional data for query processing."};
     app.add_option("-c,--collection", input_basename, "Collection basename")->required();
     app.add_option("-o,--output", output_filename, "Output filename")->required();
-    app.add_option("-l,--lambda", lambda, "Lambda parameter for variable blocks");
     auto var_block_opt = app.add_flag("--variable-block", variable_block, "Variable length blocks");
+    auto var_block_param_opt = app.add_option("-b,--block-size", fixed_block_size, "Block size for fixed-length blocks")->excludes(var_block_opt);
+    app.add_option("-l,--lambda", lambda, "Lambda parameter for variable blocks")->excludes(var_block_param_opt);
```
elshize: Add `->needs(var_block_opt)`

src/create_wand_data.cpp
```diff
@@ -42,20 +45,36 @@ int main(int argc, const char **argv) {
         (p_type == partition_type::fixed_blocks) ? "static partition" : "variable partition";
     spdlog::info("Block based wand creation with {}", partition_type_name);
 
-    binary_collection      sizes_coll((input_basename + ".sizes").c_str());
+    binary_collection sizes_coll((input_basename + ".sizes").c_str());
     binary_freq_collection coll(input_basename.c_str());
 
+    // Initialize the variant to the correct type
+    boost::variant<float, uint64_t> block_size = uint64_t(0);
```
elshize: We probably want to make it `const`, so I'd either use another function to set this up, or just wrap it in a lambda:

```cpp
auto const block_size = [&] () -> boost::variant<float, uint64_t> {
    if (variable_block) {
        return lambda;
    } else {
        return fixed_block_size;
    }
}();
```
Not that I moved checking compliance of defined options above to the CLI part.

elshize (changes requested):
If you ask me, we should probably go with `std::variant`. If we're not ready to use `match`, at least we can use `std::get_if` to easily check for the type right now.

I would also create a `BlockSize` type defined as follows:
```cpp
struct FixedSize {
    uint64_t size;
};
struct VariableSize {
    float alpha;
};
using BlockSize = std::variant<FixedSize, VariableSize>;
```

Then, we could use it one of these ways:

1. If we upgrade AppleClang
```cpp
auto t = match(block_size,
    [](FixedSize size){ return f(size); },
    [](VariableSize size){ return g(size); });
```
2. If we don't
```cpp
auto t = std::get_if<FixedSize>(block_size) != nullptr
    ? f(*std::get_if<FixedSize>(block_size))
    : g(*std::get_if<VariableSize>(block_size))
```

include/pisa/wand_data.hpp
amallia: Not yet, I asked @JMMackenzie to use `boost` because of the incompatibility.
We can migrate all the `boost::variant` occurrences all together at some point.

amallia: This is correct! and it is such a great improvement :)

include/pisa/wand_data_compressed.hpp
amallia: I agree with this one!

amallia:
Thanks @JMMackenzie for the PR! I agree with all the comments @elshize already made.
Once those are addressed we can merge!

elshize:
So if we stick to `boost::variant`, I think something like this would work:
```cpp
auto t = block_size.type() == typeid(FixedSize)
    ? f(boost::get<FixedSize>(block_size))
    : g(boost::get<VariableSize>(block_size))
```
Needs to be double-checked, not sure if the standard `typeid` works here or must use some boost version, but I imagine this should work.

Also, I suggest to use the proposed "inner" types (see https://github.com/pisa-engine/pisa/pull/187#pullrequestreview-233460491) in function signatures, e.g.:

```cpp
static_block_partition(binary_freq_collection::sequence const &seq,
                       std::vector<float> const &norm_lens) {
                       std::vector<float> const &norm_lens,
                       std::optional<FixedSize> const in_block_size = std::nullopt) {
```

include/pisa/wand_data.hpp
JMMackenzie: Refactored in the latest.

include/pisa/wand_data_compressed.hpp
JMMackenzie: Yes, good point. I have used a similar idea that you mentioned below since we are using `boost` for now. Let's revisit this in #188 when we get to it.

include/pisa/wand_utils.hpp
JMMackenzie: I have decided it's better to not allow default parameters here. Instead, I have built the default constructor for either the `VariableBlock` or `FixedBlock` type to directly read the corresponding value from the config file. So, instantiating a default object will give us the default configuration, which can be used directly. Please check it out and let me know what you think - I believe this approach is much more robust and readable.

JMMackenzie:
I have overhauled this now:

- Refactored the code to remove the `partition_type`
- Provide structs for each type that ingest default params out of config file
- Clean up using `clang_format` - I am not sure it did what it was expected to do, could you double check this?
- Added the CLI `needs` to `lambda` which improves simplicity of the `.cpp` file.

I think that's it for now. The code is greatly simplified, and seems to be working on my end.

include/pisa/wand_data_compressed.hpp
```diff
@@ -9,230 +10,205 @@
...
-                  type(type),
-                  params(params),
-                  compressor_builder(coll.num_docs(), params) {
-                spdlog::info("Storing max weight for each list and for each block...");
-            }
+            if (seq.docs.size() > configuration::get().threshold_wand_list) {
 
-            float add_sequence(binary_freq_collection::sequence const &seq, binary_freq_collection const &coll, std::vector<float> const & norm_lens, const float lambda = 0.0f){
+                auto t =
+                    (block_size.type() == typeid(FixedBlock))
```
elshize: Nitpick: no need for the outer parens.

src/create_wand_data.cpp
```diff
@@ -9,53 +10,68 @@
...
+#include "wand_data_raw.hpp"
 
 #include "CLI/CLI.hpp"
 
-int main(int argc, const char **argv) {
+int main(int argc, const char **argv)
+{
     using namespace pisa;
 
     float lambda = 0.0f;
```
elshize: Let's use `std::optional<float> lambda{};` here instead. It works out of the box. The difference is that now it will be `std::nullopt` when no option is passed. Then it makes more sense when you check for that later.

elshize: Let's use `std::optional<float> lambda{};` here instead. It works out of the box. The difference is that now it will be `std::nullopt` when no option is passed. Then it makes more sense when you check for that later.

src/create_wand_data.cpp
```diff
@@ -9,53 +10,68 @@
...
+    uint64_t fixed_block_size = 0;
```
elshize: Same as above.

src/create_wand_data.cpp
```diff
@@ -9,53 +10,68 @@
...
-    std::string partition_type_name =
-        (p_type == partition_type::fixed_blocks) ? "static partition" : "variable partition";
+    std::string partition_type_name = (lambda) ? "variable partition" : "static partition";
     spdlog::info("Block based wand creation with {}", partition_type_name);
 
-    binary_collection      sizes_coll((input_basename + ".sizes").c_str());
+    binary_collection sizes_coll((input_basename + ".sizes").c_str());
     binary_freq_collection coll(input_basename.c_str());
 
+    // Initialize the variant to the correct type, default arguments from config
```
elshize: How about we use the following code here. For one, it makes the block size `const`, and then also is clear. Rule of thumb: if you have to describe a specific block/branch in a comment, you most likely want to rearrange things to make it more clear---and then remove the comments.

```cpp
auto const block_size = [&]() -> BlockSize {
    if (variable_block) {
        return fixed_block_size ? FixedBlock(*fixed_block_size) : FixedBlock();
    } else {
        return lambda ? VariableBlock(*lambda) : VariableBlock();
    }
}();
```

elshize (changes requested):
@JMMackenzie make sure all tests are passing, seems like you haven't updated all the call sites.

src/create_wand_data.cpp
JMMackenzie: This is elegant, thanks! Incorporating it now.

src/create_wand_data.cpp
JMMackenzie: Got it, ta.

src/create_wand_data.cpp
JMMackenzie: Yes, great point!

JMMackenzie:
Good pickups, thanks a lot. I am pushing a new commit now which should rectify all of these issues (all tests pass on my system).

amallia (approved):

elshize (approved):

PR #192, opened 2019-05-06
Rename PEF

amallia:
- Rename `opt` to `pefopt`
- Rename `uniform` to `pefuniform`


elshize (approved):
Looks good but can we run `clang-format` on these files now while we're at it?

PR #220, opened 2019-07-12
Compute norm_lens at run time

amallia:
Compute `norm_lens` at run time instead of storing it precomputed. This could potentially make the BM25 scoring slightly slower, but it makes the software much more flexible as `doc_len` can be used by other scoring functions.


elshize (changes requested):
Can you reformat it with the current `.clang-format`?

amallia:
> Can you reformat it with the current `.clang-format`?

Sorry my editor was misconfigured.

elshize (approved):

elshize:
Looks sensible; let's check what travis has to say.

JMMackenzie (approved):

PR #221, opened 2019-07-15
Added term_count and collection_len

amallia:
I am adding `term_count` and `collection_len` as they are needed by QL scorer.
`term_count` is the sum of the frequencies for each document for a given term.
`collection_len` is the sum of all the `term_count` variables for all the terms.


JMMackenzie (approved):
Looks fine to me, as long as Travis is happy :-)

Issue #143, opened 2019-04-02
Index format v1.0


amallia:
I am opening this one to start a discussion about our index format.

Ideally we would like to have a property file, say `.prop` where we store meta information about the index. 
For now, I can think about storing in the property file an index version for compatibility purposes and the number of documents of the collection.
Right now the latter is stored inside the `.docs` file, which is not very elegant.
Also, if we do that we might remove the following as it is a duplicate.
https://github.com/pisa-engine/pisa/blob/46917a1d93e340c8cdd9e6b2b60b552d99df1779/include/pisa/forward_index_builder.hpp#L184 

Is there anything else we should store in there?

JMMackenzie:
This seems reasonable. 

Just as an example, Indri uses an XML format for storing metadata, and Galago uses JSON I think. 

For example, here is an Indri manifest file for an old Gov2 Index:
```
<parameters>
	<code-build-date>Apr 21 2014</code-build-date>
	<corpus>
		<document-base>1</document-base>
		<frequent-terms>177031</frequent-terms>
		<maximum-document>25205180</maximum-document>
		<total-documents>25205179</total-documents>
		<total-terms>23451774775</total-terms>
		<unique-terms>39177923</unique-terms>
	</corpus>
	<fields></fields>
	<indri-distribution>Indri development release 5.6</indri-distribution>
	<type>DiskIndex</type>
</parameters>
```

I propose that we could store the following information (perhaps):

- Version data
- Number of documents
- Number of postings lists (terms)
- Compression codec (? - would allow us to reduce command line args)

Would it also be worth storing a similar file for the wand structures?

- Variable vs Static blocks
- Range-oriented blocks (or not)
- Perhaps the mean block size
- Block compression

What do you think? This is just my thoughts, but it would definitely be nice to have this information.

amallia:
Yes, it is going to look like the `indri` manifest.  
I would use either `yaml` or `json`. 

`indri-distribution ` looks like what is going to be our version, so i would call it just `version`. We need to decide if our data version is going to corresponds to the software version. I would say so, it will make everything easier. We can rely on the major version to decide if an index is incompatible.

`total-documents` is what we need to (number of documents), but I dont see a need of  `document-base ` and `maximum-document ` (at least for now).

`unique-terms ` can be computed in O(1), but we can definitely add it for convenience. what do you think?

`total-terms ` we also need this one for language model scorers

[...]

PR #222, opened 2019-07-15
Added QL scorer (and many more...)

amallia:
- [x] Clang-format
- [x] Test efficiency
- [x] Manual verify effectiveness in all configuration

**Update**: Added PL2 scoring

**Update2**: Tested efficiency with BM25. It seems a bit slower. I suspect it is because we recompute the `q_weight` and I think it is correct to do it. Using WAND we go from `9.45` to `11.3` ms. Comments?


include/pisa/codec/compact_elias_fano.hpp
```diff
@@ -107,7 +108,7 @@ namespace pisa {
                 uint64_t v = *it++;
 
                 if (i && v < last) {
-                    throw std::runtime_error("Sequence is not sorted");
+                    throw std::runtime_error(fmt::format("Sequence is not sorted. {} < {}",v,last));
```
elshize: 1. clang-format
2. What do you think about writing something like `"Compact Elias-Fano: sequence is not sorted: value at index {} ({}) lower than its predecessor ({})", i, v, last` --- just so someone seeing the error can figure out what's up.
3. Can we rename `i` to maybe `pos` and `v` to `current`? Unless you got a better one, then by all means; just something more informative than `i`.

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -1,44 +1,45 @@
 #pragma once
 
-#include <vector>
 #include "query/queries.hpp"
-#include "scorer/bm25.hpp"
+#include <vector>
 
 namespace pisa {
 
 template <typename Index, typename WandType, typename Scorer>
```
elshize: `Scorer` is not used anymore. Btw, did you check if this impacts efficiency at all? Might be worth checking if haven't.

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -1,44 +1,45 @@
...
     using wdata_enum = typename WandType::wand_data_enumerator;
 
-    enum_type  docs_enum;
+    enum_type docs_enum;
     wdata_enum w;
-    float      q_weight;
-    Scorer     scorer;
-    float      max_weight;
+    float q_weight;
+    std::function<float(uint32_t, uint32_t)> scorer;
```
elshize: Let's maybe introduce a type alias somewhere so this can be used everywhere? What do you think?

include/pisa/scorer/scorer.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <cstdint>
+
+namespace pisa {
+
+template <typename Wand>
+struct scorer {
+   protected:
+    const Wand &m_wdata;
+   public:
+    explicit scorer(const Wand &wdata) : m_wdata(wdata) {}
```
elshize: We have to be careful about our naming convention. We have a scorer that returns another scorer (which is referred to as "scorer" in other places). We need to be more precise about what is what. Off the top of my head, I see two options:

1. This structure is referred to as a scorer factory, and it produces scorers.
2. This structure is called `scorer` and it produces `term_scorers`. Or, better yet, this one is an `index_scorer` so that there's no confusion.

I personally prefer the 2nd option. Let's discuss.

include/pisa/scorer/scorer.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+    virtual std::function<float(uint32_t, uint32_t)> operator()(uint64_t term_id) const = 0;
```
elshize: Due to the same kind of confusion I spoke off above, I think might be better to explicitly name this function for what it is, either `scorer` or `term_scorer`. What do you think?

include/pisa/scorer/scorer_factory.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "scorer.hpp"
+#include "ql.hpp"
+#include "bm25.hpp"
+
+namespace pisa {
+
+auto get_scorer = [](std::string const &scorer_name, auto const &wdata) -> std::unique_ptr<scorer<decltype(wdata)>> {
```
elshize: This is not much of a factory, just a "make" function. I suggest simply moving it to the scorer file as either a free function or a static class method called `from_name` or something like this.

include/pisa/wand_data.hpp
```diff
@@ -25,72 +28,98 @@ class wand_data {
     wand_data(LengthsIterator len_it,
               uint64_t num_docs,
               binary_freq_collection const &coll,
+              std::string const &scorer_name,
               BlockSize block_size)
     {
         std::vector<uint32_t> doc_lens(num_docs);
+        uint32_t max_len = 0;
```
elshize: Let's rename to `max_doc_len` maybe? Lots of lens going on here, might be good to clarify.

include/pisa/wand_data.hpp
```diff
@@ -25,72 +28,98 @@ class wand_data {
...
+
+        for (auto const &seq : coll) {
+            size_t term_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
+            term_counts.push_back(term_count);
+            term_lens.push_back(seq.docs.size());
+        }
+        m_doc_lens.steal(doc_lens);
+        m_term_counts.steal(term_counts);
+        m_term_lens.steal(term_lens);
+        m_avg_len = avg_len;
```
elshize: I feel like introducing these local variables only to later assign them here makes much sense. Just adds to the line count. Might as well just init, say, `m_collection_len = 0` and add to it. As far as `m_num_docs` goes, we should probably use constructor initializer (if so, make sure to move it as the first var in struct).

That said, imo this seems like a lot of logic for a constructor. I would honestly just create a static class function to construct it and execute constructor as the last line. But then again I'm not sure if you can move the mappable vector (can you?).

include/pisa/wand_data.hpp
```diff
@@ -25,72 +28,98 @@ class wand_data {
...
             collection_len += len;
         }
 
         float avg_len = float(collection_len / double(num_docs));
 
         typename block_wand_type::builder builder(coll, params);
+
+        for (auto const &seq : coll) {
+            size_t term_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
+            term_counts.push_back(term_count);
```
elshize: The naming is confusing, you're not counting terms but occurrences, right? Maybe `term_occurrence_counts` would be a better name? Same with `term_lens`---these are just term frequencies, right?

include/pisa/wand_data.hpp
```diff
@@ -25,72 +28,98 @@ class wand_data {
...
+        auto scorer_ptr = get_scorer(scorer_name, *this);
         {
             pisa::progress progress("Processing posting lists", coll.size());
+            size_t term_id = 0;
             for (auto const &seq : coll) {
-                size_t term_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
-                terms_count.push_back(term_count);
-                auto v = builder.add_sequence(seq, coll, doc_lens, avg_len, block_size);
+                auto v = builder.add_sequence(
+                    seq, coll, doc_lens, avg_len, (*scorer_ptr)(term_id), block_size);
```
elshize: See, this is awkward, it would be much more clear if it was `scorer_ptr->term_scorer(term_id)`. Also, not sure we need that `_ptr`, this is what the type system is for. It's not like you have both pointer and object and need to differentiate.

test/test_wand_data.cpp
```diff
@@ -69,14 +76,13 @@ TEST_CASE("wand_data_range")
             if (seq.docs.size() < 1024) {
                 auto max = wdata_range.max_term_weight(term_id);
                 auto &w = wdata_range.get_block_wand();
-                Score_Function<Scorer, WandType> score_func{1.f, wdata_range};
+                auto s = scorer->operator()(term_id);
```
elshize: Awkward again.

include/pisa/cursor/max_scored_cursor.hpp
```diff
@@ -1,40 +1,41 @@
 #pragma once
 
-#include <vector>
 #include "query/queries.hpp"
-#include "scorer/bm25.hpp"
+#include <vector>
 
 namespace pisa {
 
 template <typename Index, typename Scorer>
```
elshize: `Scorer` not used.

include/pisa/query/algorithm/and_query.hpp
```diff
@@ -28,41 +24,30 @@ struct and_query {
             ordered_cursors.push_back(&en);
         }
 
-
         // sort by increasing frequency
         std::sort(ordered_cursors.begin(), ordered_cursors.end(), [](Cursor *lhs, Cursor *rhs) {
-            return lhs->docs_enum.size() < rhs->docs_enum.size();
+            return lhs->size() < rhs->size();
```
elshize, line 29: How come this works, cursors don't seem to have `size`, `docid` and so on, do they? What am I missing?

include/pisa/score_opt_partition.hpp
```diff
@@ -37,9 +36,8 @@ namespace pisa {
             float sum;
             uint64_t end_sequence;
             uint64_t element_count;
-            float estimated_idf;
 
-            score_window(ForwardIterator begin, posting_t base, wand_cost_t cost_upper_bound, float fixed_cost, size_t size, float e_idf)
+            score_window(ForwardIterator begin, posting_t base, wand_cost_t cost_upper_bound, float fixed_cost, size_t size)
```
elshize, line 40: Honestly, no idea what's happening here, and what you're doing, I'll just assume that removing this `estimated_idf` is intentional and makes sense.

include/pisa/wand_data_compressed.hpp
```diff
@@ -54,6 +54,8 @@ class uniform_score_compressor {
                 uint64_t elem = *(docs_begin + i);
                 elem = elem << score_bits_size;
                 elem += *(score_begin + i);
+
+                assert(i == 0 || temp.back() < elem);
```
elshize: Since this is not performance critical, can we fail regardless of debug mode?

include/pisa/wand_data_range.hpp
```diff
@@ -124,7 +125,7 @@ class wand_data_range {
         mapper::mappable_vector<float> const &m_block_max_term_weight;
     };
 
-    enumerator get_enum(uint32_t i) const
+    enumerator get_enum(uint32_t i, float) const
```
elshize, line 128: What is this `float`? Where is it used?

include/pisa/wand_data_raw.hpp
```diff
@@ -29,21 +29,23 @@ class wand_data_raw {
...
                 auto t =
                     block_size.type() == typeid(FixedBlock)
                         ? static_block_partition(
-                              seq, doc_lens, avg_len, boost::get<FixedBlock>(block_size).size)
+                              seq, doc_lens, avg_len, scorer, boost::get<FixedBlock>(block_size).size)
                         : variable_block_partition(coll,
                                                    seq,
                                                    doc_lens,
-                                                   avg_len,
+                                                   avg_len, scorer,
```
elshize: clang-format

include/pisa/wand_utils.hpp
```diff
@@ -22,11 +22,11 @@ struct VariableBlock {
 
 using BlockSize = boost::variant<FixedBlock, VariableBlock>;
 
-template <typename Scorer = bm25>
+template <typename Scorer>
 std::pair<std::vector<uint32_t>, std::vector<float>> static_block_partition(
     binary_freq_collection::sequence const &seq,
     std::vector<uint32_t> const &doc_lens,
-    float avg_len,
+    float avg_len, Scorer scorer,
```
elshize: clang-format

src/compute_intersection.cpp
```diff
@@ -85,11 +64,11 @@ int main(int argc, const char **argv)
     else if (type == BOOST_PP_STRINGIZE(T))                         \
     {                                                               \
         if (compressed) {                                           \
-            intersect<BOOST_PP_CAT(T, _index), wand_uniform_index>( \
-                index_filename, wand_data_filename, queries, type); \
+            intersect<BOOST_PP_CAT(T, _index)>( \
```
elshize: clang-format

src/evaluate_queries.cpp
```diff
@@ -49,6 +52,8 @@ void evaluate_queries(const std::string &index_filename,
 
     WandType wdata;
 
+    auto scorer_ptr = get_scorer(scorer_name, wdata);
```
elshize: Again, I think just `scorer` is perfectly fine.

src/queries.cpp
```diff
@@ -147,6 +149,8 @@ void perftest(const std::string &index_filename,
         }
     }
 
+    auto scorer_ptr = get_scorer(scorer_name, wdata);
```
elshize: Same

elshize (changes requested):

include/pisa/scorer/scorer.hpp
amallia: Agreed with `term_scorer`

include/pisa/scorer/scorer_factory.hpp
amallia: I agree with you. Not sure it can be moved to the scorer file as it depends on `bm25.hpp` and `ql.hpp`. At the same time these classes depend on the `scorer.hpp` one since they derive from it.

include/pisa/wand_data.hpp
amallia: This can actually be removed. We do not need it anymore.

include/pisa/query/algorithm/and_query.hpp
amallia:
```
using cursor          = typename Index::document_enumerator;
```

include/pisa/score_opt_partition.hpp
amallia: Yes, it is done on purpose.

include/pisa/wand_data.hpp
amallia: I addressed the first point. Let's address the second one in a separate PR.

include/pisa/wand_data_range.hpp
amallia: This is an hack we have to deal with the current design. Unfortunately we need the max_score of the term to decompress the score, so it is useful only for wand in the compressed version. I couldnt come with a better design with the current implementation. I suggest to deal with it in a separate PR which will address a more generic and better design for wand_data.

include/pisa/wand_data.hpp
amallia: Agreed on `term_occurrence_counts`.
`term_lens` are the lengths of the posting list.  It is the number of postings per term. Do you prefer `term_posting_counts`?

include/pisa/wand_data.hpp
```diff
@@ -29,48 +29,41 @@ class wand_data {
...
+        m_avg_len = float(m_collection_len / double(num_docs));
 
         typename block_wand_type::builder builder(coll, params);
 
         for (auto const &seq : coll) {
-            size_t term_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
-            term_counts.push_back(term_count);
+            size_t term_occurrence_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
+            term_occurrence_counts.push_back(term_occurrence_count);
             term_lens.push_back(seq.docs.size());
```
elshize: How about `term_frequencies`?

include/pisa/wand_data.hpp
```diff
@@ -84,42 +77,42 @@ class wand_data {
 
     size_t doc_len(uint64_t doc_id) const { return m_doc_lens[doc_id]; }
 
-    size_t term_count(uint64_t term_id) const { return m_term_counts[term_id]; }
+    size_t term_occurrence_count(uint64_t term_id) const { return m_term_occurrence_counts[term_id]; }
 
     size_t term_len(uint64_t term_id) const { return m_term_lens[term_id]; }
```
elshize: `term_frequency`?

elshize (changes requested):

elshize (approved):
Looks good, let's see what Travis has to say, and what the manual tests show.

amallia:
I have done manual evaluation for effectiveness and all is good.

amallia:
Update the first comment. Pls check

test/test_bmw_queries.cpp
```diff
@@ -54,57 +58,63 @@ struct IndexData {
...
+                    == Approx(op_q.topk()[i].first).epsilon(0.01)); // tolerance is % relative
         }
         op_q.clear_topk();
     }
 }
 
 TEST_CASE("block_max_wand", "[bmw][query][ranked][integration]", )
 {
-    auto data = IndexData<single_index>::get();
+    for (auto &&s_name : {"bm25", "qld"}) {
```
JMMackenzie, line 96: Do we need to included the other rankers here?

include/pisa/scorer/bm25.hpp
```diff
@@ -1,30 +1,43 @@
...
+    }
 
-        //IDF (inverse document frequency)
-        static float query_term_weight(uint64_t freq, uint64_t df, uint64_t num_docs)
-        {
-            float f = (float)freq;
-            float fdf = (float)df;
-            float idf = std::log((float(num_docs) - fdf + 0.5f) / (fdf + 0.5f));
-            static const float epsilon_score = 1.0E-6;
-            return f * std::max(epsilon_score, idf) * (1.0f + k1);
```
JMMackenzie, line 26: Where did this `f *` go? Am I missing something or has the scorer changed?

JMMackenzie: Nevermind, I see what you have done. This is the query term weight, and you do this multiplication when you create the cursor.

include/pisa/wand_data_range.hpp
```diff
@@ -5,14 +5,13 @@
...
 #include "mappable/mapper.hpp"
-#include "scorer/bm25.hpp"
 #include "util/compiler_attribute.hpp"
 #include "util/util.hpp"
 #include "wand_utils.hpp"
 
 namespace pisa {
 
-template <size_t range_size = 128, size_t min_list_lenght = 1024, typename Scorer = bm25>
+template <size_t range_size = 128, size_t min_list_lenght = 1024>
```
JMMackenzie, line 14: `min_list_length` but this was already a typo before this PR.

JMMackenzie (approved):
Awesome PR, looks great.

A few general thoughts as follows.

Effectiveness
- Can we run effectiveness tests across different collections?

Efficiency
- I assume we take a minor performance hit for computing the normalized lengths for each scored document, as well as in a few other places (I believe you already mentioned this) - Would be good to quantify on a large collection with a large value of k perhaps.
- Are we sure that the dynamic pruning methods are skipping as expected still (and for the new algos?)

A final thought is that we should really get onto issue #143 because each time we merge a PR like this one we are breaking compatability with our previous index format (right?) - so it would be good to ensure we have a header or a file attached to each index so we can check that it is compliant. BTW it seems to make sense for such information to be attached to the `wand_data` file because they are ranker specific, right?
This would also allow us to remove the `scorer` arguments from the query processing binaries, I think, or at least allow the default (which is the one that the `wand_data` contains).

amallia:
> I assume we take a minor performance hit for computing the normalized lengths for each scored document, as well as in a few other places (I believe you already mentioned this) - Would be good to quantify on a large collection with a large value of k perhaps.

Apparently computing `norm_len` at runtime does not have a measurable performance difference. I have done some extensive manual testing on this, so we should be good. BTW, this part of code is already in master.

amallia:
> Can we run effectiveness tests across different collections?

Yes, I performed it on multiple collections: `robust04`, `core17`, `core18`.

test/test_bmw_queries.cpp
amallia: We could but that would make tests slower. I think that a couple of them are fine for now.

amallia:
Checked out `bf4fd98` and did some comparison with master using `gov2` and bm25 scorer (top-10).
I have used TREC 05 efficiency query log (1000 queries at random): https://trec.nist.gov/data/terabyte/05/05.efficiency_topics.gz

|Algo| master | bf4fd98|
|-----|-------|----------|
|MaxScore | 4.875 | 5.631|
|WAND | 5.461 | 6.266|
|BMW | 2.871 | 3.067|

@elshize what do you think of these numbers?

JMMackenzie:
Following Antonio, I performed the same experiment with the 150 TREC queries over gov2 with 128-element fixed-sized blocks for bm25. Results below.

_k_ = 10

|Algo| master | bf4fd98|
|-----|-------|----------|
|MaxScore | 4.744 | 5.514 |
|WAND | 3.209 | 3.629 |
|BMW | 2.669 | 2.888 |

Also ran for _k_ = 1000

|Algo| master | bf4fd98|
|-----|-------|----------|
|MaxScore | 12.012 | 14.528 |
|WAND | 11.220 | 13.276 |
|BMW | 9.035 | 9.512 |

amallia:
Replaced by https://github.com/pisa-engine/pisa/pull/285

PR #223, opened 2019-07-15
Cleanup

amallia:
Doing some cleanup to simplify `wand_data`


elshize:
Seems like `bm25` wasn't even used there, is that why you're cleaning it up or is there something else I'm missing?

amallia:
> Seems like `bm25` wasn't even used there, is that why you're cleaning it up or is there something else I'm missing?

yep. this is the reason!

elshize (approved):

PR #224, opened 2019-07-15
Max len

amallia:
Added document `max_len` for QL scorer


amallia:
Closing this one as the problem of negative score has been solved differently. We use an approach similar to what done by lucene

PR #233, opened 2019-07-28
Fix taat

### LINKED ISSUE #201: TaaT algorithms don't work with new parallel `evaluate_queries`
**Describe the bug**
When you use any of the TaaT algorithms with `evaluate_queries`, it throws a `bad_alloc` - I think this has something to do with the accumulators being init once and used by all (perhaps). Will happen even if we set `--threads 1`.

**To Reproduce**
1. Run `evaluate_queries` with any of the TaaT algos.

**Error message**
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```

**Expected behavior**
Program should run and dump trec results as normal.

**Environment info**
Operating System: Ubuntu
Compiler name: GCC (g++)
Compiler version: 7.3.0

amallia:
Fixes #201 .

Accumulator used to be a dangling reference.

### DIFF
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -109,7 +109,7 @@ void evaluate_queries(const std::string &index_filename,
     } else if (query_type == "ranked_or_taat" && wand_data_filename) {
         Simple_Accumulator   accumulator(index.num_docs());
         ranked_or_taat_query ranked_or_taat_q(k);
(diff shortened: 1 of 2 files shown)


JMMackenzie (approved):
Nice one, pretty obvious why TaaT was broken for `evaluate_queries` now :-)

PR #234, opened 2019-07-29
Small chances to taat

amallia:
@elshize You know a lot about taat and accumulators. Can you check if any of this makes sense?

### DIFF
--- a/src/evaluate_queries.cpp
+++ b/src/evaluate_queries.cpp
@@ -108,16 +108,16 @@ void evaluate_queries(const std::string &index_filename,
         };
     } else if (query_type == "ranked_or_taat" && wand_data_filename) {
         Simple_Accumulator   accumulator(index.num_docs());
-        ranked_or_taat_query ranked_or_taat_q(k);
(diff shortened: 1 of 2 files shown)


src/evaluate_queries.cpp
```diff
@@ -108,16 +108,16 @@ void evaluate_queries(const std::string &index_filename,
         };
     } else if (query_type == "ranked_or_taat" && wand_data_filename) {
         Simple_Accumulator   accumulator(index.num_docs());
-        ranked_or_taat_query ranked_or_taat_q(k);
-        query_fun = [&, ranked_or_taat_q, accumulator](Query query) mutable {
+        query_fun = [&, accumulator](Query query) mutable {
```
elshize: It looks good but I think it might make one unnecessary copy (only at the beginning, though) when you pass the accumulator to the lambda. (At least this is what I think will happen.) You could do:
```cpp
query_fun = [&, accumulator = Simple_Accumulator(index.num_docs())](Query query) mutable {
```

amallia: Fixed.

PR #235, opened 2019-07-30
Use external tokenizer

amallia:
This makes PISA even better than Lucene :P

Core18:
```
map                   	all	0.2440
P_30                  	all	0.3573
ndcg_cut_20           	all	0.3986
```


elshize:
Sounds good. Let's use it then, no reason to maintain our own.

PR #238, opened 2019-08-10
QL2

elshize:
This is work in progress, opening PR to share and see what Travis has to say.


include/pisa/codec/list.hpp
```diff
@@ -0,0 +1,7 @@
+#define PISA_INDEX_TYPES                                                                    \
```
amallia, line 1: I would rename `include/pisa/codec/list.hpp` into `include/pisa/codec/codecs.hpp`

amallia:
Replaced with QL3

PR #239, opened 2019-08-11
Update documentation

gustingonzalez:

- Fix angle brackets showing incorrectly in document_reordering.md
- Fix typo in generate_sorted_docids_mapping.py
- Update parsing.md
- Add inverting.md
- Merge index_format.md in inverting.md
- Add script to format a set of files to TREC


docs/source/parsing.md
```diff
@@ -25,30 +25,39 @@ To parse a collection into a forward index, use `parse_collection` command:
...
-- `cw09b`: forward index in binary format,
+In case you get the error `-bash: /bin/zcat: Argument list too long`, you can pass the unzipped stream using:
+
+    $ find ClueWeb09B -name '*.warc.gz' -exec zcat -q {} \;
+
+The parsing will write the following files:
+- `cw09b`: forward index in binary format.
 - `cw09b.terms`: a new-line-delimited list of sorted terms,
-  where term having ID N is on line N
+  where term having ID N is on line N+1.
```
elshize: I think it might be more clear if we stick to `N` and add something like: "(indexed from 0)" because now it suggests there's some kind of shift between ID and a line. @amallia what do you think?

docs/source/parsing.md
```diff
@@ -25,30 +25,39 @@ To parse a collection into a forward index, use `parse_collection` command:
...
-- `trecweb`: TREC web collections
-- `warc`: Web ARChive format as defined in [the format specification](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/)
-- `wapo`: TREC Washington Post Corpus
+             whitespaces, followed by the content delimited by a new line character.
+- `trectext`: TREC newswire collections.
+- `trecweb`: TREC web collections.
+- `warc`: Web ARChive format as defined in [the format specification](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/).
+- `wapo`: TREC Washington Post Corpus.
+
+In case you want to parse a file collection (for example [wiki-large](http://dg3rtljvitrle.cloudfront.net/wiki-large.tar.gz)) use the `corpusfile2trec.py` script to convert it to TREC. Then, execute the `parse_collection` command by passing the resulting file and specifying the `trectext` format.
```
elshize: What is "a file collection", is that one document per file? Let's clarify.

elshize (changes requested):
Just small questions/suggestions, and let's wait for @amallia to take a look if you don't mind.

script/corpusfile2trec.py
```diff
@@ -0,0 +1,27 @@
...
+def main():
+    parser = ArgumentParser(description="Formats a corpus file as TREC.")
+    parser.add_argument("input", type=str, help="Input directory")
+    parser.add_argument("output", type=str, help="Output filename")
+    args = parser.parse_args()
+    print("Formatting '{}' to TREC...".format(str(args.input)))
+    to_trec(compute_paths(args.input), args.output)
+
+if __name__ == "__main__":
+    main()
```
amallia: What are the assumptions here? One file corresponds to a new TREC document?
Should we call this script `files2trec` or something like that?

docs/source/parsing.md
gustingonzalez: Maybe it could be:
```suggestion
  where term on line N has the ID N-1.
```

gustingonzalez: Maybe it could be: "where the term on line N has the ID N-1".

script/corpusfile2trec.py
gustingonzalez: Yes, one file corresponds to a new TREC doc (@elshize). I think that files2trec is a shortest name (that adds clarity), but maybe it can be generic. How about docfiles2trec?

docs/source/parsing.md
gustingonzalez: Maybe it could be: "where term on line N has the ID N-1".

script/corpusfile2trec.py
amallia: I prefer `files2trec` rather than `docfiles2trec`. 

Also, I would add in the documentation that:
- one file corresponds to a new TREC document
- the filename corresponds to the document ID

docs/source/parsing.md
amallia: I agree with @elshize here. It is enough to add "(indexed from 0)". 

"where term on line N has the ID N-1" is ambiguous because you don't know if you count the lines from `0` or `1`.

amallia:
Added my comments, once fixed we can merge.

Thanks a lot for this PR. This is good stuff!

script/corpusfile2trec.py
amallia: @gustingonzalez can you rename this?

docs/source/parsing.md
gustingonzalez: I see, what do you think about "term having ID N is on line N, with N starting from 0"?

docs/source/parsing.md
gustingonzalez: @amallia, @elshize, based on your comments, I have rewritten this part as:
> In case you want to parse a set of files where each one is a document (for example, the collection [wiki-large](http://dg3rtljvitrle.cloudfront.net/wiki-large.tar.gz)), use the `files2trec.py` script to format it to TREC (take into account that each filename corresponds to a document ID). Once the file is generated, parse it with the `parse_collection` command specifying the `trectext` value for the `--format` option.

Let me know if there is ok.

docs/source/parsing.md
elshize: This sounds fine.

docs/source/parsing.md
elshize: Yep, much clearer now.

elshize (approved):
It looks good to me, @amallia you okay with merging it?

amallia:
LGTM

PR #240, opened 2019-08-15
Add usage examples

binshengliu:

- An example of indexing and retrieving.
- A script for dumping document text.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ Testing/
 docs/_build/
 
 node_modules
+
+compile_commands.json
--- a/examples/index_robust04.sh
+++ b/examples/index_robust04.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
-PISA_PREFIX=~/src/pisa
-PISA_BIN=$PISA_PREFIX/build/bin
-THREADS=20
+set -e
+
+THREADS=$(nproc)
 
 COLLECTION=(/research/remote/collections/TREC/disk4/fr.gz
 	    /research/remote/collections/TREC/disk4/ft.gz
@@ -12,19 +12,16 @@ COLLECTION=(/research/remote/collections/TREC/disk4/fr.gz
 ROB_PREFIX=./robust04
 
 # Generate robust04 robust04.terms robust04.urls robust04.documents
-zcat ${COLLECTION[@]} | $PISA_BIN/parse_collection -j $THREADS -f trectext --content-parser html --stemmer krovetz -o $ROB_PREFIX
+zcat ${COLLECTION[@]} | parse_collection -j $THREADS -f trectext --content-parser html --stemmer krovetz -o $ROB_PREFIX
 
 # Generate robust04.docs robust04.freqs robust04.sizes
-$PISA_BIN/invert -i $ROB_PREFIX -o $ROB_PREFIX -j 40 --term-count $(wc -l < $ROB_PREFIX.terms)
+invert -i $ROB_PREFIX -o $ROB_PREFIX -j $THREADS --term-count $(wc -l < $ROB_PREFIX.terms)
 
 # Generate robust04.index.opt
-$PISA_BIN/create_freq_index -t opt -c $ROB_PREFIX -o $ROB_PREFIX.index.opt --check
+create_freq_index -t pefopt -c $ROB_PREFIX -o $ROB_PREFIX.index.opt --check
 
 # Generate robust04.wand
-$PISA_BIN/create_wand_data -c $ROB_PREFIX -o $ROB_PREFIX.wand
+create_wand_data -c $ROB_PREFIX -o $ROB_PREFIX.wand
 
-# Test queries
-$PISA_BIN/queries -t opt -a block_max_wand -i $ROB_PREFIX.index.opt -w $ROB_PREFIX.wand -q $PISA_PREFIX/test/test_data/queries
-
-# Test indri output
-echo '301:international organized crime' | $PISA_BIN/evaluate_queries -t opt -a block_max_wand -i $ROB_PREFIX.index.opt -w $ROB_PREFIX.wand --documents $ROB_PREFIX.doclex --terms $ROB_PREFIX.termlex
+# Test TREC output
+echo '301:international organized crime' | evaluate_queries -t pefopt -a block_max_wand -i $ROB_PREFIX.index.opt -w $ROB_PREFIX.wand --documents $ROB_PREFIX.doclex --terms $ROB_PREFIX.termlex
--- a/examples/index_robust04.sh
+++ b/examples/index_robust04.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+PISA_PREFIX=~/src/pisa
+PISA_BIN=$PISA_PREFIX/build/bin
+THREADS=20
+
+COLLECTION=(/research/remote/collections/TREC/disk4/fr.gz
+	    /research/remote/collections/TREC/disk4/ft.gz
+	    /research/remote/collections/TREC/disk5/fbis.gz
+	    /research/remote/collections/TREC/disk5/latimes.gz)
+
+ROB_PREFIX=./robust04
+
+# Generate robust04 robust04.terms robust04.urls robust04.documents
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -22,3 +22,5 @@ Testing/
 docs/_build/
 
 node_modules
+
+compile_commands.json
```
amallia, line 26: What is this?

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+PISA_PREFIX=~/src/pisa
```
amallia: Can you get PISA's folder with bash instead?

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+PISA_BIN=$PISA_PREFIX/build/bin
+THREADS=20
```
amallia: Can we default it to the number of available threads?

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+
+COLLECTION=(/research/remote/collections/TREC/disk4/fr.gz
+	    /research/remote/collections/TREC/disk4/ft.gz
+	    /research/remote/collections/TREC/disk5/fbis.gz
+	    /research/remote/collections/TREC/disk5/latimes.gz)
+
+ROB_PREFIX=./robust04
+
+# Generate robust04 robust04.terms robust04.urls robust04.documents
+zcat ${COLLECTION[@]} | $PISA_BIN/parse_collection -j $THREADS -f trectext --content-parser html --stemmer krovetz -o $ROB_PREFIX
```
amallia: Did you test the krovetz stemmer? How is it working for you?

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Generate robust04.docs robust04.freqs robust04.sizes
+$PISA_BIN/invert -i $ROB_PREFIX -o $ROB_PREFIX -j 40 --term-count $(wc -l < $ROB_PREFIX.terms)
```
amallia: Why `-j 40`?

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Generate robust04.index.opt
+$PISA_BIN/create_freq_index -t opt -c $ROB_PREFIX -o $ROB_PREFIX.index.opt --check
```
amallia: `opt` is outdated. We now use `pefopt` (https://github.com/pisa-engine/pisa/pull/192)

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Generate robust04.wand
+$PISA_BIN/create_wand_data -c $ROB_PREFIX -o $ROB_PREFIX.wand
+
+# Test queries
+$PISA_BIN/queries -t opt -a block_max_wand -i $ROB_PREFIX.index.opt -w $ROB_PREFIX.wand -q $PISA_PREFIX/test/test_data/queries
```
amallia: - Same here `opt`->`pefopt`
- this queries are not the one we want to use with `Robust04`. You can use queries with actual terms by adding the lexicon parameter `--terms $ROB_PREFIX.fwd.termlex` (for an example look at https://github.com/osirrc/pisa-docker/blob/master/search#L12-L23)

examples/index_robust04.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Test indri output
```
amallia: Indri?

script/corpora_text.py
```diff
@@ -0,0 +1,89 @@
...
+                        help='Read docno from stdin')
+
+    parser.add_argument('prefix', type=file_path)
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_arguments()
+    corpora = CorporaText(args.prefix)
```
amallia, line 84: Can you clarify what is the input here? Forward index or inverted index?

amallia:
Can you add some documentation about the script for dumping document text?
Is the script able to dump the original document or only the list of distinct terms that are contained in a document?

.gitignore
binshengliu: It is generated by `cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=1`

The file is used by clangd and other tools to parse the source tree and support function jump. People who use it to read code would want to ignore it as its content is local to the compiling system.

https://clang.llvm.org/docs/JSONCompilationDatabase.html

examples/index_robust04.sh
binshengliu: When I started fiddling I couldn't get `make install` work out of the box (there were some dependency errors if I remember correctly), so I run the binaries from the building directory directly.

examples/index_robust04.sh
binshengliu: It worked pretty well for me as far as I can tell according to the end-to-end output.

I got crash when stemming some tokens I manually parsed but I didn't dig into that so not sure what happened exactly.

examples/index_robust04.sh
binshengliu: Thanks, that's good to know. There used to be some dependency errors with `make install`. So I intended to run these binaries to do some sanity check on the index and binary file dependencies. If it's confusing I think we can remove it.

examples/index_robust04.sh
binshengliu: TREC format actually. I'm just used to Indri.

script/corpora_text.py
binshengliu: The input is docnos, the output is the content of the doc dumped from the forward index.

binshengliu:
Yep. I can add some documentation if you think it's useful to include it in the repo. I use it quite often to dump the original document (so I don't need to do tokenization and stemming myself).

examples/index_robust04.sh
binshengliu: The dependency was libtbb.so.2. #242

examples/index_robust04.sh
binshengliu:
```

Comment (binshengliu) [reply examples/index_robust04.sh]:
```
$ wc -l plaintext 
1000 plaintext
$ cat plaintext | parse_collection --batch-size 500 --output ./discard --stemmer krovetz ; rm -f discard*
[2019-08-23 12:18:23.720] [info] Number of threads: 56
Segmentation fault (core dumped)
```

```
(gdb) bt                                                                                                                                                      
#0  0x00007f797c23a244 in std::__1::basic_istream<char, std::__1::char_traits<char> >::sentry::sentry(std::__1::basic_istream<char, std::__1::char_traits<char
> >&, bool) () from /home/sl8/S3676608/.local/lib/libc++.so.1
#1  0x000000000047bd70 in std::__1::operator>><char, std::__1::char_traits<char>, std::__1::allocator<char> > (__is=..., __str=...)
    at /home/sl8/S3676608/.local/bin/../include/c++/v1/istream:1336                                                                                           
#2  0x000000000047488b in pisa::parse_plaintext_content(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&&, std::__1::fun
ction<void (std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&&)>) (content=<optimized out>, process=...)
    at /research/remote/petabyte/users/binsheng/pisa/include/pisa/forward_index_builder.hpp:63
```

The crash turns out to be caused by clang's libc++. Seems to be an upstream bug.

File: .gitignore (hunk 1 above)
Comment (elshize) [reply .gitignore:26]: Yes, this should be ignored.

File: script/corpora_text.py (hunk 9 above)
Comment (elshize) [reply script/corpora_text.py:84]: Isn't this basically the same as this?

```
read_collection - read collections.
Usage: /home/michal/pisa/build/bin/read_collection [OPTIONS] first [last]

Positionals:
  first INT REQUIRED          Element number
  last INT                    Element number

Options:
  -h,--help                   Print this help message and exit
  -c,--collection TEXT REQUIRED
                              Collection file name
  -m,--map TEXT               String map file name
```

This, as far as I remember, works both for forward and inverted index. The advantage of it might be performance. Or does your program do something more I missed?

binshengliu: Great to know. I think I missed this program. I'll try it.

PR #241, opened 2019-08-16
QL3

elshize:
Opening this because I can't make Travis to run on the other branch, so I want to test if it works when I create a new one.


include/pisa/query/algorithm/block_max_maxscore_query.hpp
```diff
@@ -114,12 +114,13 @@ struct block_max_maxscore_query {
 };
 
 template <typename Index, typename Wand, typename Scorer>
-[[nodiscard]] auto block_max_maxscore_executor(Index const &index,
+[[nodiscard]] inline auto block_max_maxscore_executor(Index const &index,
```
elshize: @amallia with `inline`, it's fast but fails to build on mac, without it, it slows down significantly.

elshize:
Preliminary results suggest everything is in order now, but:

- we need more test,
- we need to clean up some unused code.

@amallia

include/pisa/block_posting_list.hpp
```diff
@@ -1,357 +1,393 @@
...
+        // write max
+        *((uint32_t *)&out[begin_block_maxs + 4 * b]) = block.max;
+
+        // copy block
+        block.append_docs_block(out);
+        block.append_freqs_block(out);
+    }
+}
+
+//struct simdbp_block;
```
elshize: Remove the noise

include/pisa/codec/list.hpp
```diff
@@ -0,0 +1,7 @@
+#define PISA_INDEX_TYPES                                                                    \
+    (ef)(single)(pefuniform)(pefopt)(block_optpfor)(block_varintg8iu)(block_streamvbyte)(         \
+        block_maskedvbyte)(block_interpolative)(block_qmx)(block_varintgb)(block_simple8b)( \
+        block_simple16)(block_simdbp)(block_mixed)
+#define PISA_BLOCK_CODEC_TYPES                                                                \
```
elshize, line 5: Is this still used? If not, we should remove.

include/pisa/codec/qmx.hpp
```diff
@@ -1,50 +1,22 @@
...
-        std::vector<uint32_t> buf(2 * n + overflow);
-        qmx_codec.decode(buf.data(), n, in, enc_len);
-        for (size_t i = 0; i < n; ++i)
-        {
-            *out = buf[i];
-            ++out;
-        }
-        return in + enc_len;
-    }
+    static void encode(std::uint32_t const *in,
```
elshize, line 12: @amallia I forgot about this one, it most likely has overhead for block codecs. We can make it inline again, but tbh I'm kind of okay with the slowdown for the block codecs. Take a look at the results and tell me what you think.

include/pisa/forward_index_builder.hpp
```diff
@@ -28,7 +28,7 @@
 #include "parsing/html.hpp"
 #include "payload_vector.hpp"
 #include "type_safe.hpp"
-#include "tokenizer.hpp"
+#include "tokenizer/term_tokenizer.hpp"
```
elshize: It's a library now, let's move it up and surround with brackets to indicate that.

include/pisa/invert.hpp
```diff
@@ -232,7 +232,7 @@ auto invert_range(gsl::span<gsl::span<Term_Id const>> documents,
                                  size_t             threads) -> uint32_t
 {
```
elshize: clang-format

include/pisa/payload_vector.hpp
```diff
@@ -208,7 +207,9 @@ struct Payload_Vector_Buffer {
 };
 
 template <typename InputIterator, typename PayloadEncodingFn>
-auto encode_payload_vector(InputIterator first, InputIterator last, PayloadEncodingFn encoding_fn)
+auto encode_payload_vector(InputIterator first,
```
elshize, line 210: clang-format

include/pisa/score_opt_partition.hpp
```diff
@@ -48,8 +46,7 @@ namespace pisa {
                     , m_fixed_cost(fixed_cost)
                     , sum(0)
                     , end_sequence(size)
-                    , element_count(0)
-                    , estimated_idf(e_idf) {}
```
elshize, line 52: @amallia I'm sure I've asked you about it and that's why I removed it, but can you confirm this is something we don't need?

include/pisa/score_opt_partition.hpp
```diff
@@ -109,19 +106,18 @@ namespace pisa {
...
             float max = 0;
             float sum = 0;
             std::for_each(begin, begin + size, [&](auto& b) {
               max = std::max(max, std::get<1>(b));
-              sum += std::get<1>(b) * estimated_idf;
+              sum += std::get<1>(b);
             });
 
-            wand_cost_t single_block_cost = size*max*estimated_idf - sum;
+            wand_cost_t single_block_cost = size*max - sum;
```
elshize: Run clang-format on the entire file.

include/pisa/scorer/bm25.hpp
```diff
@@ -1,30 +1,59 @@
...
+template <typename Wand>
+struct bm25 : public index_scorer<Wand> {
+    static constexpr float b = 0.4;
+    static constexpr float k1 = 0.9;
+
+    using index_scorer<Wand>::index_scorer;
+
+    static float doc_term_weight(uint64_t freq, float norm_len)
+    {
+        float f = (float)freq;
```
elshize: `static_cast`

include/pisa/scorer/index_scorer.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace pisa {
+
+template <class Iter>
+struct scorer_traits;
+
+using term_scorer_t = std::function<float(uint32_t, uint32_t)>;
+
+template <typename Wand>
+struct index_scorer {
```
elshize, line 16: Are we still using it?

include/pisa/topk_queue.hpp
```diff
@@ -1,15 +1,68 @@
...
 #include <algorithm>
-#include "util/util.hpp"
+#include <cstdint>
+#include <vector>
+
 #include "util/likely.hpp"
 
 namespace pisa {
 
+/* struct TopKQueue { */
```
elshize: Let's use this refactored version?

tools/awesome_queries.cpp
```diff
@@ -0,0 +1,170 @@
+#include <functional>
```
elshize: Remove this file

include/pisa/codec/list.hpp
elshize: Scratch that, it **is** used.

include/pisa/scorer/index_scorer.hpp
elshize: I guess we are now using it as a base class for all scorers, but only to reuse some code. I removed some unnecessary includes from unrelated files (coming up in next push).

elshize:
The most recent commit should fix `ranked_or`; I executed `ranked_and` instead, hence the difference in results. I'm looking into TAAT now.

elshize:
Actually, it turns out that it's the `master` results for TAAT are lower than others, not the other way around. So I actually fixed it somehow. Anyway, I'll confirm the `ranked_or` fix later, because I'm now building Clueweb09B for `ql3` to verify results on that and look into performance.

include/pisa/accumulator/simple_accumulator.hpp
```diff
@@ -1,23 +1,26 @@
...
-    void init() { std::fill(begin(), end(), 0.0); }
-    void accumulate(uint32_t doc, float score) { operator[](doc) += score; }
-    void aggregate(topk_queue &topk) {
-        uint64_t docid = 0u;
-        std::for_each(begin(), end(), [&](auto score) { 
-            if(topk.would_enter(score)) { topk.insert(score, docid); } 
-            docid += 1;
-        });
-    }
+struct TopKQueue;
```
amallia, line 9: Why this? `topk_queue.hpp` is pretty small overall

elshize: Because we don't need it, so there's no reason to parse it at this point. I agree that in this particular situation we won't get much, but these things compound. But most importantly, our imports tend to be unnecessarily bloated, so reducing them is not a bad idea.

And most importantly, why not? Are we gaining anything by adding the include?

tools/evaluate_queries/def.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include "freq_index.hpp"
+#include "index_types.hpp"
+#include "query/queries.hpp"
+#include "scorer/bm25.hpp"
+#include "scorer/dph.hpp"
+#include "scorer/pl2.hpp"
+#include "scorer/qld.hpp"
+
+namespace pisa {
+
```
amallia, line 12: I guess `def.hpp` stays for defines. Shall we rename it to `defines.hpp` or something similar?

tools/evaluate_queries/ranked.cpp
```diff
@@ -0,0 +1,81 @@
...
+    PISA_RANKED_OR_QUERY_LOOP(bm25, T, wand_data_compressed) \
+    PISA_RANKED_OR_QUERY_LOOP(dph, T, wand_data_compressed)  \
+    PISA_RANKED_OR_QUERY_LOOP(pl2, T, wand_data_compressed)  \
+    PISA_RANKED_OR_QUERY_LOOP(qld, T, wand_data_compressed)  \
+/**/
+BOOST_PP_SEQ_FOR_EACH(LOOP_BODY, _, PISA_INDEX_TYPES);
+#undef LOOP_BODY
+#undef PISA_RANKED_OR_QUERY_LOOP
+
+#define PISA_RANKED_AND_QUERY_LOOP(SCORER, INDEX, WAND)                                \
```
amallia, line 44: Why is this a macro? Can't be a function?

elshize: It could but, imo, it ends up being messier because you still have to define explicit specialization to compile in the same unit. And so it looks something like this:

1. Introduce a function template for AND:

```cpp
template <typename Index, typename Scorer>
auto query_loop_and(Index const &index, Scorer scorer, std::vector<Query> const &queries, int k)
    -> std::vector<ResultVector>
{
    std::vector<ResultVector> results(queries.size());
    auto run = ranked_and_query(k);
    for (std::size_t qidx = 0; qidx < queries.size(); ++qidx) {
        auto query = queries[qidx];
        auto cursors = make_scored_cursors(index, scorer, query);
        run(gsl::make_span(cursors), index.num_docs());
        results[qidx] = run.topk();
    }
    return results;
}
```
2. Explicit instantiation.
```cpp
#define LOOP_BODY(R, DATA, T)                                                     \
    template <>                                                                   \
    auto query_loop_and<BOOST_PP_CAT(T, _index), bm25<wand_data<wand_data_raw>>>( \
        BOOST_PP_CAT(T, _index) const &index,                                     \
        bm25<wand_data<wand_data_raw>> scorer,                                    \
        std::vector<Query> const &queries,                                        \
        int k)                                                                    \
        ->std::vector<ResultVector>;                                              \
    /* REPEATED FOR ALL (scorer, wand) */
/**/
BOOST_PP_SEQ_FOR_EACH(LOOP_BODY, _, PISA_INDEX_TYPES);
#undef LOOP_BODY
```

Note that the last part you need to repeat 8 times, and each time the scorer and wand data are in two different places far from each other; it's much more difficult to maintain.

Unless you have yet another idea how to do this...

tools/evaluate_queries/def.hpp
elshize: In that case, I would just go with `definitions.hpp`

PR #244, opened 2019-09-04
Update query processing

gustingonzalez:
- Refactor queries.hpp to avoid tokenization when using query ids
- Make the 'stopwords' parameter require 'terms' option
- Fix term existence checking when stopwords are load in term processor

### DIFF
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -33,7 +33,7 @@ struct Query {
     std::vector<float> term_weights;
 };
 
-[[nodiscard]] auto query_to_raw(std::string query_string, std::optional<std::string> &id)
+[[nodiscard]] auto query_to_raw(std::string const &query_string, std::optional<std::string> &id)
 {
     // query id : terms (or ids)
     auto colon = std::find(query_string.begin(), query_string.end(), ':');
@@ -91,11 +91,11 @@ struct Query {
     return {id, parsed_query, {}};
 }
 
-[[nodiscard]] std::function<void(std::string)> compute_parse_query_function(
+[[nodiscard]] std::function<void(const std::string)> compute_parse_query_function(
     std::vector<Query> &queries,
-    std::optional<std::string> terms_file,
-    std::optional<std::string> stopwords_filename,
-    std::optional<std::string> stemmer_type) {
+    const std::optional<std::string> terms_file,
+    const std::optional<std::string> stopwords_filename,
+    const std::optional<std::string> stemmer_type) {
     if (terms_file) {
         auto term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type);
         return [&](std::string const &query_line) {
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -55,19 +55,15 @@ struct Query {
     std::vector<term_id_type> parsed_query;
     for (auto term_iter = tokenizer.begin(); term_iter != tokenizer.end(); ++term_iter) {
         auto raw_term = *term_iter;
-        try {
-            auto term = term_processor(raw_term);
-            if (term) {
-                if (!term_processor.is_stopword(*term)) {
-                    parsed_query.push_back(std::move(*term));
-                } else {
-                    spdlog::warn("Term `{}` is a stopword and will be ignored", *term);
-                }
+        auto term = term_processor(raw_term);
+        if (term) {
+            if (!term_processor.is_stopword(*term)) {
+                parsed_query.push_back(std::move(*term));
             } else {
-                spdlog::warn("Term `{}` not found and will be ignored", raw_term);
+                spdlog::warn("Term `{}` is a stopword and will be ignored", *term);
             }
-        } catch (std::invalid_argument &err) {
-            spdlog::warn("Could not parse `{}` to a number", raw_term);
+        } else {
+            spdlog::warn("Term `{}` not found and will be ignored", raw_term);
         }
     }
     return {std::move(id), std::move(parsed_query), {}};
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -94,7 +94,7 @@ class TermProcessor {
             if (stopwords_filename) {
                 std::ifstream is(*stopwords_filename);
                 io::for_each_line(is, [&](auto &&word) {
-                    if (auto processed_term = process(std::move(word)); process) {
+                    if (auto processed_term = process(std::move(word)); processed_term.has_value()) {
                         stopwords.insert(*processed_term);
                     }
                 });
@@ -104,6 +104,14 @@ class TermProcessor {
         bool is_stopword(term_id_type term)
         {
             return stopwords.find(term) != stopwords.end();
+        }
+
+        std::vector<term_id_type> get_stopwords()
+        {
+            std::vector<term_id_type> v;
+            v.insert(v.end(), stopwords.begin(), stopwords.end());
+            sort(v.begin(), v.end());
+            return v;
         }
 };
 
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -70,25 +70,23 @@ struct Query {
             spdlog::warn("Could not parse `{}` to a number", raw_term);
         }
     }
-    return {id, parsed_query, {}};
+    return {std::move(id), std::move(parsed_query), {}};
 }
 
 [[nodiscard]] auto parse_query_ids(std::string const &query_string) -> Query
 {
     auto [id, raw_query] = split_query_at_colon(query_string);
     std::vector<term_id_type> parsed_query;
-    std::vector<std::string> splitted_query;
-    boost::split(splitted_query, raw_query, boost::is_any_of("\t"));
+    std::vector<std::string> term_ids;
+    boost::split(term_ids, raw_query, boost::is_any_of("\t"));
     try {
-        std::transform(splitted_query.begin(),
-                       splitted_query.end(),
-                       std::back_inserter(parsed_query),
-                       [](const std::string &val) { return std::stoi(val); });
+        auto to_int = [](const std::string &val) { return std::stoi(val); };
+        std::transform(term_ids.begin(), term_ids.end(), std::back_inserter(parsed_query), to_int);
     } catch (std::invalid_argument &err) {
         spdlog::error("Could not parse term identifiers of query `{}`", raw_query);
         exit(1);
     }
-    return {id, parsed_query, {}};
+    return {std::move(id), std::move(parsed_query), {}};
 }
 
 [[nodiscard]] std::function<void(const std::string)> compute_parse_query_function(
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -1,26 +1,20 @@
 #pragma once
 
-#include <functional>
 #include <iostream>
 #include <memory>
 #include <optional>
 #include <sstream>
 #include <string>
 #include <string_view>
-#include <unordered_map>
 
-#include <KrovetzStemmer/KrovetzStemmer.hpp>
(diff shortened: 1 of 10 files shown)


include/pisa/query/queries.hpp
```diff
@@ -32,37 +32,116 @@ namespace pisa {
...
+            else
+            {
+                throw std::invalid_argument("Unknown stemmer");
+            }
+
+            // Loads stopwords.
+            if (stopwords_filename) {
+                std::ifstream is(*stopwords_filename);
+                io::for_each_line(is, [&](auto &&word) {
+                    if (auto processed_term = process(std::move(word)); processed_term.has_value()) {
```
gustingonzalez: Previous to this fix, if a stopword was not present in "terms", it was assigned the id "164320464 (0x9cb54d0)".

gustingonzalez: Previous to this fix, if a stopword was not present in "terms", it was assigned the id "4.294.953.632".

include/pisa/query/queries.hpp
```diff
@@ -32,37 +32,116 @@ namespace pisa {
...
     std::optional<std::string> id;
     std::vector<term_id_type>  terms;
     std::vector<float>  term_weights;
 };
 
-[[nodiscard]] inline auto parse_query(
-    std::string const &query_string,
-    TermProcessor process_term,
-    std::optional<std::unordered_set<term_id_type>> const &stopwords = std::nullopt) -> Query
+class TermProcessor {
```
amallia: Can we move it to a separate file (maybe `term_processor.hpp`)?

include/pisa/query/queries.hpp
```diff
@@ -74,16 +153,48 @@ struct Query {
...
+        std::transform(splitted_query.begin(), splitted_query.end(), std::back_inserter(parsed_query),
+            [](const std::string& val) {return std::stoi(val);});
+    } catch (std::invalid_argument& err) {
+        spdlog::error("Could not parse term identifiers of query `{}`", raw_query);
+        exit(1);
+    }
+    return {id, parsed_query, {}};
+}
+
+[[nodiscard]] std::function<void(std::string)> compute_parse_query_function(
```
amallia: why note just `parse_query`?

amallia:
@elshize you have reviewed the other PR, so it would be better if you also give a look to this one.

include/pisa/query/queries.hpp
elshize: A general comment: we use `clang-format` to format source files (https://clang.llvm.org/docs/ClangFormat.html). There is already a `.clang-format` in the repo you should use. Most likely, there is a plugin for the editor/IDE of your choosing, e.g., https://github.com/rhysd/vim-clang-format for `vim`.

There are some files that are just not formatted this way, and this is one of them, so don't be thrown off by the fact that you see a lot of changes after running it.

include/pisa/query/queries.hpp
```diff
@@ -32,37 +32,116 @@ namespace pisa {
...
-    std::optional<std::unordered_set<term_id_type>> const &stopwords = std::nullopt) -> Query
+class TermProcessor {
+    private:
+        std::unordered_set<term_id_type> stopwords;
+
+    public:
+        // Method implemented in constructor according to the specified stemmer.
+        std::function<std::optional<term_id_type>(std::string)> process;
+
+        TermProcessor(std::optional<std::string> terms_file,
```
elshize: Since we're not owning the arguments, let's make these `const &`.

include/pisa/query/queries.hpp
```diff
@@ -32,37 +32,116 @@ namespace pisa {
...
-    std::string const &query_string,
-    TermProcessor process_term,
-    std::optional<std::unordered_set<term_id_type>> const &stopwords = std::nullopt) -> Query
+class TermProcessor {
+    private:
+        std::unordered_set<term_id_type> stopwords;
+
+    public:
+        // Method implemented in constructor according to the specified stemmer.
+        std::function<std::optional<term_id_type>(std::string)> process;
```
elshize: I don't know how I feel about having this exposed like this. How about we move it to `private`, and expose a method:
```cpp
[[nodiscard]] auto process(std::string term) -> std::optional<term_id_type>
{
    return m_process(std::move(term));
}
```
In fact, I feel like `process` is redundant, that is, it doesn't really add any information, so maybe an `operator()` would work here better? @amallia what is your opinion?

include/pisa/query/queries.hpp
```diff
@@ -32,37 +32,116 @@ namespace pisa {
...
+        std::vector<term_id_type> get_stopwords()
+        {
+            std::vector<term_id_type> v;
+            v.insert(v.end(), stopwords.begin(), stopwords.end());
+            sort(v.begin(), v.end());
+            return v;
+        }
+};
+
+[[nodiscard]] auto query_to_raw(std::string query_string, std::optional<std::string> &id)
```
elshize: Several things here:
1. Let's return a pair `(id, raw_query)` instead of taking a reference to `id`.
2. Let's take query by `const &`.
3. We can return the ID as `std::optional<std::string>` and then `move` it later to the actual `Query`.
4. We should return the part after the colon as a `string_view` to avoid copying. `TermTokenizer` already takes a view anyway.
5. I think we can do better job naming the function to explain what we're doing; something like `split_at_colon` would be clearer.

For example:

```cpp
[[nodiscard]]  auto split_at_colon(std::string const& query_string) -> std::pair<std::optional<std::string>, std::string_view>;
```

include/pisa/query/queries.hpp
```diff
@@ -74,16 +153,48 @@ struct Query {
     return {id, parsed_query, {}};
```
elshize: Let's move:
```cpp
return {std::move(id), std::move(parsed_query), {}};
```

include/pisa/query/queries.hpp
```diff
@@ -74,16 +153,48 @@ struct Query {
...
-bool read_query(term_id_vec &ret, std::istream &is = std::cin,
-                std::function<term_id_type(std::string)> process_term = [](auto str) {
-                    return std::stoi(str);
-                }) {
+[[nodiscard]] auto parse_query_ids(std::string const &query_string) -> Query
+{
+    std::optional<std::string> id = std::nullopt;
+    std::vector<term_id_type> parsed_query;
+    auto raw_query = query_to_raw(query_string, id);
+    std::vector<std::string> splitted_query;
```
elshize: `splitted` is very archaic; how about we just call it `term_ids`?

include/pisa/query/queries.hpp
```diff
@@ -74,16 +153,48 @@ struct Query {
...
+    boost::split(splitted_query, raw_query, boost::is_any_of("\t"));
+    try {
+        std::transform(splitted_query.begin(), splitted_query.end(), std::back_inserter(parsed_query),
+            [](const std::string& val) {return std::stoi(val);});
+    } catch (std::invalid_argument& err) {
+        spdlog::error("Could not parse term identifiers of query `{}`", raw_query);
+        exit(1);
+    }
+    return {id, parsed_query, {}};
```
elshize: Again, should move.

include/pisa/query/queries.hpp
```diff
@@ -74,16 +153,48 @@ struct Query {
...
+}
+
+[[nodiscard]] std::function<void(std::string)> compute_parse_query_function(
+    std::vector<Query> &queries,
+    std::optional<std::string> terms_file,
```
elshize: Same `const &` comment.

elshize (changes requested):
Left some comments, but great job overall!

include/pisa/query/queries.hpp
gustingonzalez: It's true, the method is exposed and can be overwritten externally, and it should not be like that.

amallia (changes requested):

include/pisa/query/queries.hpp
gustingonzalez: The function doesn't parse the query, it only computes and then returns a function that does it (depending on whether terms or their ids are being used). Even so, the name could be improved, what do you think about "compute_query_parser"?

include/pisa/query/queries.hpp
```diff
@@ -32,70 +26,112 @@ namespace pisa {
...
     auto pos = colon == query_string.end() ? query_string.begin() : std::next(colon);
     auto raw_query = std::string_view(&*pos, std::distance(pos, query_string.end()));
-    TermTokenizer tokenizer(raw_query);
+    return {std::move(id), std::move(raw_query)};
+}
+
+[[nodiscard]] auto parse_query_terms(std::string const &query_string, TermProcessor term_processor)
+    -> Query
+{
+    auto id_and_raw_query = split_query_at_colon(query_string);
```
elshize: In c++17, we can destruct these:
```cpp
auto [query_id, raw_query] = split_query_at_colon(query_string);
```
which is much cleaner in this case.

include/pisa/query/queries.hpp
```diff
@@ -32,70 +26,112 @@ namespace pisa {
...
     }
-    return {id, parsed_query, {}};
+    return {std::move(id_and_raw_query.first), std::move(parsed_query), {}};
+}
+
+[[nodiscard]] auto parse_query_ids(std::string const &query_string) -> Query
+{
+    auto id_and_raw_query = split_query_at_colon(query_string);
+    auto id = id_and_raw_query.first;
+    auto raw_query = id_and_raw_query.second;
```
elshize: Same as above, we can destruct in one line.

include/pisa/query/queries.hpp
```diff
@@ -32,70 +26,112 @@ namespace pisa {
...
+
+    std::vector<term_id_type> parsed_query;
+    std::vector<std::string> term_ids;
+    boost::split(term_ids, raw_query, boost::is_any_of("\t"));
```
elshize: I was thinking, let's maybe split on any whitespace? I don't see a reason why this should be strictly a tab.

include/pisa/query/queries.hpp
```diff
@@ -32,70 +26,112 @@ namespace pisa {
...
+        return [&](std::string const &query_line) {
+            queries.push_back(parse_query_terms(query_line, term_processor));
+        };
+    } else {
+        return
+            [&](std::string const &query_line) { queries.push_back(parse_query_ids(query_line)); };
+    }
+}
+
+bool read_query(term_id_vec &ret, std::istream &is = std::cin)
```
elshize, line 109: So here this is confusing, I don't see any reason why such a function should exist but not support string terms; that's weird. Are there many places we use this one? If not, maybe we can get rid of it altogether---my feeling is that parsing a query line (like the function above does) should be decoupled from reading actual lines. This can be done by any number of ways, but most notably there is a `io::for_each_line` function that is used in many places anyway. I think this is a better approach.

elshize (changes requested):
Looking better, a few more comments.

include/pisa/query/queries.hpp
elshize: I personally would go for `resolve_query_parser`.

include/pisa/query/queries.hpp
gustingonzalez: I check this function isn't used anywhere (instead it is used `io::for_each_line`). Thus, probably it must be removed.

elshize: Yeah, if not used, let's remove it. It's legacy code.

gustingonzalez: @elshize, I was wrong. This function is used in `selective_queries` and `profile_queries`.

elshize: Ok, that's cool, let's just leave it. I'll create a separate issue for that so we don't forget, we can take any relevant discussion there.

elshize:
For reference: #255

elshize (approved):
Looks good, thanks, great work, and really appreciated!

PR #257, opened 2019-09-21
Bad alloc

### LINKED ISSUE #256: bad_alloc when using query terms
`queries` with `--terms` crashes.

I've more or less found the problem, but wasn't able to finish the fix tonight.

elshize:
Fixes #256

### DIFF
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -60,7 +60,7 @@ struct Query {
             if (!term_processor.is_stopword(*term)) {
                 parsed_query.push_back(std::move(*term));
             } else {
-                spdlog::warn("Term `{}` is a stopword and will be ignored", *term);
+                spdlog::warn("Term `{}` is a stopword and will be ignored", raw_term);
             }
         } else {
             spdlog::warn("Term `{}` not found and will be ignored", raw_term);
@@ -92,12 +92,13 @@ struct Query {
 
 [[nodiscard]] std::function<void(const std::string)> compute_parse_query_function(
     std::vector<Query> &queries,
-    const std::optional<std::string> terms_file,
-    const std::optional<std::string> stopwords_filename,
-    const std::optional<std::string> stemmer_type) {
+    std::optional<std::string> const &terms_file,
+    std::optional<std::string> const &stopwords_filename,
+    std::optional<std::string> const &stemmer_type)
+{
     if (terms_file) {
-        auto term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type);
-        return [&](std::string const &query_line) {
+        return [&, term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type)](
+                   std::string const &query_line) {
             queries.push_back(parse_query_terms(query_line, term_processor));
         };
     } else {
--- a/include/pisa/query/queries.hpp
+++ b/include/pisa/query/queries.hpp
@@ -90,20 +90,22 @@ struct Query {
     return {std::move(id), std::move(parsed_query), {}};
 }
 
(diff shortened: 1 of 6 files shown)


include/pisa/query/queries.hpp
```diff
@@ -92,12 +92,13 @@ struct Query {
...
-    const std::optional<std::string> stopwords_filename,
-    const std::optional<std::string> stemmer_type) {
+    std::optional<std::string> const &terms_file,
+    std::optional<std::string> const &stopwords_filename,
+    std::optional<std::string> const &stemmer_type)
+{
     if (terms_file) {
-        auto term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type);
-        return [&](std::string const &query_line) {
+        return [&, term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type)](
```
gustingonzalez: Sorry for this bug. This is due to `term_processor` is destroyed after exiting of the function, isn't it? In the other hand, looking again, would it be an adventage capturing only `&queries` instead `&`?

include/pisa/query/queries.hpp
```diff
@@ -92,12 +92,13 @@ struct Query {
 
 [[nodiscard]] std::function<void(const std::string)> compute_parse_query_function(
```
gustingonzalez: In the last PR it was pending to simplify the name of this function to `compute_query_parser`, `resolve_query_parser` or similar. Maybe this bug fix could be used to improve this too.

include/pisa/query/queries.hpp
gustingonzalez: Sorry for this bug introduced after refactoring. This is due to `term_processor` is destroyed after exiting of the function, isn't it? In the other hand, looking again, would it be an advantage capturing only `&queries` instead `&`?

include/pisa/query/queries.hpp
```diff
@@ -92,12 +92,13 @@ struct Query {
...
-    const std::optional<std::string> terms_file,
-    const std::optional<std::string> stopwords_filename,
-    const std::optional<std::string> stemmer_type) {
+    std::optional<std::string> const &terms_file,
+    std::optional<std::string> const &stopwords_filename,
+    std::optional<std::string> const &stemmer_type)
+{
     if (terms_file) {
-        auto term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type);
-        return [&](std::string const &query_line) {
```
gustingonzalez, line 100: Sorry for this bug introduced after refactoring. I understand that it was caused because `term_processor` was destroyed after exiting of the function, is that correct?

In the other hand, would it be an advantage capturing only `&queries` instead `&`?

elshize: It might be a good idea to capture things one by one, especially when capturing by reference. Nevertheless, we have to move term processor into a generalized capture because there's no capture-by-move operator, like `&` for references and `=` for copies. the alternative to what I did is:
```cpp
auto term_processor = TermProcessor(terms_file, stopwords_filename, stemmer_type);
return [&queries, term_processor = std::move(term_processor)](
           std::string const &query_line) {
    queries.push_back(parse_query_terms(query_line, term_processor));
};
```
Here, you explicitly move what is previously constructed.

elshize: On another note, we should be using some static code analyzer in our CI to find these things. I'm reasonably sure, say, `clang-tidy` would find this.

amallia:
What I see in this PR:
- a fix for an UB (return a destroyed object)
- renamed function with something more meaningful 
- better code formatting

Travis was complaining and I should have fixed it. 

It looks good to me and thanks for the PR. We can merge 👍

PR #266, opened 2019-10-12
Added sampling task

amallia:
Simple task to sample an uncompressed inverted index.
Given an input, an output basename and a rate it produces `.docs`, `.freqs` and `.sizes`.
`.sizes` is just a copy of the original one, while `.docs` and `.freqs`are samples of the originals.

Note: the minimum size of a posting list is 1. The size of the sampled lists is the ceil of the original size times the sampling rate.


src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+
+int main(int argc, char **argv) {
+
+    using namespace pisa;
+    std::string input_basename;
+    std::string output_basename;
+    float rate;
+
+    CLI::App app{"sample_inverted_index - a tool for sampling an inverted index."};
```
elshize: I think we should leave out the name of the program for two reasons:

1. They are unnecessary, as the name of the program will be right above and right below, e.g.:
```
$ ./bin/evaluate_queries -h
Retrieves query results in TREC format.
Usage: ./bin/evaluate_queries [OPTIONS]
```
so it's redundant.

2. It introduces unnecessary possibility of diverging printed info from the actual program name (if we change it, we need to change the doc).

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
+#include <string>
+#include <random>
+#include <fstream>
+
+#include "invert.hpp"
+
+#include "binary_freq_collection.hpp"
+
+#include "util/progress.hpp"
```
elshize: Is there a reason for these extra spaces?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    using namespace pisa;
+    std::string input_basename;
+    std::string output_basename;
+    float rate;
+
+    CLI::App app{"sample_inverted_index - a tool for sampling an inverted index."};
+    app.add_option("-c,--collection", input_basename, "Input collection basename")->required();
+    app.add_option("-o,--output", output_basename, "Output collection basename")->required();
+    app.add_option("-r,--rate", rate, "Sampling rate")->required();
```
elshize: Maybe we should add extra info, I feel like it might not be clear right away for everyone. Something like "proportional size of the produced index." Something that clearly describes what it is. It could even be several sentences if need be. It's better to be precise than brief, what do you think?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+    CLI11_PARSE(app, argc, argv);
+
+    if(rate < 0 or rate > 1)
```
elshize: `clang-format`

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+        std::abort();
+    }
+    binary_freq_collection input(input_basename.c_str());
+
+    // Copy '.sizes' file
+    std::ifstream  sis(input_basename + ".sizes");
+    std::ofstream  sos(output_basename + ".sizes");
+    sos << sis.rdbuf();
+    sis.close();
+    sos.close();
```
elshize: How about, instead of the comment, we use
```cpp
boost::filesystem::copy_file(
    fmt::format("{}.sizes", input_basename),
    fmt::format("{}.sizes", output_basename),
    boost::filesystem::copy_option::overwrite_if_exists
);
```

elshize: Incidentally, I think we should make use of the `fmt` library when it comes to string formatting.

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+        pisa::progress progress("Sampling inverted index", input.size());
+        for (auto const &plist : input) {
+            std::vector<uint32_t> docs(plist.docs.begin(), plist.docs.end());
+            std::vector<uint32_t> freqs(plist.freqs.begin(), plist.freqs.end());
+
+            auto seed = unsigned ( std::time(0) );
+            std::mt19937 eng1(seed);
+            auto eng2 = eng1;
+            std::shuffle(docs.begin(), docs.end(), eng1);
+            std::shuffle(freqs.begin(), freqs.end(), eng2);
```
elshize: I'm almost sure copying here is not necessary. It's passed by a forwarding reference, not an rvalue reference.

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,72 @@
...
+            auto new_size = plist.docs.size() * rate;
```
elshize: This `new_size` ends up being a `float` and then implicitly turned into an `int` later. Let's make it clear that this is a size:
```cpp
auto new_size = static_cast<std::size_t>(plist.docs.size() * rate);
```

src/sample_inverted_index.cpp
elshize: This is actually incorrect. You're sampling documents and frequencies independently, so they don't match.

Moreover, there is a `std::sample` function that will do the sampling for you without having to copy and shuffle the entire lists. Plus, it is stable, so there's no need to sort after.

Here's a quick example of how it would work: https://gcc.godbolt.org/z/GAgeFP

elshize (changes requested):
Can we move list sampling to the library and test? I'm thinking of a function that will take an input collection, output streams (or even file names), and a sampling function as a template parameter.

```cpp
template <typename SampleFn>
void sample_index(binary_collection const&, std::ostream&, std::ostream&, SampleFn&&);
```

Then, we can unit-test it with a stub sample function.

Also, we should add an optional `--seed` parameter to the program, so that we are able to have reproducible results if desired.

src/sample_inverted_index.cpp
amallia: Okay!

src/sample_inverted_index.cpp
amallia: No, just forgot to do a clang-format pass

src/sample_inverted_index.cpp
amallia: Sure. great snippet.

src/sample_inverted_index.cpp
amallia: It should be correct as I am shuffling using the same seed and state. This is why I have eng1 and eng2. It might not be the cleanest, but it should work

src/sample_inverted_index.cpp
amallia: Sure thanks

src/sample_inverted_index.cpp
elshize: Oh, that's why. Again, take a look at the link, we can avoid copying all data and sorting.

elshize: To be more precise, we still have to copy indices, but it's fewer copies and no sorting necessary.

elshize:
Two additional points about sampling.

First, there's no need to use actual vector of indices. In another branch that isn't merged yet, I introduced this little file `int_iter.hpp` (https://github.com/pisa-engine/pisa/blob/and-lookup/include/pisa/int_iter.hpp) that defines an "integer iterator", so that you can now do this:

```cpp
std::smaple(iter(0), iter(len), ...);
```

That said, I feel like this is still not as efficient as it gets, because the algorithm doesn't know that it iterates over a range of consecutive numbers.

I googled a bit and found this Floyd's sampling algorithm that is supposed to be proven to be statistically sound as well:

```cpp
std::unordered_set<int> pickSet(int N, int k, std::mt19937& gen)
{
    std::unordered_set<int> elems;
    for (int r = N - k; r < N; ++r) {
        int v = std::uniform_int_distribution<>(1, r)(gen);

        // there are two cases.
        // v is not in candidates ==> add it
        // v is in candidates ==> well, r is definitely not, because
        // this is the first iteration in the loop that we could've
        // picked something that big.

        if (!elems.insert(v).second) {
            elems.insert(r);
        }   
    }
    return elems;
}
```
(copied from https://stackoverflow.com/questions/28287138/c-randomly-sample-k-numbers-from-range-0n-1-n-k-without-replacement)

More info about the algorithm: https://blog.acolyer.org/2018/01/30/a-sample-of-brilliance/

Stripped of the comments, it's a few lines of code, so I think it's worth considering. We can put it in `stats.hpp` or something.

With that, our algorithm would take a sample, sort it (I think we can even afford using `std::set` to already have them sorted), and then remap accordingly.

What do you think about this approach?

elshize:
Actually, scratch my comment about `std::sample` being inefficient on a forward iterator. They also only need index of an object and so it happens to be also its value. So I think the `int_iter` solution is perfectly fine if you want to add that file. It will definitely be useful in other places when you would otherwise use `std::iota` but don't need to allocate.

elshize:
Note that this most likely can be solved by ranges, but this is much more lightweight.

src/sample_inverted_index.cpp
amallia: I decided to go with `std::sample` rather than more complicated solutions since it is not critical

amallia:
@elshize updated. We are missing tests right now. But it is not part of PISA library for now. 
Once we move it to the library I will write some tests, but for now I propose to merge it as it is

external/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ target_compile_options(ParallelSTL INTERFACE -Wno-unused-parameter
                                            -Wno-sign-compare
                                            -Wno-reorder
                                            -Wno-unused-local-typedefs
+                                           -Wno-unused-variable
```
elshize, line 80: Why?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,93 @@
...
+                    sample_size,
+                    std::mt19937{std::random_device{}()});
+
+        std::vector<std::uint32_t> sampled_docs;
+        std::vector<std::uint32_t> sampled_freqs;
+        for (auto index : sample) {
+            sampled_docs.push_back(docs[index]);
+            sampled_freqs.push_back(freqs[index]);
+        }
+        return std::make_pair(sampled_docs, sampled_freqs);
```
elshize: Since we're using a function and passing an lvalue, doesn't this mean we'll have to copy? It's happens at passing param to function, so I don't think we can count on RVO, can we?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,93 @@
...
+    app.add_option("--seed", seed, "Seed state");
+    CLI11_PARSE(app, argc, argv);
+
+    if (rate < 0 or rate > 1) {
+        spdlog::error("Sampling rate should be between 0 and 1.");
+        std::abort();
+    }
+
+    auto random_sampling = [&](auto &docs, auto &freqs) {
+        auto sample_size = std::ceil(docs.size() * rate);
```
elshize: Again, `std::ceil` returns a float and that's what you're passing to `std::sample`, which is why compilation fails on Travis.

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    {
+        pisa::progress progress("Sampling inverted index", input.size());
+        for (auto const &plist : input) {
+            std::vector<uint32_t> docs(plist.docs.begin(), plist.docs.end());
+            std::vector<uint32_t> freqs(plist.freqs.begin(), plist.freqs.end());
+
+            std::vector<std::uint32_t> sampled_docs;
+            std::vector<std::uint32_t> sampled_freqs;
+            tie(sampled_docs, sampled_freqs) = sample_fn(docs, freqs);
```
elshize: Does this not work?
```cpp
auto [sampled_docs, sampled_freqs] = sample_fn(docs, freqs);
```
I imagine it should since you're later only passing these as references to span. Am I missing something?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,93 @@
...
+    std::ofstream fos(output_basename + ".freqs");
+
+    auto document_count = static_cast<uint32_t>(input.num_docs());
+    write_sequence(dos, gsl::make_span<uint32_t const>(&document_count, 1));
+
+    {
+        pisa::progress progress("Sampling inverted index", input.size());
+        for (auto const &plist : input) {
+            std::vector<uint32_t> docs(plist.docs.begin(), plist.docs.end());
+            std::vector<uint32_t> freqs(plist.freqs.begin(), plist.freqs.end());
```
elshize: Are sequences in the collection not random access? If not, they should (open an issue?). If they are, we can just pass them instead of vectors, right?

elshize (changes requested):
> @elshize updated. We are missing tests right now. But it is not part of PISA library for now.
> Once we move it to the library I will write some tests, but for now I propose to merge it as it is

It might not be a part of the library, but isn't part of your research? Even assuming you thoroughly tested it manually, a regression test would be very useful, no?

src/sample_inverted_index.cpp
```diff
@@ -0,0 +1,93 @@
...
+        ->required();
+    app.add_option("--seed", seed, "Seed state");
+    CLI11_PARSE(app, argc, argv);
+
+    if (rate < 0 or rate > 1) {
+        spdlog::error("Sampling rate should be between 0 and 1.");
+        std::abort();
+    }
+
+    auto random_sampling = [&](auto &docs, auto &freqs) {
```
elshize: Sorry, another comment: I think for testing this is not the best approach: I think we want to extract `SampleFn` that only samples indices, basically only the random stuff. Because later on, one of the very important things to test is that documents and frequencies are properly aligned. You can't test this right now.

external/CMakeLists.txt
amallia: ParallelSTL generates some warnings. I am only limiting this to ParallelSTL's code. 
Ideally we should submit a PR to ParallelSTL's repository. I have done that many times and they take forever to accept a PR. So I prefer to ignore for now

src/sample_inverted_index.cpp
amallia: Sorry! I fixed it, but never pushed it.

external/CMakeLists.txt
elshize: Ok.

src/sample_inverted_index.cpp
amallia: It might. Let me try :)

src/sample_inverted_index.cpp
amallia: I will make it random accessible

src/sample_inverted_index.cpp
elshize: I know that it works funky when you want to later, say, capture one of them in a lambda (which is fixed in C++20 btw), but I think it should work here.

src/sample_inverted_index.cpp
elshize: Awesome :+1:

include/pisa/util/index_build_utils.hpp
```diff
@@ -161,4 +164,44 @@ void sample_inverted_index(const std::string &input_basename,
...
+    std::ofstream dos(output_basename + ".docs");
+    std::ofstream fos(output_basename + ".freqs");
+
+    auto document_count = static_cast<uint32_t>(input.num_docs());
+    write_sequence(dos, gsl::make_span<uint32_t const>(&document_count, 1));
+
+    {
+        pisa::progress progress("Sampling inverted index", input.size());
+        for (auto const &plist : input) {
+            auto sample = sample_fn(plist.docs.size());
```
elshize: I think in any reasonable case, we'll want the sampling to be stable, that is, we don't want to reorder the postings we keep. But here, we're opening ourselves to a situation where it's not ensured. I think it is reasonable to expect a stable sampling function but not to assume one is given. I suggest adding a comment at this function's signature that `sample_fn` must be stable, or that it must return sorted vectors, and here run `std::is_sorted` and fail if it's `false`. I think it's a reasonable trade-off (samples should be quite small) for the safety we get.

We could also just sort here if we want to be more flexible, but it's kind of wasteful, and I think it's reasonable to expect a sorted list of indices (the sampling method can sort itself if it knows it's not stable).

include/pisa/util/index_build_utils.hpp
```diff
@@ -161,4 +164,44 @@ void sample_inverted_index(const std::string &input_basename,
...
+                                 fmt::format("{}.sizes", output_basename),
+                                 boost::filesystem::copy_option::overwrite_if_exists);
+
+    std::ofstream dos(output_basename + ".docs");
+    std::ofstream fos(output_basename + ".freqs");
+
+    auto document_count = static_cast<uint32_t>(input.num_docs());
+    write_sequence(dos, gsl::make_span<uint32_t const>(&document_count, 1));
+
+    {
```
elshize: This block is unnecessary. We're exiting the scope right after anyway.

include/pisa/util/index_build_utils.hpp
```diff
@@ -161,4 +164,44 @@ void sample_inverted_index(const std::string &input_basename,
...
+                sampled_freqs.push_back(plist.freqs[index]);
+            }
+
+            write_sequence(dos, gsl::span<uint32_t const>(sampled_docs));
+            write_sequence(fos, gsl::span<uint32_t const>(sampled_freqs));
+            progress.update(1);
+        }
+    }
+    dos.close();
+    fos.close();
```
elshize: We're leaving the scope, so they will be closed anyway, no need to do it by hand, we can remove these.

elshize (changes requested):

include/pisa/util/index_build_utils.hpp
```diff
@@ -161,4 +164,44 @@ void sample_inverted_index(const std::string &input_basename,
     }
 }
 
+
+template <typename SampleFn>
+void sample_inverted_index(std::string const &input_basename,
+                  std::string const &output_basename,
+                  SampleFn &&sample_fn)
```
elshize: Oh, and `clang-format` while you're at it ;)

elshize (approved):

Issue #246, opened 2019-09-09
Add information about termmap and docmap in docs


elshize:
It's missing from: https://pisa.readthedocs.io/en/latest/parsing.html

PR #268, opened 2019-10-15
Update parsing.md

JMMackenzie:
For #246 

- Add some documentation about the `lexicon` binary.

### DIFF
--- a/docs/source/parsing.md
+++ b/docs/source/parsing.md
@@ -43,15 +43,20 @@ The parsing process will write the following files:
 - `cw09b`: forward index in binary format.
 - `cw09b.terms`: a new-line-delimited list of sorted terms,
   where term having ID N is on line N, with N starting from 0.
+- `cw09b.termlex`: a binary representation of the `.terms` file that is used to look up term identifiers at query time.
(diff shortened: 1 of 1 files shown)


docs/source/parsing.md
```diff
@@ -49,6 +49,34 @@ The parsing process will write the following files:
   line N, with N starting from 0. Also, keep in mind that each ID corresponds with
   an ID of the `cw09b.documents` file.
 
+### Generating lexicon and document map
+Once the index has been parsed, you can generate the document map and lexicon which will be used at query time.
```
elshize: Term and document lexicons are actually built at the end of `parse_collection`. So we might want to point that out and maybe emphasize that this command can be used to read or make lookups. I think it still make sense to explain the lexicon generation but maybe phrase it such that it's clear it can be done for any new-line delimited text file. Also maybe we should say that reverse lookup (string to ID) will work correctly only if the tokens are sorted.

elshize: The reason for the last one is that we perform a binary search to find it.

JMMackenzie: OK, have pushed some updates. Could you please take another look?

Issue #263, opened 2019-10-09
Bugs found with Clang static analyzer


elshize:
Logic error | Dereference of null pointer | include/pisa/bit_vector.hpp | skip | 426 | 29 | View Report | Report Bug | Open File
-- | -- | -- | -- | -- | -- | -- | -- | --
Logic error | Dereference of null pointer | include/pisa/bit_vector.hpp | next | 410 | 22 | View Report | Report Bug | Open File
Logic error | Division by zero | include/pisa/codec/compact_elias_fano.hpp | offsets | 29 | 18 | View Report | Report Bug | Open File
Logic error | Result of operation is garbage or undefined | include/pisa/codec/integer_codes.hpp | read_gamma | 26 | 22 | View Report | Report Bug | Open File
Logic error | Result of operation is garbage or undefined | include/pisa/bit_vector.hpp | get_bits | 239 | 25 | View Report | Report Bug | Open File
Logic error | Result of operation is garbage or undefined | include/pisa/codec/integer_codes.hpp | read_gamma | 26 | 24 | View Report | Report Bug | Open File

Details are in the program output: [report.zip](https://github.com/pisa-engine/pisa/files/3704980/report.zip)

We can ignore the ones from FastPFor; it's not a big deal and I've already created a PR in the lib repo: https://github.com/lemire/FastPFor/pull/65

elshize:
Note that the above results are incomplete, we should re-run `scan-build` tool to get the full list.

elshize:
No longer seeing any errors in our code.

PR #272, opened 2019-10-20
Fix TBB leaks in tests

### LINKED ISSUE #265: Use task_scheduler_init in tests where TBB is used
Currently, tests that use TBB to, e.g., build a test index, fail with memory leaks when run with sanitizers.

The fix is to always include

```cpp
tbb::task_scheduler_init init(1);
```
at the beginning of each test (number of threads doesn't necessarily need to be 1).

elshize:
Fixes #265

There are still other problems that the sanitizer finds in the following tests:
```
The following tests FAILED:
	  5 - test_block_codecs:block_codecs (Failed)
	  9 - test_bmw_queries:block_max_wand (Failed)
	 25 - test_freq_index:freq_index (Failed)
	 43 - test_partition_fwd_index:copy_sequence (Failed)
	 44 - test_partition_fwd_index:Rearrange sequences (Failed)
	 46 - test_partitioned_sequence:partitioned_sequence (Failed)
	 54 - test_positive_sequence:positive_sequence (Failed)
```

Likely some of them (or all) are related to the problems clang static analyzer found: #263

### DIFF
--- a/test/test_forward_index_builder.cpp
+++ b/test/test_forward_index_builder.cpp
@@ -167,7 +167,7 @@ void write_batch(std::string const &basename,
 
 TEST_CASE("Merge forward index batches", "[parsing][forward_index]")
 {
-    tbb::task_scheduler_init init(2);
+    tbb::task_scheduler_init init;
     Temporary_Directory tmpdir;
     auto dir = tmpdir.path();
     GIVEN("Three batches on disk")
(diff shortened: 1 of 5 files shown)


test/test_forward_index_builder.cpp
```diff
@@ -157,6 +167,7 @@ void write_batch(std::string const &basename,
 
 TEST_CASE("Merge forward index batches", "[parsing][forward_index]")
 {
+    tbb::task_scheduler_init init(2);
```
JMMackenzie: I don't think we even need to pass a number here, though it probably doesn't matter.

test/test_forward_index_builder.cpp
```diff
@@ -298,6 +314,7 @@ TEST_CASE("Parse HTML content", "[parsing][forward_index][unit]")
 
 TEST_CASE("Build forward index", "[parsing][forward_index][integration]")
 {
+    tbb::task_scheduler_init init(2);
```
JMMackenzie: Same as above.

JMMackenzie (approved):

elshize:
Thanks @JMMackenzie good point on the no. threads, I removed them.

PR #273, opened 2019-10-21
Fixing static analyzer errors

### LINKED ISSUE #12: test_bit_vector fails nondeterministically
At the very least, we should:

- Draw a random seed
- Print the seed out
- Use the seed for generating bit vectors
- Debug using the seed that fails

Also should consider running `ctest -VV` on Travis to be able to capture the output from there as well.

elshize:
Fixes #12

### DIFF
--- a/test/test_bit_vector.cpp
+++ b/test/test_bit_vector.cpp
@@ -1,190 +1,199 @@
 #define CATCH_CONFIG_MAIN
 #include "catch2/catch.hpp"
 
+#include <cstdlib>
+
+#include <rapidcheck.h>
+
+#include "bit_vector.hpp"
+#include "mappable/mapper.hpp"
 #include "test_common.hpp"
 #include "test_rank_select_common.hpp"
 
-#include <cstdlib>
-#include "mappable/mapper.hpp"
-#include "bit_vector.hpp"
-
 TEST_CASE("bit_vector")
 {
-    srand(42);
+    rc::check([](std::vector<bool> v) {
+        {
+            pisa::bit_vector_builder bvb;
+            for (size_t i = 0; i < v.size(); ++i) {
+                bvb.push_back(v[i]);
+            }
 
-    std::vector<bool> v = random_bit_vector();
-
-    {
-        pisa::bit_vector_builder bvb;
-        for (size_t i = 0; i < v.size(); ++i) {
-            bvb.push_back(v[i]);
+            pisa::bit_vector bitmap(&bvb);
+            test_equal_bits(v, bitmap, "Random bits (push_back)");
         }
 
-        pisa::bit_vector bitmap(&bvb);
-        test_equal_bits(v, bitmap, "Random bits (push_back)");
-    }
+        {
+            pisa::bit_vector_builder bvb(v.size());
+            for (size_t i = 0; i < v.size(); ++i) {
+                bvb.set(i, v[i]);
+            }
+            bvb.push_back(0);
+            v.push_back(0);
+            bvb.push_back(1);
+            v.push_back(1);
 
-    {
-        pisa::bit_vector_builder bvb(v.size());
-        for (size_t i = 0; i < v.size(); ++i) {
-            bvb.set(i, v[i]);
-        }
-        bvb.push_back(0);
-        v.push_back(0);
-        bvb.push_back(1);
-        v.push_back(1);
-
-        pisa::bit_vector bitmap(&bvb);
-        test_equal_bits(v, bitmap, "Random bits (set)");
-    }
-
-    uint64_t ints[] = {uint64_t(-1), uint64_t(1) << 63, 1, 1, 1, 3, 5, 7, 0xFFF, 0xF0F, 1, 0xFFFFFF, 0x123456, uint64_t(1) << 63, uint64_t(-1)};
-    {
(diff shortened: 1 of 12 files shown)


test/test_bit_vector.cpp
```diff
@@ -1,190 +1,199 @@
...
 {
-    srand(42);
-    uint64_t n = 20000;
-    std::vector<bool> v = random_bit_vector(n);
-
-    // punch some long gaps in v
-    for (size_t g = 0; g < n / 1000; ++g) {
-        ssize_t l = std::min(ssize_t(rand() % 256), ssize_t(v.size() - g));
-        std::fill(v.begin(), v.begin() + l, 0);
-    }
```
elshize, line 112: @amallia Do you have any idea what actually was the point of this? It's weird, because it always fills from the beginning of the vector to some `l`, so effectively it will just make one "gap" at the beginning, won't it?

test/test_bit_vector.cpp
```diff
@@ -1,190 +1,199 @@
...
+                for (size_t d = 0; d < std::min(size_t(256), size_t(v.size() - pos)); ++d) {
+                    if (v[pos + d] == 0) {
+                        pisa::bit_vector::unary_enumerator ee(bitmap, pos);
+                        ee.skip0(skip);
+
+                        uint64_t expected_pos = pos + d;
+                        for (; !v[expected_pos] && expected_pos < v.size(); ++expected_pos)
+                            ;
+                        if (!v[expected_pos])
+                            break;
```
elshize: @amallia Is this supposed to break when no `true` is found until the end of the vector? `expected_pos` will be equal to `v.size()` in such case, and this will be out of bounds. Also, in the `for` loop above, the order in the `&&` should be reversed because, again, we will access something out of bounds and then checking the bound after.

Am I understanding what this code was supposed to do correctly?

test/test_bit_vector.cpp
amallia: It looks like a bug actually. Nice catch!

test/test_bit_vector.cpp
amallia: I think you are right! Wrong order and second test can be out of bounds

test/test_bit_vector.cpp
elshize: So then, the intention is to introduce some contiguous spans of zeroes in the vector, is that correct?

elshize:
Please note that after running the tests on more randomized examples, it doesn't pass. In particular, it's `bit_vector_unary_enumerator` that fails (see Travis).

I'll make the fixes to the test and see if it still happens.

test/test_bit_vector.cpp
```diff
@@ -1,190 +1,199 @@
...
-                    for (; !v[expected_pos] && expected_pos < v.size(); ++expected_pos);
-                    if (!v[expected_pos]) break;
+            for (size_t r = 0; r < ones.size(); ++r) {
+                for (size_t k = 0; k < std::min(size_t(256), size_t(ones.size() - r)); ++k) {
+                    pisa::bit_vector::unary_enumerator ee(e);
+                    uint64_t pos_skip = ee.skip_no_move(k);
                     uint64_t pos = ee.next();
-                    MY_REQUIRE_EQUAL(expected_pos, pos,
-                                     "pos = " << pos << " skip = " << skip);
+                    MY_REQUIRE_EQUAL(ones[r], pos, "r = " << r << " k = " << k);
```
elshize: @amallia You know you can use `CAPTURE` macro to print a variable if the test fails, like so:
```cpp
CAPTURE(r);
CAPTURE(k);
```
Or did you choose this over it because of something I'm not seeing?

elshize:
There are still two buffer overflows (seem to be reads right after allocated memory):
1. `test_bmw_queries` in https://github.com/pisa-engine/pisa/blob/master/include/pisa/score_opt_partition.hpp#L90
2. `test_block_codecs` in https://github.com/pisa-engine/pisa/blob/master/include/pisa/codec/interpolative_coding.hpp#L98

Also, there is still one error from the static analyzer. The problem is that `bit_vector::unary_enumerator` has a default constructor that set `uint64_t const *m_data;` to `nullptr`. It is then possible to call any of the member functions that read data from it.

This is not the best design, since we are opening ourselves to some undefined behavior. The easiest fix is to remove the default constructor, but then we run into `compact_elias_fano`, which contains it, and not only does it have its own default constructor, but also the other constructor fails to initialize this field, and therefore tries calling the default constructor of `bit_vector::unary_enumerator` (which we wanted to remove). And then they are other structures that do something very similar. Unfortunately, I'm not very familiar with this piece of code, and it's not clear to me if we can initialize this enumerator at object construction for, say, `compact_elias_fano`, and if so, what value should that be?

In any case, I see two different acceptable approaches here:
1. We either always initialize all of these things at construction (no default constructors), or
2. We check at runtime if we have a valid object.

Otherwise, things can become unpleasant. Clearly, (1) is preferable if possible. And I guess it should be, but maybe we need to redesign a bit? @amallia can we talk about this tomorrow?

elshize:
@amallia let's merge this and open issue for outstanding stuff? Review when get a chance.

amallia:
I agree

PR #274, opened 2019-10-22
Added terms len into wand file

amallia:
As discussed with @elshize


elshize:
Looks good to me. I'm only thinking how we should approach these breaking changes. We're still not stable, so that's okay in general, but since we will likely introduce more format changes in 1.0, maybe at some point we should create 1.0 branch where we add these things gradually. What do you think @amallia @JMMackenzie ?

elshize (approved):

JMMackenzie:
I agree. Also, why are we doing this? What was wrong with doing `list.size()` here with the underlying postings list?

elshize:
@JMMackenzie We need this for sampling --- in order to retrieve the original scores, we need to have access to its real length. So this is used for scoring.

PR #280, opened 2019-10-28
v1

elshize:
I've been working on starting a draft of index format specification and some code examples.

Let's discuss!

- (very early stage of) draft and lib code in [`include/pisa/v1`](https://github.com/pisa-engine/pisa/tree/v1/include/pisa/v1)
- examples in [`test/test_v1.cpp`](https://github.com/pisa-engine/pisa/blob/v1/test/test_v1.cpp)


elshize:
As of the latest commit, we can compress an index (`compress` program) and run queries both for evaluation and benchmarks with `query` program. It's not yet parameterized by algorithm, so only ranked OR runs but it actually performs very well, compared to the old ranked OR. Below results for blocked posting lists with SIMDBP encoding on Robust (on my laptop).

New:
```
[2019-11-11 21:01:47.675] [info] Mean: 1568.39
[2019-11-11 21:01:47.675] [info] 50% quantile: 1013
[2019-11-11 21:01:47.675] [info] 90% quantile: 3199
[2019-11-11 21:01:47.675] [info] 95% quantile: 6066
```

Old:
```
[2019-11-11 21:01:53.201] [stderr] [info] Mean: 1751.2
[2019-11-11 21:01:53.201] [stderr] [info] 50% quantile: 1079
[2019-11-11 21:01:53.202] [stderr] [info] 90% quantile: 3483
[2019-11-11 21:01:53.202] [stderr] [info] 95% quantile: 6693
```

The more interesting algorithms aren't implemented yet, but stay tuned.

Also, the compression is concurrent and quite fast, Robust compresses in seconds.

elshize:
A quick update: I implemented precomputed scores (only as `float`s right now, without quantization) and store them as 4-bytes each. It takes a lot of space, but the idea is to maybe quantize to one byte each score, and then have a very fast score lookup. I'm expecting it to be at least as fast. I'll also experiment later with different codecs once I have integers instead of floats.

```
[2019-11-11 21:47:38.576] [info] Mean: 793.68
[2019-11-11 21:47:38.576] [info] 50% quantile: 418
[2019-11-11 21:47:38.576] [info] 90% quantile: 1948
[2019-11-11 21:47:38.576] [info] 95% quantile: 4284
```

include/pisa/v1/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+> Example: `v1/posting_format_header.hpp`.
+
+We should store the type of the postings in the file, as well as encoding used.
+**This might be tricky because we'd like it to be an open set of values/encodings.**
+
+```
+Header := Version, Type, Encoding
+Version := Major, Minor, Path
+Type := ValueId, Count
```
amallia, line 56: I am a bit confused by `Type`. What are `ValueId` and `Count`?

include/pisa/v1/README.md
```diff
@@ -0,0 +1,79 @@
...
+Each _posting file_ contains a list of blocks of data, each related to a single term,
+preceded by a header encoding information about the type of payload.
+
+> Do we need the header? I would say "yes" because even if we store the information
+> somewhere else, then we might want to (1) verify that we are reading what we think
+> we are reading, and (2) verify format version compatibility.
+> The latter should be further discussed.
+
+```
+Posting File := Header, [Posting Block]
```
amallia, line 37: Is there one `Header` for each `Posting Block`?

include/pisa/v1/README.md
elshize: `ValueId` would be the type, and count would be how many of those, as in one, or pair or tuple, etc.

Actually, so far I implemented it like this: https://github.com/pisa-engine/pisa/pull/280/files#diff-2a007c99bc1af07f1fb150c293383559R71

Another approach would be to always have scalars in one file, and join multiple ones for tuples. But then we can't store arrays of undetermined length (say, positional index). But all of this is up for discussion.

include/pisa/v1/README.md
elshize: No, one header, followed by a list of blocks.

elshize:
Quick tests on Clueweb09B show essentially the same results for BM25 ranked OR as before, while the average drops from `392.932` to `267.576` with precomputed quantized scores of length 1-byte without further compression.

PR #283, opened 2019-11-10
Create generate_mapping_from_permutations.py

amallia:

Takes two files containing a list of documents and generates a file with
mappings between position in the first file to the position in the second
according to the standard format '<ID> <new ID>' to use with the
'suffle_docids' script.

Example:
```
file1.txt

a
b
c
```

```
file2.txt

c
b
a
d
```

```
mapping.txt

0 2
1 1
2 0
```

@elshize @JMMackenzie @gustingonzalez


elshize (approved):
Looks good to me.

PR #288, opened 2019-11-25
Pass top-k queue to the query algorithm

amallia:


amallia:
@elshize while I fix this, could you give it a look in the meantime?

include/pisa/query/algorithm/range_taat_query.hpp
```diff
@@ -0,0 +1,48 @@
...
+            return 0;
+        }
+
+        accumulator.init();
+
+        for (size_t end = range_size;
+             end + range_size <= max_docid; end += range_size) {
+            process_range(cursors, end, accumulator);
+        }
+        process_range(cursors, max_docid, accumulator);
```
elshize: I think we'll end up processing twice the last range if the total length is a multiplier of the range size. I think the `for` condition should be instead `end + range_size < max_docid`, and that should fix it.

elshize (changes requested):
Looks good other than the one comment.

elshize:
Looks good, what's the problem with Travis, is it timing out?

PR #289, opened 2019-11-26
Sampling

amallia:


include/pisa/wand_data.hpp
```diff
@@ -48,21 +50,37 @@ class wand_data {
...
 
-        for (auto const &seq : coll) {
-            size_t term_occurrence_count = std::accumulate(seq.freqs.begin(), seq.freqs.end(), 0);
-            term_occurrence_counts.push_back(term_occurrence_count);
-            term_posting_counts.push_back(seq.docs.size());
+        {
+           pisa::progress progress("Storing terms statistics", coll.size());
+           size_t term_id = 0;
+           for (auto const &seq : coll) {
+               if(dropped_term_ids.find(term_id) != dropped_term_ids.end()){
```
elshize: Side note: did you know that c++20 finally added `contains` method to sets? So progressive.

elshize (approved):
Looks good to me!

include/pisa/util/inverted_index_utils.hpp
```diff
@@ -12,12 +12,12 @@ void emit(std::ostream &os, const uint32_t *vals, size_t n)
 
 void emit(std::ostream &os, uint32_t val) { emit(os, &val, 1); }
 
-
 // sample_fn must be stable, it must return a sorted vector
 template <typename SampleFn>
 void sample_inverted_index(std::string const &input_basename,
                            std::string const &output_basename,
-                           SampleFn &&sample_fn)
+                           SampleFn &&sample_fn,
```
gustingonzalez, line 20: Why not use a `std::function` instead a template?

include/pisa/util/inverted_index_utils.hpp
```diff
@@ -32,9 +32,16 @@ void sample_inverted_index(std::string const &input_basename,
     auto document_count = static_cast<uint32_t>(input.num_docs());
     write_sequence(dos, gsl::make_span<uint32_t const>(&document_count, 1));
     pisa::progress progress("Sampling inverted index", input.size());
+    size_t term = 0;
     for (auto const &plist : input) {
-        auto sample = sample_fn(plist.docs.size());
-        assert(std::is_sorted(std::begin(sample),std::end(sample)));
+        auto sample = sample_fn(plist.docs);
+        if (sample.size() == 0) {
+            dropped_term_ids.push_back(term);
```
gustingonzalez: `dropped_term_ids.push_back(term++)?`

src/create_wand_data.cpp
```diff
@@ -41,6 +43,7 @@ int main(int argc, const char **argv)
     app.add_flag("--compress", compress, "Compress additional data");
     app.add_option("-s,--scorer", scorer_name, "Scorer function")->required();
     app.add_flag("--range", range, "Create docid-range based data")->excludes(var_block_opt);
+    app.add_option("--dropped_terms", dropped_terms_filename, "A filename containing a list of term IDs that we want to drop");
```
gustingonzalez: Why not `--terms_to_drop`? In same way, the description maybe could be improved as:` "A filename containing a list of term IDs to drop"`.

src/sample_inverted_index.cpp
```diff
@@ -24,28 +26,71 @@ int main(int argc, char **argv)
     app.add_option("-o,--output", output_basename, "Output collection basename")->required();
     app.add_option("-r,--rate", rate, "Sampling rate (proportional size of the output index)")
         ->required();
+    app.add_option("-t,--type", type, "Sampling type")->required();
+    app.add_option("--dropped_terms",
```
gustingonzalez: Same as above. Maybe `--terms_to_drop` sounds better.

include/pisa/util/inverted_index_utils.hpp
elshize: I'm good with either, we're using `std::function` anyway, so might as well put it in the signature.

include/pisa/util/inverted_index_utils.hpp
elshize: Actually, at this rate, why don't we just say:
```cpp
for (auto term = 0; term < input.size(); term += 1, progress.update(1)) {
    auto& plist = input[term];
    ...
}
```
Less error-prone than keeping track of the counters in two places.

src/create_wand_data.cpp
elshize: I actually agree, "terms to drop" is way clearer. But I would personally go with `--terms-to-drop` with dashes instead of underscores.

src/create_wand_data.cpp
```diff
@@ -41,6 +43,7 @@ int main(int argc, const char **argv)
     app.add_flag("--compress", compress, "Compress additional data");
     app.add_option("-s,--scorer", scorer_name, "Scorer function")->required();
     app.add_flag("--range", range, "Create docid-range based data")->excludes(var_block_opt);
```
gustingonzalez: Maybe we should define a criteria regarding the use of the `document id` and the `term id `concepts, specially in the usage info. For example, `docid/doc ID/document identifier/etc` and `termid/term ID/term identifier/etc` are used respectively.

gustingonzalez: Maybe we could define a criteria regarding the use of the document id and the term id concepts, specially in the usage info. For example, docid/doc ID/document identifier/etc and termid/term ID/term identifier/etc are used respectively.

gustingonzalez, line 45: Maybe we could define a criteria regarding the use of the `document id` and the `term id` concepts, specially in the usage info. For example, `docid/doc ID/document identifier/etc` and `termid/term ID/term identifier/etc` are used respectively.

src/sample_inverted_index.cpp
```diff
@@ -24,28 +26,71 @@ int main(int argc, char **argv)
     app.add_option("-o,--output", output_basename, "Output collection basename")->required();
     app.add_option("-r,--rate", rate, "Sampling rate (proportional size of the output index)")
         ->required();
+    app.add_option("-t,--type", type, "Sampling type")->required();
```
gustingonzalez, line 30: Maybe the usage of the `type` option could show the supported samplings (`random_postings` and `random_docids`), explaining them briefly.

include/pisa/util/inverted_index_utils.hpp
```diff
@@ -12,12 +12,12 @@ void emit(std::ostream &os, const uint32_t *vals, size_t n)
...
 void emit(std::ostream &os, uint32_t val) { emit(os, &val, 1); }
 
-
 // sample_fn must be stable, it must return a sorted vector
 template <typename SampleFn>
 void sample_inverted_index(std::string const &input_basename,
                            std::string const &output_basename,
-                           SampleFn &&sample_fn)
+                           SampleFn &&sample_fn,
+                           std::vector<size_t> &dropped_term_ids)
```
JMMackenzie: Any reason we use a `vector` here and an `unordered_set` later? I realize we are pushing terms in here, and we are finding them in the later case, so I am guessing that is why. I suppose this could be an `unordered_set` too though?

Probably does not matter :smile:

JMMackenzie (approved):
Looks good (+ the comments the others left already), great work!

src/create_wand_data.cpp
amallia: Yes, we could do this in a separate PR

include/pisa/util/inverted_index_utils.hpp
amallia: We cannot do the above @elshize, there is no `operator[]` for `binary_freq_collection` for now. We need to add it, but I dont want to do it in this PR as it might break other things.

include/pisa/util/inverted_index_utils.hpp
amallia: Yes, you are right. Lets go with `unordered_set` everywhere

PR #296, opened 2019-12-22
AOL query extractor

amallia:


JMMackenzie (approved):
It looks good to me. 

One thing which is not required 100% (but I think would be nice) is to document the input file formats. Someone unfamiliar with the AOL logs might not know the input format.

PR #302, opened 2020-01-01
Reverse url and update doclex order 

amallia:
Happy new year!

Just a script to reverse the URLs contained in the `.url` file.

Also, a small update to `shuffle_docids` to update the `.doclex` file

### DIFF
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
+from urllib.parse import urlsplit
+import re
+
+def reverse_url(url):
+    components = urlsplit(url)
+    hostname = components.netloc.partition(':')[0]
+    result = '.'.join(hostname.split('.')[::-1])
+    if components.path:
+        result += components.path
+    if components.query:
+        result += '.' + components.query
+    if components.fragment:
+        result += components.fragment
+
+    return re.sub(r'\W', '.', result).lower()
+
+
+def main():
+    parser = ArgumentParser(description="Take a text file of URLs (one per line) \
+                             and reverse each URL to use it to reorder the index.")
+    parser.add_argument("input", help="Input file containing one URL per line")
+    parser.add_argument("output", help="Output file containing one reversed URL per line")
+    args = parser.parse_args()
+
+    # Documents dict, where each element maps an id with a document (or url).
+    with open(args.input, "r") as in_file, open(args.output, mode='w') as out_file:
+        for line in in_file:
+            out_file.write(reverse_url(line)+"\n")
+
+if __name__ == "__main__":
+    main()
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -5,7 +5,7 @@ from urllib.parse import urlsplit
 import re
 
 """
->>> reverse_url("http://some.url/to?reverse=true")
+>>> reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
 gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
 """
 def reverse_url(url):
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -4,11 +4,11 @@ from argparse import ArgumentParser
 from urllib.parse import urlsplit
 import re
 
-"""
->>> reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
-gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
-"""
 def reverse_url(url):
+    """
+    >>> reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
+    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
+    """
     components = urlsplit(url)
     hostname = components.netloc.partition(':')[0]
     result = '.'.join(hostname.split('.')[::-1])
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -7,7 +7,7 @@ import re
 def reverse_url(url):
     """
     >>> reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
-    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm
+    'gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm'
     """
     components = urlsplit(url)
     hostname = components.netloc.partition(':')[0]
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -7,7 +7,7 @@ import re
 def reverse_url(url):
     """
     >>> reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
-    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
+    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm
     """
     components = urlsplit(url)
     hostname = components.netloc.partition(':')[0]
--- a/script/reverse_urls.py
+++ b/script/reverse_urls.py
@@ -32,7 +32,7 @@ def main():
     # Documents dict, where each element maps an id with a document (or url).
     with open(args.input, "r") as in_file, open(args.output, mode='w') as out_file:
         for line in in_file:
-            out_file.write(reverse_url(line)+"\n")
+            out_file.write(reverse_url(line[:-1])+"\n")
 
 if __name__ == "__main__":
     main()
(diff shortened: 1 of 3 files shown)


script/reverse_urls.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
+from urllib.parse import urlsplit
+import re
+
+def reverse_url(url):
```
elshize, line 7: Can we add a doctest showcasing what this does? Something like:
```
"""
>>> http://some.url/to?reverse=true
RESULT
"""
```
It's not clear right away what it means and the body is not trivial, so it's good to have a doc+test for clarity.

src/shuffle_docids.cpp
```diff
@@ -1,104 +1,120 @@
...
+  CLI::App app{"Shuffle document IDs."};
+  app.add_option("-c,--collection", input_basename, "Collection basename")
+      ->required();
+  app.add_option("-o,--output", output_basename, "Output basename");
+  app.add_option("--mapping-filename", mapping_filename, "Ordering file is of the form <current ID> <new ID>");
+  auto docs_opt = app.add_option("--documents", documents_filename, "Documents lexicon");
+  app.add_option("--reordered-documents", reordered_documents_filename, "Reordered documents lexicon")->needs(docs_opt);
+  CLI11_PARSE(app, argc, argv);
+
+  constexpr uint32_t seed = 1729;
```
elshize: Should we add an option to override it from command line?

elshize (changes requested):

src/shuffle_docids.cpp
amallia: Can you check now?

script/reverse_urls.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
+from urllib.parse import urlsplit
+import re
+
+"""
+>>> reverse_url("http://some.url/to?reverse=true")
+gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
```
elshize: In python, comments to functions go underneath the declaration line, so
```python
def reverese_url(url):
    """
    >>> ...
    """
    <code>
```
Now that you can run it, it fails:
```
$ python -m doctest script/reverse_urls.py
**********************************************************************
File "script/reverse_urls.py", line 9, in reverse_urls.reverse_url
Failed example:
    reverse_url("http://some.url/to?reverse=true")
Expected:
    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
Got:
    'url.some.to.reverse.true'
**********************************************************************
1 items had failures:
   1 of   1 in reverse_urls.reverse_url
***Test Failed*** 1 failures.
```
I think you forgot to update the argument.

script/reverse_urls.py
```diff
@@ -0,0 +1,34 @@
...
+    parser = ArgumentParser(description="Take a text file of URLs (one per line) \
+                             and reverse each URL to use it to reorder the index.")
+    parser.add_argument("input", help="Input file containing one URL per line")
+    parser.add_argument("output", help="Output file containing one reversed URL per line")
+    args = parser.parse_args()
+
+    # Documents dict, where each element maps an id with a document (or url).
+    with open(args.input, "r") as in_file, open(args.output, mode='w') as out_file:
+        for line in in_file:
+            out_file.write(reverse_url(line)+"\n")
```
gustingonzalez: Here is missing the -1 to avoid the '\n' in the line. Otherwise, an url as "https://www.google.com" is converted to "com..google.www".

By the way the https and http is missing in the final result too. Does it matter?

gustingonzalez: Here is missing the -1 to avoid the '\n' in the line. Otherwise, an url as 'https://www.google.com' is converted to 'com..google.www'.

By the way, the 'https' and 'http' are also missing in the returned result. Does it matter?

amallia: >By the way, the 'https' and 'http' are also missing in the returned result. Does it matter?

We do not want `http,https` here

>Here is missing the -1 to avoid the '\n' in the line. Otherwise, an url as 'https://www.google.com' is converted to 'com..google.www'.

Where is `-1` missing?

script/reverse_urls.py
amallia: Added

script/reverse_urls.py
elshize: `for line in in_file:` will produce strings with `\n` at the end, so what you usually do is `out_file.write(reverse_url(line[:-1])+"\n")`

elshize (changes requested):

script/reverse_urls.py
```diff
@@ -0,0 +1,34 @@
...
+    if components.query:
+        result += '.' + components.query
+    if components.fragment:
+        result += components.fragment
+
+    return re.sub(r'\W', '.', result).lower()
+
+
+def main():
+    parser = ArgumentParser(description="Take a text file of URLs (one per line) \
```
gustingonzalez, line 26: Why not unify this script with [this other](https://github.com/pisa-engine/pisa/blob/master/script/generate_sorted_docids_mapping.py)? Maybe could be added an option like `--reverse-url`.

script/reverse_urls.py
amallia: `reverse_url` is already taking care of that actually

script/reverse_urls.py
```diff
@@ -0,0 +1,34 @@
...
+from argparse import ArgumentParser
+from urllib.parse import urlsplit
+import re
+
+def reverse_url(url):
+    components = urlsplit(url)
+    hostname = components.netloc.partition(':')[0]
+    result = '.'.join(hostname.split('.')[::-1])
+    if components.path:
+        result += components.path
```
gustingonzalez, line 16: This can be simplified as: `result += components.path or ''`

script/reverse_urls.py
amallia: These are scripts that perform different tasks, I prefer to keep them in separate files

gustingonzalez:
> Happy new year!
> 
> Just a script to reverse the URLs contained in the `.url` file.
> 
> Also, a small update to `shuffle_docids` to update the `.doclex` file

Happy new year 🎉!

Left some comments.

script/reverse_urls.py
amallia: True, but since it is a cascade of multiple `if` statements, where the others cannot be changed, I prefer to keep them consistent.

script/reverse_urls.py
amallia: Fixed. Sorry I was editing from Github :)

script/reverse_urls.py
amallia: I take it back. You are right, fixed!

gustingonzalez: But the [\n](url) becomes a `.`, adding an 'extra dot' in the url. In a url with params, there is no problem (the dot is added at the end), but if the url has no parameters, the result is something like `com..some.www

gustingonzalez: But the \n becomes a ., adding an 'extra dot' in the url. In a url with params, there is no problem (the dot is added at the end), but if the url has no parameters, the result is something like `com..some.www`

elshize (changes requested):
The test is still failing, missing the quotes and have a superfluous period. For more about python doctests: https://docs.python.org/3/library/doctest.html
```python
# python -m doctest script/reverse_urls.py
**********************************************************************
File "script/reverse_urls.py", line 9, in reverse_urls.reverse_url
Failed example:
    reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
Expected:
    gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
Got:
    'gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm'
**********************************************************************
1 items had failures:
   1 of   1 in reverse_urls.reverse_url
***Test Failed*** 1 failures.
```

amallia:
> The test is still failing, missing the quotes and have a superfluous period. For more about python doctests: https://docs.python.org/3/library/doctest.html
> 
> ```python
> # python -m doctest script/reverse_urls.py
> **********************************************************************
> File "script/reverse_urls.py", line 9, in reverse_urls.reverse_url
> Failed example:
>     reverse_url("http://www.face.bnl.gov/FACEDailyReports/Facts_I/Archive/03132003/ArchivePage.htm")
> Expected:
>     gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm.
> Got:
>     'gov.bnl.face.www.facedailyreports.facts_i.archive.03132003.archivepage.htm'
> **********************************************************************
> 1 items had failures:
>    1 of   1 in reverse_urls.reverse_url
> ***Test Failed*** 1 failures.
> ```

You are right. There was an extra dot

elshize:
> You are right. There was an extra dot

Yes, plus you need to put the quotes around. Can you just run the test before you push? `python -m doctest script/reverse_urls.py` if nothing prints out, it passes.

amallia:
> > You are right. There was an extra dot
> 
> Yes, plus you need to put the quotes around. Can you just run the test before you push? `python -m doctest script/reverse_urls.py` if nothing prints out, it passes.

Done! Sorry I was editing from Github as I was working on a separate branch. That was probably annoying, my bad!

elshize (approved):

gustingonzalez (approved):

JMMackenzie (approved):
Late to the party (blame Australia's timezone :-)

Looks good to me.

amallia:
Thank you guys! I appreciate your help

PR #303, opened 2020-01-01
Add index slicing

gustingonzalez:
Add a (no random) sampling option to take an index portion from its beginning.

### DIFF
--- a/include/pisa/util/inverted_index_utils.hpp
+++ b/include/pisa/util/inverted_index_utils.hpp
@@ -1,5 +1,4 @@
 #pragma once
-#include <unordered_set>
 #include "binary_freq_collection.hpp"
 #include "invert.hpp"
 #include "util/progress.hpp"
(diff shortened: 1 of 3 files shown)


amallia:
What is the use case?

gustingonzalez:
> What is the use case?

Mainly to avoid the entire index load in some tests. Particularly, refactoring and playing with the chunk size of the posting list, I was obtained a decompression inconsistency, but only in large indexes (specifically, ClueWeb). I couldn't even find it with unit tests. This index slicing helped me to find it.

gustingonzalez:
> What is the use case?

Mainly to avoid the entire index load when testing. Particularly, refactoring and playing with the chunk size of the posting list, I was obtained a decompression inconsistency, but only in large indexes (specifically, ClueWeb). I couldn't even find it with unit tests. This index slicing helped me to find it.

PR #304, opened 2020-01-02
Pass thresholds to heap

amallia:
Set the threshold of the heap before running the query.

### DIFF
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -39,7 +39,7 @@ struct topk_queue {
         return true;
     }
 
-    bool would_enter(float score) const { return m_q.size() < m_k || score > m_threshold; }
+    bool would_enter(float score) const { return score > m_threshold; }
 
     void finalize() {
         std::sort_heap(m_q.begin(), m_q.end(), min_heap_order);
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -6,6 +6,7 @@
 
 namespace pisa {
 
+using Threshold = float;
 struct topk_queue {
     using entry_type = std::pair<float, uint64_t>;
 
@@ -52,6 +53,10 @@ struct topk_queue {
     }
 
     [[nodiscard]] std::vector<entry_type> const &topk() const noexcept { return m_q; }
+
+    void set_threshold(Threshold t) noexcept {
+        m_threshold = t;
+    }
 
     void clear() noexcept {
         m_q.clear();
--- a/src/queries.cpp
+++ b/src/queries.cpp
@@ -46,7 +46,8 @@ void extract_times(Fn fn,
     for (auto &&[qid, query] : enumerate(queries)) {
         do_not_optimize_away(fn(query, thresholds[qid]));
         std::generate(times.begin(), times.end(), [&fn, &q = query, &t = thresholds[qid]]() {
-            return run_with_timer<std::chrono::microseconds>([&]() { do_not_optimize_away(fn(q, t)); })
+            return run_with_timer<std::chrono::microseconds>(
+                       [&]() { do_not_optimize_away(fn(q, t)); })
                 .count();
         });
         auto mean =
@@ -149,12 +150,10 @@ void perftest(const std::string &index_filename,
     if (thresholds_filename) {
         std::string t;
         std::ifstream tin(*thresholds_filename);
-    size_t idx = 0;
-        while (std::getline(tin, t)) {
-            thresholds[idx] = std::stof(t);
-        idx += 1;
-        }
-        assert(idx == queries.size());
+        auto t_begin = std::istream_iterator<float>(tin);
+        auto t_end = std::istream_iterator<float>();
+        assert(std::distance(t_begin, t_end) == queries.size());
+        std::copy(t_begin , t_end, thresholds.begin());
     }
 
     auto scorer = scorer::from_name(scorer_name, wdata);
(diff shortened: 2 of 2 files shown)


src/queries.cpp
```diff
@@ -140,13 +145,16 @@ void perftest(const std::string &index_filename,
...
     if (thresholds_filename) {
         std::string t;
         std::ifstream tin(*thresholds_filename);
+    size_t idx = 0;
         while (std::getline(tin, t)) {
-            thresholds.push_back(std::stof(t));
+            thresholds[idx] = std::stof(t);
+        idx += 1;
         }
+        assert(idx == queries.size());
```
elshize: There are some weird indents here, but more importantly, can we simply use `transform`?
```cpp
std::copy(std::istream_iterator<float>(tin), std::istream_iterator<float>(), thresholds.begin());
```

src/queries.cpp
```diff
@@ -156,77 +164,84 @@ void perftest(const std::string &index_filename,
...
-            query_fun = [&](Query query) {
+            query_fun = [&](Query query, Threshold) {
                 or_query<true> or_q;
                 return or_q(make_cursors(index, query), index.num_docs());
             };
         } else if (t == "wand" && wand_data_filename) {
-            query_fun = [&](Query query) {
+            query_fun = [&](Query query, Threshold t) {
                 topk_queue topk(k);
+        topk.set_threshold(t);
```
elshize: Idk what's up with these indents, but can we run `clang-format` please?

elshize (changes requested):
Looks good but left minor comments.

src/queries.cpp
```diff
@@ -149,12 +150,10 @@ void perftest(const std::string &index_filename,
...
         std::ifstream tin(*thresholds_filename);
-    size_t idx = 0;
-        while (std::getline(tin, t)) {
-            thresholds[idx] = std::stof(t);
-        idx += 1;
-        }
-        assert(idx == queries.size());
+        auto t_begin = std::istream_iterator<float>(tin);
+        auto t_end = std::istream_iterator<float>();
+        assert(std::distance(t_begin, t_end) == queries.size());
```
elshize: :+1: for the check but it won't make a difference with `assert` because I doubt you will run it in debug anyway. There's no reason to disable this for release, so I suggest a hard fail here. Maybe we should have some kind of `unconditional_assert` or `release_assert` or whatever better name you can come up with, because I'm expecting this to be in several places at least. I suggest it to simply `spdlog::error(...)` a message and then invoke `std::exit(1)`.

I used something similar but throwing error instead. But it's not a bad idea to have a version that exits right away, too.

```cpp
template <typename Message>
inline void runtime_assert(bool condition, Message&& message)
{
    if (not condition) {
        if constexpr (std::is_invocable_r_v<std::string, Message>) {
            throw std::runtime_error(message());
        } else {
            throw std::runtime_error(std::forward<Message>(message));
        }
    }
}
```

It's good to have something that also accepts an invokable so you don't have to evaluate stuff if `condition == true`, say:

```cpp
runtime_assert(some_condition, [&] { return string_builder.str(); });
```
Here, you only pay for the `if` if everything's okay. I learned that from rust where it's a standard lint: https://rust-lang.github.io/rust-clippy/master/#or_fun_call :)

JMMackenzie:
This looks fine to me.

A quick note: what we did in our 2019 SIGIR paper was to actually maintain two thresholds, one which checks against the *initial* value (the one passed in), and another which checks the actual heap threshold.

They both have two specific use cases. The first one is to check whether you want to *score* a document, and the second is to check whether you want to put that document into the heap.

We did this because if you set `m_threshold` as some value, and you score a document based on the upper-bound sum, it's quite possible that the document will have `score < m_threshold`. However, since we already went to the trouble of scoring it, we may as well put it into the heap (this was our logic). This is why we kept two thresholds --- one would be for managing the flow control of the dynamic pruning, and the other to manage the actual top-k heap.

I think either way is fine, but some care should be taken (ie document the behavior).

include/pisa/topk_queue.hpp
```diff
@@ -53,6 +54,10 @@ struct topk_queue {
 
     [[nodiscard]] std::vector<entry_type> const &topk() const noexcept { return m_q; }
 
+    void set_threshold(Threshold t) noexcept {
```
JMMackenzie, line 57: See my outer level comment :-)

elshize:
> we may as well put it into the heap (this was our logic).

You mean in case its partial score is somehow higher than another's full score?

amallia:
From my understanding, dynamic pruning is managed by the threshold we pass; inserts are managed by the actual threshold.

It's true that you dont want to waste a scored documents, but if you know the actual threshold what is the point of inserting if you know for sure that another one will substitute it sooner or later.
It is a different story if you know that your threshold can overshoot. Am I right?

elshize:
@amallia This is my understanding, Joel's thing makes sense only if we overshoot and keep the results; otherwise, or if we rerun query in case of overestimating, then it makes no difference. Is that right @JMMackenzie ?

JMMackenzie:
> > we may as well put it into the heap (this was our logic).
> 
> You mean in case its partial score is somehow higher than another's full score?

Not quite. On the outer level, you are asking whether there is enough evidence, based on the upper-bound scores, to bother scoring some document. If yes, you are going to score the whole thing (assuming DaaT). So let us consider an example where we have two thresholds, one is the *scoring_threshold* used for dynamic pruning, and the other is the *heap_entry_threshold* used for putting documents into the heap.

Assuming a cold start, we have an empty heap and a *score_threshold* of, say, 5:
1. We find a document _d_ with a upper-bound score > 5 - we must score this document.
2. We score _d_ and find that it has a true score of 3.

In the current approach, _d_ will get discarded and processing continues. At the end of processing, we may have scored a number of documents, but our heap will be empty.

In the approach I am discussing, we will have at least n documents in the heap, where n is the number of times the *score_threshold* was exceeded. This seems better (in practice) because we still have some (possibly bad) results, but at least we can surface some docs.

Make sense?

JMMackenzie:
@amallia 
@elshize 

If you are certain that you'll have at least _k_ documents which have a score above your initial threshold, then yes, it wont make any difference. We only did this because as you said, our predictions had the potential to overshoot.

elshize:
Yes, however, what we're currently interested (this might change?) is safe processing, and the fact that we don't have enough documents means we overshot, and need to re-run. Otherwise we wouldn't be able to know that, would we?

JMMackenzie:
> Yes, however, what we're currently interested (this might change?) is safe processing, and the fact that we don't have enough documents means we overshot, and need to re-run. Otherwise we wouldn't be able to know that, would we?

Actually, you can tell: if `score_threshold <= heap_entry_threshold` then you found all safe docs. But there needs to be some careful implementation to make sure you update the (right) thresholds at the right times. But yep, as you said, if you only want correct results, this approach (the current approach) is fine :-)

elshize (approved):

PR #307, opened 2020-01-14
Fixes for #306 

### LINKED ISSUE #306: Incorrect size for binary sequence in Invert
The beginning of the `.docs` canonical expects a single sequence, `1 N` where `N` represents the number of documents in the index. However, the code seems to take the number of terms instead.

https://github.com/pisa-engine/pisa/blob/5a7f149fb6a54e168ce70c51179e967c2e3bbe6a/include/pisa/invert.hpp#L176

We should double check that this is a bug, and then replace this with the correct size. I think we can use something like:
```
std::uint32_t count = index.document_sizes.size();
```

amallia:
Fixes #306.

### DIFF
--- a/include/pisa/invert.hpp
+++ b/include/pisa/invert.hpp
@@ -173,7 +173,7 @@ namespace invert {
         std::ofstream dstream(basename + ".docs");
         std::ofstream fstream(basename + ".freqs");
         std::ofstream sstream(basename + ".sizes");
-        std::uint32_t count = index.documents.size();
+        std::uint32_t count = index.document_sizes.size();
         write_sequence(dstream, gsl::make_span<uint32_t const>(&count, 1));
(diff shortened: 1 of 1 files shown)


elshize:
Do we know why we didn't experience any problems? Is that because we don't use it when compressing? I found only one usage in `wand_data_compressed` but I guess we haven't been using it for any experiments...

elshize (approved):

amallia:
@elshize the reason is that we never use that value. Please note this is not the value that is written into the index, which we write correctly (https://github.com/pisa-engine/pisa/blob/master/include/pisa/invert.hpp#L292)
It is only the value we write into the batches, but it is not really used in the end.

JMMackenzie (approved):

JMMackenzie:
Thanks for fixing!

PR #308, opened 2020-01-14
Fix thresholds load

amallia:
Thresholds do not get load with std::copy because `std::istream_iterator` does not support `\n` as separator


JMMackenzie (approved):

elshize (approved):

elshize:
> Thresholds do not get load with std::copy because `std::istream_iterator` does not support `\n` as separator

That's bizarre, didn't know that. Anyway, looks good.

PR #310, opened 2020-01-21
Added tool to map textual queries to our query ID format

amallia:


src/map_queries.cpp
```diff
@@ -0,0 +1,43 @@
...
+int main(int argc, const char **argv)
+{
+    spdlog::drop("");
+    spdlog::set_default_logger(spdlog::stderr_color_mt(""));
+
+    std::string query_filename;
+    std::string terms_filename;
+    std::optional<std::string> stopwords_filename;
+    std::optional<std::string> stemmer = std::nullopt;
+    std::string separator = "\t";
```
JMMackenzie, line 20: What sort of input are we expecting here? I thought our queries were formatted like:
```
123:hello abc test
```

JMMackenzie (commented):
Hey, looks fine, quick question about the format though :)

src/map_queries.cpp
JMMackenzie: Wait, I think I follow now.

This program reads queries of the form:
```
123:hello abc test
```

And outputs them like
```
1\t2\t3
```
where `1,2,3` are the identifiers for `hello abc test` - right?

Perhaps we should also output the query identifier?

amallia: Yes you are right.

I can add the query identifier too

src/map_queries.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    using boost::adaptors::transformed;
+    using boost::algorithm::join;
+    for (auto &&q : queries) {
+        if(query_id and q.id) {
+            std::cout << *(q.id) << ":";
+        }
+        std::cout
+            << join(q.terms | transformed([](auto d) { return std::to_string(d); }), separator)
+            << std::endl;
```
elshize: Flushing with `std::endl` is very inefficient, just replace with `'\n'`. Also throw a `clang-format` for a good measure.

elshize (approved):
Looks good. One nit pick for the future: since `\n` is one character, we can further improve performance by passing it as such `'\n'`. I don't expect it to make a difference here, though. Still, it's a good default because why not?

PR #311, opened 2020-01-22
Count postings

amallia:


src/count_postings.cpp
```diff
@@ -0,0 +1,34 @@
...
+    CLI11_PARSE(app, argc, argv);
+
+    binary_collection coll((input_basename + ".docs").c_str());
+    auto firstseq = *coll.begin();
+    if (firstseq.size() != 1) {
+        throw std::invalid_argument("First sequence should only contain number of documents");
+    }
+
+    std::size_t count = 0;
+    for(auto&& p : coll) {
```
elshize, line 28: So isn't it true that we're now counting in the first "fake" sequence? It's not a big deal because it's just one "posting" but still not correct, is it?

JMMackenzie (approved):
I approve with the condition of @elshize comment about the first "fake" sequence (we'll be off by 1 :-)

elshize (approved):

PR #313, opened 2020-02-05
Common CLI arguments

### LINKED ISSUE #93: Easier use of standard command line options

elshize:
Common CLI arguments are extracted to a separate file to clean up tools.

**FYI** `shuffle_docids.cpp` is the same but I applied `clang-format`

Fixes: #93 #135

### DIFF
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -118,15 +118,18 @@ namespace arg {
     struct Thresholds {
         explicit Thresholds(CLI::App *app)
         {
-            app->add_option(
-                   "-T,--thresholds", m_thresholds_filename, "File containing query thresholds")
-                ->required();
+            m_option = app->add_option("-T,--thresholds",
+                                       m_thresholds_filename,
+                                       "File containing query thresholds")
+                           ->required();
         }
 
         [[nodiscard]] auto thresholds_file() const { return m_thresholds_filename; }
+        [[nodiscard]] auto *thresholds_option() { return m_option; }
 
        private:
         std::string m_thresholds_filename;
+        CLI::Option *m_option;
     };
 
     struct Threads {
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -15,11 +15,21 @@ namespace pisa {
 
 namespace arg {
 
+    enum class Encoding : bool { Legacy, New };
+
+    template <Encoding encoding = Encoding::New>
     struct Index {
         explicit Index(CLI::App *app)
         {
             app->add_option("-i,--index", m_basename, "Inverted index basename")->required();
-            app->add_option("-e,--encoding", m_encoding, "Index encoding")->required();
+            if constexpr (encoding == Encoding::New) {
+                app->add_option("-e,--encoding", m_encoding, "Index encoding")->required();
+            } else {
+                app->add_option("-e,-t,--encoding,--type",
+                                m_encoding,
+                                "Index encoding\nDEPRECATED: -t/--type; USE: -e/--encoding")
+                    ->required();
+            }
             app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
             app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file");
         }
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -1,0 +1,174 @@
+#pragma once
+
+#include <optional>
+#include <string>
+#include <thread>
+
+#include <CLI/CLI.hpp>
+#include <range/v3/view/getlines.hpp>
+#include <range/v3/view/transform.hpp>
+
+#include "io.hpp"
+#include "query/queries.hpp"
+
+namespace pisa {
+
+namespace arg {
+
+    struct Index {
+        explicit Index(CLI::App *app)
+        {
+            app->add_option("-i,--index", m_basename, "Inverted index basename")->required();
+            app->add_option("-e,--encoding", m_encoding, "Index encoding")->required();
+            app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
+            app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file");
+        }
+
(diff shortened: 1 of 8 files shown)


tools/app.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+namespace pisa {
+
+namespace arg {
+
+    struct Index {
+        explicit Index(CLI::App *app)
+        {
+            app->add_option("-i,--index", m_basename, "Inverted index basename")->required();
+            app->add_option("-e,--encoding", m_encoding, "Index encoding")->required();
```
JMMackenzie, line 22: I agree that `-e` is better than `-t` but I'm sad that it will break my scripts...

JMMackenzie (approved):
Looks good to me!

tools/app.hpp
elshize: Yeah, this is up for debate, but I think it will hurt less now than later if we wanted to change it :) I think we can also simply use both with `"-i,-t,--encoding"`. But I'm not so sure that's a great idea. @amallia thoughts?

elshize: @JMMackenzie @amallia take a look at the latest commit and tell me what you think.

JMMackenzie: Looks cool, that's one way to fix it in the meantime :-) We can then deprecate the legacy options later? Though I am willing to pay the price mow if we want to just go for the new options.

elshize: I think the good time would probably be v1.0 to get rid of it. But on the other hand, maybe we want to reuse `-t` for something different, then we might wanna remove it right away. I don't have a clear preference. I think it might come down to @amallia

tools/app.hpp
```diff
@@ -0,0 +1,184 @@
...
+       private:
+        std::string m_basename;
+        std::string m_encoding;
+        std::optional<std::string> m_wand_data_path;
+        bool m_wand_compressed = false;
+    };
+
+    enum class QueryMode : bool { Ranked, Unranked };
+
+    template <QueryMode Mode = QueryMode::Ranked, int DefaultK = 1'000>
```
amallia: I propose not to have a default `k`. You should always specify it. The reason is that not in all the tasks we print it and so it might cause confusion if you accidentally forget to specify it.

tools/app.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+        [[nodiscard]] auto algorithm() const -> std::string const & { return m_algorithm; }
+
+       private:
+        std::string m_algorithm;
+    };
+
+    enum class Required : bool { Required, Optional };
+
+    template <Required required = Required::Required>
```
amallia: Why the scorer can be optional? I think it should be always required.

tools/app.hpp
```diff
@@ -0,0 +1,184 @@
...
+            if constexpr (encoding == Encoding::New) {
+                app->add_option("-e,--encoding", m_encoding, "Index encoding")->required();
+            } else {
+                app->add_option("-e,-t,--encoding,--type",
+                                m_encoding,
+                                "Index encoding\nDEPRECATED: -t/--type; USE: -e/--encoding")
+                    ->required();
+            }
+            app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
+            app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file");
```
amallia: Should we make this flag depend on `wand`?

tools/app.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+    enum class QueryMode : bool { Ranked, Unranked };
+
+    template <QueryMode Mode = QueryMode::Ranked, int DefaultK = 1'000>
+    struct Query {
+        explicit Query(CLI::App *app)
+        {
+            app->add_option("-q,--queries", m_query_file, "Path to file with queries", false);
+            app->add_option(
+                "--stopwords", m_stop_words, "List of blacklisted stop words to filter out");
```
amallia: Depends on `queries` and maybe `terms`?
You pass stopwords, but than you dont pass any `stemmer` or `terms` file. What happens?

tools/app.hpp
```diff
@@ -0,0 +1,184 @@
...
+            } else {
+                return m_scorer;
+            }
+        }
+
+       private:
+        std::optional<std::string> m_scorer;
+    };
+
+    template <Required required = Required::Optional>
```
amallia: I have problem with defaults. Sorry :)
Thresholds are always optional, right?

amallia (changes requested):

tools/app.hpp
elshize: Makes sense.

tools/app.hpp
elshize: I was probably thinking of when it's precomputed. But I guess we'd still specify a scorer. I'll get rid of it.

amallia (approved):

PR #314, opened 2020-02-06
Change default run name from "R0" to "PISA"

JMMackenzie:
I noticed most other systems use the system name as the default, I think it's a bit nicer.


elshize (approved):
Sounds good.

amallia (approved):

PR #315, opened 2020-02-07
99% quantile in queries tool

amallia:


JMMackenzie (approved):
Can we add it to the `stats_line` as well?

Otherwise looks good to me

amallia:
> Can we add it to the `stats_line` as well?
> 
> Otherwise looks good to me

Added. Thanks

elshize (approved):

PR #318, opened 2020-02-24
Fix maskedvbyte and varintgb buffer initialization when encoding

### LINKED ISSUE #317: Output buffer issue in maskedvbyte and varintgb encoders
There is an issue with the encoding buffer of `maskedvbyte` and `varintgb` encoders:

https://github.com/pisa-engine/pisa/blob/84a1f3ee528083f4dd7675a133a6ddc73b4da3ec/include/pisa/codec/maskedvbyte.hpp#L24

https://github.com/pisa-engine/pisa/blob/84a1f3ee528083f4dd7675a133a6ddc73b4da3ec/include/pisa/codec/varintgb.hpp#L251

Since `thread_local` is used to define the `buf` variable, it's initialized only at first `encode()` call. Therefore, if the following call exceeds the size of the previously defined `buf`, then a `segmentation fault` will be thrown.

Anyway, there are no consequences by now, because if `n < block_size` then interpolative coding is used. This means that the buffer is defined when `n == block_size` (which is enough for any output). Even so, conceptually it sounds something weird, and it's a potential bug in case of not use interpolative coding for chunks smaller than the `block_size`.

Possible solutions:
- Set buffer size as `2 * block_size * sizeof(uint32_t)` instead `2 * n * sizeof(uint32_t)`.
- Define the buffer size by using the `resize()` method.

gustingonzalez:
Fixes #317.

### DIFF
--- a/include/pisa/codec/maskedvbyte.hpp
+++ b/include/pisa/codec/maskedvbyte.hpp
@@ -21,7 +21,7 @@ struct maskedvbyte_block {
             interpolative_block::encode(src, sum_of_values, n, out);
             return;
         }
-        thread_local std::vector<uint8_t> buf(2 * n * sizeof(uint32_t));
+        thread_local std::vector<uint8_t> buf(2 * block_size * sizeof(uint32_t));
         size_t out_len = vbyte_encode(src, n, buf.data());
         out.insert(out.end(), buf.data(), buf.data() + out_len);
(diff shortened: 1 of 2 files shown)


JMMackenzie (approved):

elshize (approved):
Fantastic!

PR #319, opened 2020-02-24
Add support for rerunning unsafe queries

elshize:


tools/queries.cpp
```diff
@@ -60,16 +60,25 @@ void op_perftest(Functor query_func,
...
+    std::size_t num_reruns = 0;
+    spdlog::info("Safe: {}", safe);
 
     for (size_t run = 0; run <= runs; ++run) {
         size_t idx = 0;
         for (auto const &query : queries) {
             auto usecs = run_with_timer<std::chrono::microseconds>([&]() {
                 uint64_t result = query_func(query, thresholds[idx]);
+                if (safe && result < k) {
+                    spdlog::warn("results ({}) < k ({})", result, k);
```
amallia: We are timing this. Can we drop this log line?

tools/queries.cpp
```diff
@@ -152,7 +163,7 @@ void perftest(const std::string &index_filename,
         std::ifstream tin(*thresholds_filename);
         size_t idx = 0;
         while (std::getline(tin, t)) {
-            thresholds[idx] = std::stof(t);
+            thresholds[idx] = std::stof(t) - 0.0001;
```
amallia: Maybe we should fix the thresholds instead. 
I propose to get rid of `- 0.0001` here and fix the way we compute thresholds in a separate PR.

amallia (changes requested):
Looks good. I would like to change just a couple of things

amallia (approved):

JMMackenzie (changes requested):
If you don't pass any thresholds then `safe` should remain false. Otherwise any query with fewer than `k` results will run twice for no reason, right?

JMMackenzie (approved):

PR #321, opened 2020-02-27
Quantized index

amallia:
Replaces #230


include/pisa/quantizer.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include "configuration.hpp"
+
+namespace pisa {
+
+float dequantize(uint64_t value)
+{
+    const float quant = 1.f / configuration::get().reference_size;
+    return quant * value;
```
elshize: Why not `static_cast<float>(value) / reference_size`?

include/pisa/quantizer.hpp
```diff
@@ -0,0 +1,21 @@
...
+}
+
+uint64_t quantize(float value)
+{
+    float quant = 1.f / configuration::get().reference_size;
+    uint64_t pos = 1;
+    while (value > quant * pos)
+        pos++;
+    return pos;
+}
```
elshize: If I understand correctly, this is supposed to take a number from between 0 and 1, correct? If so, then I would add a debug assert for that.

But also, isn't what you're doing here simply `ceil(value * reference_size)`?

Also, about the design: this is quite confusing, I must say. It took me a while to understand what and why you're doing this. How about we do something like the following:

```cpp
template <std::uin32_t bytes = 1>
struct LinearQuantizer {
    explicit LinearQuantizer(float max) : m_max(max), m_scale(static_cast<float>(1 << (bytes * 8)) / max) {
        Expects(bytes <= 4);
    }
    [[nodiscard]] auto operator()(float value) const -> std::uint32_t {
        Expects(value <= m_max);
        return static_cast<float>(value) * m_scale;
    }
  private:
    float m_max;
    float m_scale;
};
```
Note that I used intentionally `uint32_t` because I think there's no application where we want to exceed 4 bytes, at least I think so.

Now, you can later simply build it with `LinearQuantizer quantizer(max);` and use it `quantizer(score)`.

I think the intention is much clearer.

Of course you may also now implement dequantizer within the same struct...

include/pisa/scorer/quantized.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include "index_scorer.hpp"
+namespace pisa {
+
+template <typename Wand>
+struct quantized : public index_scorer<Wand> {
+
+    using index_scorer<Wand>::index_scorer;
+    term_scorer_t term_scorer(uint64_t term_id) const
+    {
+        auto s = [&](uint32_t doc, uint32_t freq) {
```
elshize: Unnecessary reference capture.

include/pisa/wand_data_compressed.hpp
```diff
@@ -91,6 +89,7 @@ class uniform_score_compressor {
     }
 };
 
+template<bool QuantizedIndex = false>
```
elshize: Here's the problem with `bool`s as template params: when you get to creating the object, it's `wand_data_compressed<true>`. But `true` what?

Instead I suggest:

```cpp
enum class ScoreType : bool { Float, Quantized };
```

It still can act like a bool, but it has a name.

elshize: Also, `clang-format`

tools/create_freq_index.cpp
```diff
@@ -86,25 +137,28 @@ void create_collection(InputCollection const &input,
 
     if (output_filename) {
         mapper::freeze(coll, (*output_filename).c_str());
-        if (check) {
-            verify_collection<InputCollection, CollectionType>(input,
-                                                               (*output_filename).c_str());
+        if (check and not quantized) {
```
elshize, line 144: How about if `check and quantized` then we print out a warning or info message that it's not being verified? That or block it at CLI. I think a message would be fine at the moment.

tools/queries.cpp
```diff
@@ -326,32 +330,46 @@ int main(int argc, const char **argv)
...
-                                                              app.scorer(),              \
-                                                              extract,                   \
-                                                              safe);                     \
+#define LOOP_BODY(R, DATA, T)                                                                \
+    }                                                                                        \
+    else if (app.index_encoding() == BOOST_PP_STRINGIZE(T))                                  \
+    {                                                                                        \
+        if (app.is_wand_compressed()) {                                                      \
+            if (quantized) {                                                                 \
+                perftest<BOOST_PP_CAT(T, _index), wand_uniform_index_quantized>(             \
```
elshize: Actually, here's what you can do if you apply the same arguments to different functions depending on a condition:

```cpp
auto params = std::make_tuple(a, b, c, d, e);
...
if (a) { std::apply(f<A>, params); }
else if (b) { std::apply(f<B>, params); }
else if (c) { std::apply(f<C>, params); }
```

tools/thresholds.cpp
```diff
@@ -57,13 +60,17 @@ void thresholds(const std::string &index_filename,
         float threshold = 0.0;
         if (results.size() == k) {
             threshold = results.back().first;
+            if (quantized) {
+                threshold = dequantize(threshold) * wdata.index_max_term_weight();
```
elshize: Why do we want to unquantize those? Don't we need to use quantized thresholds with a quantized index?

elshize (approved):

include/pisa/configuration.hpp
```diff
@@ -1,71 +1,72 @@
...
+        fillvar("PISA_FIXCOST", fix_cost, 64);
+        fillvar("PISA_LOG_PART", log_partition_size, 7);
+        fillvar("PISA_THRESHOLD_WAND_LIST", threshold_wand_list, 0);
+        fillvar("PISA_THREADS", worker_threads, std::thread::hardware_concurrency());
+        fillvar("PISA_HEURISTIC_GREEDY", heuristic_greedy, false);
+        fillvar("PISA_FIXED_COST_WAND_PARTITION", fixed_cost_wand_partition, 12.0);
+        fillvar("PISA_EPS1_WAND", eps1_wand, 0.01);
+        fillvar("PISA_EPS2_WAND", eps2_wand, 0.4);
+        fillvar("PISA_SCORE_REFERENCES_SIZE", reference_size, 128);
+        fillvar("PISA_QUANTIZTION_BITS", quantization_bits, 8);
```
elshize, line 57: Is it used anywhere at all? (I can't see it anywhere.) If not, let's delete. If so, does it really need to be?

include/pisa/linear_quantizer.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <gsl/gsl_assert>
+#include <cmath>
+
+namespace pisa {
+
+struct LinearQuantizer {
+    explicit LinearQuantizer(float max, uint8_t bits) : m_max(max), m_scale(static_cast<float>(1 << (bits)) / max) {
+       Expects(bits <= 32);
```
elshize: Let's do at least `Expects(bits <= 32 and bits > 0)`. We can also set a min bits if that makes sense. Bit 0 definitely doesn't and can probably do some ugly shit if passed.

elshize: Now that I think about it, we can easily afford a runtime check here, so we should just throw if incorrect.
```cpp
if (bits > 32 or bits == 0) {
    throw std::runtime_error(fmt::format("Linear quantizer must take a number of bits between 1 and 32 but {} passed", bits));
}
```

include/pisa/linear_quantizer.hpp
```diff
@@ -0,0 +1,20 @@
...
+    }
+    [[nodiscard]] auto operator()(float value) const -> std::uint32_t {
+       Expects(value <= m_max);
```
elshize: `clang-format`

tools/queries.cpp
```diff
@@ -328,48 +328,36 @@ int main(int argc, const char **argv)
...
+                                  app.queries(),
+                                  app.thresholds_file(),
+                                  app.index_encoding(),
+                                  app.algorithm(),
+                                  app.k(),
+                                  app.scorer(),
+                                  extract,
+                                  safe);
+    // ...
+    // if (a) { std::apply(f<A>, params); }
```
elshize: Forgot to remove these...

elshize (changes requested):

include/pisa/configuration.hpp
elshize: My bad, I was looking at a wrong diff then. Still I think this should be handled through CLI default value. This is some dark magic right here.

amallia: Deferred to https://github.com/pisa-engine/pisa/issues/322

include/pisa/wand_data_compressed.hpp
```diff
@@ -29,18 +31,16 @@ class uniform_score_compressor {
...
+            LinearQuantizer quantizer(max_score, configuration::get().quantization_bits);
             std::vector<uint32_t> score_indexes;
             score_indexes.reserve(effective_scores.size());
             for (const auto &score : effective_scores) {
-                size_t pos = 1;
-                while (score > quant * pos)
-                    pos++;
-                score_indexes.push_back(pos - 1);
+                auto q = quantizer(score);
+                score_indexes.push_back(q - 1);
```
elshize: Not a big deal but if you introduce a variable named `q`, might as well just call it directly in `score_indexes.push_back(quantizer(score) - 1);` now, if that were some informative name with a description, sure, you might wanna do that, or if the call is long and you want to make it cleaner. Here, it just makes you look at two lines to understand one.

include/pisa/wand_data_compressed.hpp
```diff
@@ -86,11 +86,14 @@ class uniform_score_compressor {
 
     static float inline score(uint32_t index)
```
elshize: By `index` you mean `quantized_score`? If so, can we just name it this way?

include/pisa/wand_data_compressed.hpp
```diff
@@ -113,27 +116,20 @@ class wand_data_compressed {
...
-                }
-                auto ind = compressor_builder.compress_data(t.second);
-
-                compressor_builder.add_posting_list(t.first.size(), t.first.begin(), ind.begin());
-
                 max_term_weight.push_back(max_score);
                 total_elements += seq.docs.size();
                 total_blocks += t.first.size();
+
+                tmp_docs.push_back(std::move(t.first));
```
elshize: I think `tmp_docs` just a terrible name. I'm a bit confused myself as what it is. Is that the max block documents and `tmp_scores` are the unquantized block max scores? Maybe we should say it then, `unquantized_block_max_scores` and `block_max_documents` or something like that?

tools/evaluate_queries.cpp
```diff
@@ -207,33 +212,48 @@ int main(int argc, const char **argv)
...
+                    app.wand_data_path(),                                                    \
+                    app.queries(),                                                           \
+                    app.thresholds_file(),                                                   \
+                    app.index_encoding(),                                                    \
+                    app.algorithm(),                                                         \
+                    app.k(),                                                                 \
+                    documents_file,                                                          \
+                    app.scorer(),                                                            \
+                    run_id);                                                                 \
+            } else {                                                                         \
```
elshize: You can use the same trick with tuple as in `queries` to make it simpler and less error-prone.

tools/thresholds.cpp
```diff
@@ -73,31 +76,48 @@ int main(int argc, const char **argv)
...
+            if (quantized) {                                                                  \
+                thresholds<BOOST_PP_CAT(T, _index), wand_uniform_index_quantized>(            \
+                    app.index_filename(),                                                     \
+                    app.wand_data_path(),                                                     \
+                    app.queries(),                                                            \
+                    app.index_encoding(),                                                     \
+                    app.scorer(),                                                             \
+                    app.k(),                                                                  \
+                    quantized);                                                               \
+            } else {                                                                          \
```
elshize: Again, same as for `evaluate_queries`

elshize (changes requested):

elshize (approved):

PR #323, opened 2020-03-05
Initial interface for index v1

elshize:
**Work in progress**


include/pisa/index_metadata.hpp
```diff
@@ -0,0 +1,106 @@
...
+/// have a postings file that contains data for one posting list after another,
+/// and an offset file.
+struct PostingFilePaths {
+    std::filesystem::path postings;
+    std::filesystem::path offsets;
+};
+
+/// Index metadata stores required paths and properties of an index.
+struct IndexMetadata {
+    /// Path to the metadata file. I suggest YAML.
```
JMMackenzie, line 34: Just pointing back to a prior discussion on filetypes for metadata: https://github.com/pisa-engine/pisa/issues/143

JMMackenzie (commented):
I need to think a little more about storing things (and whether to have things together or not) but I think this is nice. I also pointed back to an old issue where we discussed storage of metadata.

elshize:
> I need to think a little more about storing things (and whether to have things together or not) but I think this is nice. I also pointed back to an old issue where we discussed storage of metadata.

@JMMackenzie Do you have any concerns about storing documents and frequencies/scores separately other than any possible performance impact?

JMMackenzie:
> > I need to think a little more about storing things (and whether to have things together or not) but I think this is nice. I also pointed back to an old issue where we discussed storage of metadata.
> 
> @JMMackenzie Do you have any concerns about storing documents and frequencies/scores separately other than any possible performance impact?

Other than performance, just for simplicity. It's kinda annoying having numerous files for a single index.

elshize:
The performance thing is fair. I think we can match performance or at least be very close (especially for precomputed scores) but needs more investigation. As far as many files, I guess, but also I think we should have one metadata file and then we don't have to manually keep track of all the files, which means we have actually fewer files to keep track of (one) than now (index, wand, termlex, doclex, etc.).

Here's my pitch why this is good:
- It saves space: if we have documents and payloads together, it means that if we want to precompute scores, we essentially need to duplicate all the posting lists. If you want to precompute several scores (say, BM25 and QL, and what have you), that's a lot of unnecessary bytes.
- It actually simplifies a lot in the long run: If we come up with a good working abstraction, we can easily test different combinations, say, PEF for IDs and SIMD-BP for freqs/scores, or compressed documents with uncompressed 1-byte scores, etc. It's very composable. Now, it more messy, more code, therefore more bug-prone and difficult to maintain. Note that our `freq_index` (e.g., for PEF) is a separate entity and it does already separate documents from frequencies. At least for this one, by simply putting frequencies and precomputed scores in separate files from documents, we avoid duplicating all those documents, without any real disadvantage.

Another, slightly different approach would be to keep frequencies in the same file as documents, and simply say that it's always there (which we would assume anyway). But I would still remove that interleaving of documents and frequencies because then we have to access all that data we're not using if we use precomputed scores.

Any opinions about what I just said?

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+    [[nodiscard]] auto document_length(DocId docid) const -> std::uint32_t;
+    [[nodiscard]] auto avg_document_length() const -> float;
+    [[nodiscard]] auto normalized_document_length(DocId docid) const -> float;
+
+   protected:
+    [[nodiscard]] auto fetch_postings(TermId term, PostingData const& data) const
+        -> gsl::span<std::byte const>;
+
+   private:
+    gsl::span<std::uint32_t const> m_document_lengths;
```
amallia, line 49: Let's think is we really want to store the document lengths. True that we need those for scoring, but do we assume that we will always score? 
I believe this data should be part of some scoring metadata file.

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+    float m_avg_document_length;
```
amallia, line 50: Same as above.

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+    std::any m_source;
```
amallia, line 51: What is this? Where is it used?

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+    config::Scorer scorer;
+};
+
+struct Bm25Scorer {
+    // TODO: all the other stuff needed.
+    config::Scorer scorer;
+};
+
+template <typename DocumentCursor, typename FrequencyCursor, typename ScoreCursor>
+class DaatIndex {
```
amallia, line 96: I think the name is misleading, but it is not a big deal for now.
DAAT is the way we process, not really the way the index is organized. Same thing for SAAT.
Better names would be `DocumentSortedIndex` and `ImpactSortedIndex`

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    [[nodiscard]] auto frequencies(TermId term) const -> FrequencyCursor
+    {
+        return m_frequency_reader.read(this->fetch_postings(term, m_frequencies));
+    }
+
+    [[nodiscard]] auto quantized_scores(config::Scorer scorer, TermId term) const -> ScoreCursor
+    {
+        return m_score_readers.at(scorer).read(this->fetch_postings(term, m_scores.at(scorer)));
+    }
```
amallia, line 113: I think that either we want to get the `frequencies` or the `quantized_scores` but never both.

include/pisa/index.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    template <typename Scorer>
+    [[nodiscard]] auto scored_postings(
+        TermId term,
+        Scorer scorer,
+        std::optional<MaxScoreType> max_score_type = std::nullopt) const;
+
+   private:
+    PostingData m_documents;
+    PostingData m_frequencies;
```
amallia, line 125: One small overhead of storing the two separately is that we need to double the offsets. Not saying it is a bad thing, just saying that it is an additional cost.

include/pisa/index_metadata.hpp
```diff
@@ -0,0 +1,106 @@
...
+    /// Returns query parser with appropriate stemming.
+    [[nodiscard]] auto query_parser(std::optional<std::filesystem::path> const& stop_words =
+                                        std::nullopt) const -> std::function<Query(std::string)>;
+};
+
+struct DaatIndexMetadata : public IndexMetadata {
+    /// Paths to documents.
+    PostingFilePaths documents;
+    /// Paths to frequencies.
+    PostingFilePaths frequencies;
```
amallia, line 71: Why isn't this optional? A DAAT index does not always need frequencies nor scores.

PR #325, opened 2020-03-06
clang-tidy cmake support

elshize:
This one only add cmake clang-tidy support, but does not turn it on by default.


elshize:
Closing in favor of #326

PR #327, opened 2020-03-06
Finish clang-format and add Travis check

elshize:


.travis.yml
```diff
@@ -85,14 +92,25 @@ before_install:
       export PATH="/usr/local/opt/ccache/libexec:$PATH";
     fi
   - eval "${MATRIX_EVAL}"
+  - if [[ "$CLANG_FORMAT" == "On" ]]; then
+      curl https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -;
+      echo "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main" | sudo tee -a /etc/apt/sources.list;
+      sudo apt-get -qq update;
+      sudo apt-get -y install clang-format-9;
```
amallia: Why do we need to specify the version (`9`)?

Why manual install? Can we just do
```packages: ['clang-format']```

amallia (changes requested):

.travis.yml
elshize: Because the default version is 6, and some options have changed since, and as far as I can tell, there's no way of conforming to both versions, so I think it's better to go with the new one, since it's also reacher in many aspects and have additional options.

Which version are you running locally?

PR #329, opened 2020-03-10
Fix CLI in create_freq_index

### LINKED ISSUE #328: Quantization segmentation faults
Since we now support quantization I wanted to mess around with it and came across this bug.

When building a quantized index via `create_freq_index` I get a segmentation fault.

The problem is that we instantiate a `scorer` without the scorer having `m_wdata` -- we then try to do `auto term_weight = query_term_weight(term_len, this->m_wdata.num_docs());` when we call `term_scorer` which leads to the segmentation fault.

I'm not sure what the best approach is to fix this - perhaps we can compute the normalized term lengths and the number of documents *first* and add those into the `wand_data` when we initialize it? Then, the scorer would have the necessary information to begin quantization.

I have also added some minor fixes to the quantization code in the branch `quant-fixes` so we can develop on that for this fix.

amallia:
Fixes #328 
Thanks @JMMackenzie

### DIFF
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -117,6 +117,27 @@ namespace arg {
         std::string m_algorithm;
     };
 
+    struct Quantize {
+        explicit Quantize(CLI::App* app)
+        {
+            auto* wand = app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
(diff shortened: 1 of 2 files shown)


elshize (approved):
:rocket:

PR #330, opened 2020-03-10
Removed almost all config

amallia:
Our way to config removal


include/pisa/sequence/partitioned_sequence.hpp
```diff
@@ -1,427 +1,412 @@
...
+    template <typename Iterator>
+    static std::vector<uint32_t> compute_partition(
+        Iterator begin,
+        uint64_t universe,
+        uint64_t n,
+        global_parameters const& params,
+        uint64_t fix_cost = 64,
+        double eps1 = 0.03,
+        double eps2 = 0.3,
+        double eps3 = 0.01)
```
elshize, line 342: Can we add some comments as to what these four are? Just a few words and maybe a reference to some other resource/paper if that's more difficult to explain.

include/pisa/wand_utils.hpp
```diff
@@ -59,30 +55,29 @@ std::pair<std::vector<uint32_t>, std::vector<float>> static_block_partition(
...
 std::pair<std::vector<uint32_t>, std::vector<float>> variable_block_partition(
-    binary_freq_collection const &coll,
-    binary_freq_collection::sequence const &seq,
+    binary_freq_collection const& coll,
+    binary_freq_collection::sequence const& seq,
     Scorer scorer,
-    const float lambda)
+    const float lambda,
+    double eps1 = 0.01,
+    double eps2 = 0.4)
```
elshize: Again, maybe we can briefly say a few words about these.

tools/create_freq_index.cpp
```diff
@@ -52,8 +52,7 @@ void dump_index_specific_stats(pisa::pefopt_index const& coll, std::string const
         }
     }
 
-    pisa::stats_line()("type", type)("eps1", conf.eps1)("eps2", conf.eps2)(
-        "fix_cost", conf.fix_cost)("docs_avg_part", long_postings / docs_partitions)(
+    pisa::stats_line()("type", type)("docs_avg_part", long_postings / docs_partitions)(
```
elshize, line 55: So what is the idea here? Are we dropping configuring these params (because they just work this way) or are planning to bring it back with command-line option?

tools/create_wand_data.cpp
```diff
@@ -34,17 +33,20 @@ int main(int argc, const char** argv)
...
     app.add_option("-o,--output", output_filename, "Output filename")->required();
-    auto var_block_opt = app.add_flag("--variable-block", variable_block, "Variable length blocks");
-    auto var_block_param_opt =
-        app.add_option("-b,--block-size", fixed_block_size, "Block size for fixed-length blocks")
-            ->excludes(var_block_opt);
-    app.add_option("-l,--lambda", lambda, "Lambda parameter for variable blocks")
-        ->excludes(var_block_param_opt)
-        ->needs(var_block_opt);
+    auto block_group = app.add_option_group("blocks");
+    auto block_size_opt = block_group->add_option(
```
elshize, line 37: :+1:

elshize (commented):
Some minor comments, but looks good. Also, would you be okay with making some of the parameters strongly-typed, instead of passing doubles around? We can do it in another PR but I just want to throw it out there, seems like little work for a lot of security, since we're mostly passing it around before taking it value for some calculation later in the process.

tools/create_freq_index.cpp
amallia: The idea is that there is no need to print these numbers. They are not configurable, so I personally don’t really need to see them every time. 
If we will make it configurable through CLI then we will print them somewhere

elshize (approved):

elshize (approved):

elshize:
Feel free to merge this if you are done.

amallia:
Ok I will do the type safe thing separately then

PR #342, opened 2020-03-18
[clang-tidy] hicpp-move-const-arg

elshize:
Added tidy check on Travis that only checks for those several things I have fixed. I also fixed some `auto` stuff that seem to have gotten lost (while merging I suppose).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ matrix:
                 - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main'
                   key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
               packages:
-                - ['clang-9', 'libstdc++-9-dev']
+                - ['clang-9', 'libstdc++-7-dev']
       env: MATRIX_EVAL="CC=clang-9 && CXX=clang++-9 && COVERAGE=Off && TIDY=On"
     - os: osx
       osx_image: xcode11.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,7 +115,7 @@ script:
         make -j2;
         if [[ "$PISA_ENABLE_CLANG_TIDY" != "On"]]; then
             CTEST_OUTPUT_ON_FAILURE=TRUE ctest -j2;
-        fi;
+        fi
       fi
     - if [[ "$CLANG_FORMAT" == "On" ]]; then
         cd ..;
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,8 @@ matrix:
       addons:
           apt:
               sources:
+                - ubuntu-toolchain-r-test
+                - llvm-toolchain-xenial-6.0
                 - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main'
                   key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
               packages:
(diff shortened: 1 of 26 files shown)


.travis.yml
```diff
@@ -84,6 +84,19 @@ matrix:
...
+      compiler: clang
+      addons:
+          apt:
+              sources:
+                - ubuntu-toolchain-r-test
+                - llvm-toolchain-xenial-6.0
+                - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-9 main'
+                  key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
+              packages:
+                - ['clang-tidy-9', 'clang-9', 'libstdc++-7-dev']
```
amallia, line 98: libstdc++-7-dev or libstdc++-8-dev? Which one is the right one here?

.travis.yml
```diff
@@ -99,10 +112,12 @@ before_install:
   
 script:
     - cd build
-    - cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=${COVERAGE} -DPISA_COMPILE_HEADERS=${PISA_COMPILE_HEADERS} ${CMAKE_OPTIONS} ..
+    - cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=${COVERAGE} -DPISA_COMPILE_HEADERS=${PISA_COMPILE_HEADERS} -DPISA_ENABLE_CLANG_TIDY=${TIDY} -DPISA_CLANG_TIDY_EXECUTABLE=clang-tidy-9 ${CMAKE_OPTIONS} ..
     - if [[ "$PISA_COMPILE_HEADERS" != "On" && "$CLANG_FORMAT" != "On" ]]; then
         make -j2;
-        CTEST_OUTPUT_ON_FAILURE=TRUE ctest -j2;
+        if [[ "$TIDY" != "On" ]]; then
```
amallia, line 118: What’s this for?

CMakeLists.txt
```diff
@@ -76,11 +77,13 @@ file(GLOB_RECURSE PISA_SRC_FILES FOLLOW_SYMLINKS "src/*cpp")
 list(SORT PISA_SRC_FILES)
 
 if (PISA_ENABLE_CLANG_TIDY)
-    find_program(CLANGTIDY clang-tidy)
+    find_program(CLANGTIDY "${PISA_CLANG_TIDY_EXECUTABLE}")
     if(CLANGTIDY)
-        set(CMAKE_CXX_CLANG_TIDY ${CLANGTIDY})
+        set(
+            CMAKE_CXX_CLANG_TIDY
+            ${CLANGTIDY} --checks=-*,hicpp-use-auto,hicpp-move-const-arg,modernize-loop-convert,readability-braces-around-statements --warnings-as-errors=*)
```
amallia: Can we have a clang tidy file for configuration?

.travis.yml
elshize: "Don't run tests if running `clang-tidy`" I figured it might make sense not to push our testing capabilities. I can turn it back on, but that means additional 15 min or so.

.travis.yml
elshize: As far as I understand, it doesn't need to be a particular version as long as it supports c++17. Note that we never install it for `gcc` builds, which means it uses whatever's already there. I don't remember why we do that for clang, but there might have been some issues with earlier versions, I presume. Note that `libstdc++` is the GNU implementation anyway, not LLVM's.

.travis.yml
amallia: But aren’t you running tests? Isn’t that ctest command inside the branch?

PR #343, opened 2020-03-18
Fix range

amallia:
I think it is incorrect to check `end + range_size < max_docid`. @elshize what do you think?


amallia:
Nevermind

PR #345, opened 2020-03-23
Readability implicit bool conversion

elshize:


test/test_wand_data.cpp
```diff
@@ -119,7 +119,7 @@ TEST_CASE("wand_data_range")
         REQUIRE(live_blocks.size() == ceil_div(collection.num_docs(), 64));
 
         for (int i = 0; i < live_blocks.size(); ++i) {
-            if (live_blocks[i] == 0) {
+            if (static_cast<int>(live_blocks[i]) == 0) {
```
amallia: `live_blocks` is a vector of book, lets just compare to `true` and `false`

include/pisa/bit_vector.hpp
```diff
@@ -17,13 +17,13 @@ class bit_vector_builder {
   public:
     using bits_type = std::vector<uint64_t>;
 
-    bit_vector_builder(uint64_t size = 0, bool init = 0) : m_size(size)
+    bit_vector_builder(uint64_t size = 0, bool init = false) : m_size(size)
     {
-        m_bits.resize(detail::words_for(size), uint64_t(-init));
-        if (size) {
+        m_bits.resize(detail::words_for(size), uint64_t(-static_cast<int>(init)));
```
amallia: Can we cast to `uint64_t`?

test/test_compact_elias_fano.cpp
```diff
@@ -54,18 +54,18 @@ TEST_CASE_METHOD(sequence_initialization, "compact_elias_fano_construction")
...
         }
 
-        if (!b && rank0 && (rank0 % (1 << of.log_sampling0)) == 0) {
+        if (!b && (rank0 != 0u) && (rank0 % (1 << of.log_sampling0)) == 0) {
             uint64_t ptr_offset =
                 of.pointers0_offset + ((rank0 >> of.log_sampling0) - 1) * of.pointer_size;
             MY_REQUIRE_EQUAL(pos, bv.get_bits(ptr_offset, of.pointer_size), "rank0 = " << rank0);
         }
-        rank += b;
+        rank += static_cast<unsigned long>(b);
```
amallia: `uint64_t` please

test/test_wand_data.cpp
elshize: Good catch, it was automatic fix, and since we were comparing to 0, it was actually casting it to `int` first.

include/pisa/bit_vector.hpp
```diff
@@ -19,7 +19,7 @@ class bit_vector_builder {
 
     bit_vector_builder(uint64_t size = 0, bool init = false) : m_size(size)
     {
-        m_bits.resize(detail::words_for(size), uint64_t(-static_cast<int>(init)));
+        m_bits.resize(detail::words_for(size), init ? std::numeric_limits<std::uint64_t>::max() : 0U);
```
amallia, line 22: why? I believe this is a different behaviour

elshize: Looks correct to me: https://godbolt.org/z/GPEg2g what behavior are you expecting?

amallia: I didn’t realize it was a bool. I thought it was an unsigned int

PR #350, opened 2020-03-27
Extract compression to headers

elshize:


tools/create_freq_index.cpp
```diff
@@ -159,30 +31,29 @@ int main(int argc, char** argv)
...
-            check,                                                  \
-            app.index_encoding(),                                   \
-            app.wand_data_path(),                                   \
-            app.scorer(),                                           \
-            app.quantize());                                        \
+#define LOOP_BODY(R, DATA, T)                                                   \
+    }                                                                           \
+    else if (app.index_encoding() == BOOST_PP_STRINGIZE(T))                     \
+    {                                                                           \
+        pisa::create_collection<pisa::BOOST_PP_CAT(T, _index), wand_raw_index>( \
```
amallia: Can we rename it to `compress_index`?

amallia (changes requested):
I imagine most of the code was just extracted without changes.

amallia (approved):

PR #354, opened 2020-04-03
Regression experiment documentation

amallia:
Added regression experiment for Robust04


docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,80 @@
+# PISA: Regressions for [Disks 4 & 5](https://trec.nist.gov/data_disks.html) (Robust04)
```
elshize: It sounds like we something regressed for Disks 4 & 5. Maybe call it Regression Test(s) or something similar.

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,80 @@
...
+gzip -dc $(find /path/to/disk45/ -type f -name '*.*z' \( -path '*/disk4/fr94/[0-9]*/*' -o -path '*/disk4/ft/ft*' -o -path '*/disk5/fbis/fb*' -o -path '*/disk5/latimes/la*' \)) | bin/parse_collection -f trectext -b 10000 --stemmer porter2 --content-parser html -o robust04/fwd
+```
+
+You can replace `gzip -dc` with `zcat` on Linux or `gzcat` on MacOS.
+The directory `/path/to/disk45/` should be the root directory of [TREC Disks 4 & 5](https://trec.nist.gov/data_disks.html).
+
+### Inverting
+
+```
+bin/invert -i robust04/fwd -o robust04/inv -b 400000 --term-count $(cat robust04/fwd.terms | wc -l)
```
elshize: I think we can ignore the `bin` part, especially since we're not even saying how we build it and where we're running it from, don't you?

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+### Reordering
```
elshize, line 32: Maybe we can say a word or two about what kind of reordering we're doing? Just a sentence that we're reordering document IDs with this and that algorithm.

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+bin/recursive_graph_bisection -c robust04/inv -o robust04/inv.bp --documents robust04/fwd.doclex --reordered-documents robust04/fwd.bp.doclex
+```
+
+### Meta data
```
elshize, line 44: Again, just a short info: we're computing additional data used for bmw algorithm or something.

elshize (changes requested):
A few comments but here's an idea that we've kind of touched on with @JMMackenzie before. How about create a docker image that executes the entire script? You'd have to mount the input directory with the collection, but then all else would just work automatically. This way, we can very quickly run these tests ourselves.

amallia:
> A few comments but here's an idea that we've kind of touched on with @JMMackenzie before. How about create a docker image that executes the entire script? You'd have to mount the input directory with the collection, but then all else would just work automatically. This way, we can very quickly run these tests ourselves.

I am a bit against Docker for this. It is good if we want to automatically keep updated, for example, effectiveness results on several corpuses. I think having a recipe-stile guide is still good in order to document how to use PISA. 

Also, docker is not always great if you want to use it in interactive mode.

elshize:
I am not against the guide, it's good to have it, for sure. But I think docker image also would be nice.

docs/source/experiments/regression-robust04.md
JMMackenzie: Yes agree, could just do `/path/to/pisa/build/bin` or something instead

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Evaluation
+
+Qrels can be downloaded from NIST:
+[TREC 2004 Robust Track (Topics 301-450 & 601-700)](http://trec.nist.gov/data/robust/qrels.robust2004.txt)
+
+
+```
+wget http://trec.nist.gov/data/robust/qrels.robust2004.txt
+trec_eval -m map -m P.30 -m ndcg_cut.20 qrels.robust2004.txt run.robust2004.bm25.title.robust2004.txt
```
JMMackenzie, line 104: We might want to point out that users will require `trec_eval` or we may want to contain it in a test directory?

JMMackenzie (commented):
Looks good to me. I am agnostic on the Docker issue. Could always come back to it if we like?

elshize:
Yeah, we can do (or not) docker later, just saying...

amallia:
@elshize @JMMackenzie I have updated the documentation according to your comments.

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,86 @@
...
+First, we will create a directory where all the indexes are going to be stored:
+
+```
+mkdir robust04
+```
+
+### Parsing
+
+```
+gzip -dc $(find /path/to/disk45/ -type f -name '*.*z' \( -path '*/disk4/fr94/[0-9]*/*' -o -path '*/disk4/ft/ft*' -o -path '*/disk5/fbis/fb*' -o -path '*/disk5/latimes/la*' \)) | bin/parse_collection -f trectext -b 10000 --stemmer porter2 --content-parser html -o robust04/fwd
```
elshize: Can we make these listings more reader-friendly, i.e., break them into reasonable-length lines so that they fit in the text width?

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+You can replace `gzip -dc` with `zcat` on Linux or `gzcat` on MacOS.
+The directory `/path/to/disk45/` should be the root directory of [TREC Disks 4 & 5](https://trec.nist.gov/data_disks.html).
+
+### Inverting
+
+```
+/path/to/pisa/build/bin/invert -i robust04/fwd -o robust04/inv -b 400000 --term-count $(cat robust04/fwd.terms | wc -l)
```
elshize: The `--term-count` is now optional. If missing, it will try to read the term lexicon size, which in this scenario is produced by `parse_collection`, so maybe we should keep it simple and remove it?

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+### Reordering
+We perform Recursive Graph Bisection (aka BP) algorithm, which is currently the state-of-the-art for minimizing the compressed space used by an inverted index (or graph) through document reordering.
```
elshize: Maybe lets add a link to the paper?

docs/source/experiments/regression-robust04.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Reordering
+We perform Recursive Graph Bisection (aka BP) algorithm, which is currently the state-of-the-art for minimizing the compressed space used by an inverted index (or graph) through document reordering. 
+
+```
+/path/to/pisa/build/bin/recursive_graph_bisection -c robust04/inv -o robust04/inv.bp --documents robust04/fwd.doclex --reordered-documents robust04/fwd.bp.doclex
+```
+
+### Meta data
+To perform BM25 queries it is necessary to build an additional file containing the parameters needed to compute the score, such as the document lengths. The file can be built with the following command:
```
elshize: I would say "information" instead of "parameters" and at the end of the sentence add: "such as the document lengths, **max scores, or block max structure.**"

elshize: Maybe let's also change the last sentence to something like: _The following command builds a metadata file with block-max structure with blocks of fixed size of 64 postings:_

elshize (changes requested):

docs/source/experiments/regression-robust04.md
amallia: `max scores, or block max structure` are performance features which are not really needed here since we are measuring only about effectiveness. Also, they are not required by BM25

amallia:
@elshize addressed comments

elshize:
@amallia I added more formatting fixes, take a look if okay.

elshize (approved):

PR #358, opened 2020-04-06
Shard tools

elshize:
This introduces a `shards` tool that can be used to run several tools for shards without a bug-prone bash script.

Before merging, I want to update the docs, so that's missing, but I want to get the feedback first.


JMMackenzie (approved):
I'm a fan. Looks good to me, nice work.

include/pisa/sharding.hpp
```diff
@@ -1,12 +1,12 @@
 #pragma once
 
 #include <algorithm>
+#include <filesystem>
```
JMMackenzie: Hmm, Travis isn't happy though.

```
Scanning dependencies of target invert

[ 53%] Building CXX object tools/CMakeFiles/invert.dir/invert.cpp.o

In file included from /home/travis/build/pisa-engine/pisa/tools/app.hpp:16:0,

                 from /home/travis/build/pisa-engine/pisa/tools/invert.cpp:13:

/home/travis/build/pisa-engine/pisa/include/pisa/sharding.hpp:4:10: fatal error: filesystem: No such file or directory

 #include <filesystem>
```

JMMackenzie (commented):
See comment about Travis build

include/pisa/sharding.hpp
elshize: yeah, I'll switch to boost tomorrow.

amallia (approved):

Issue #339, opened 2020-03-16
Improving the understanding of what PISA is


JMMackenzie:
Given recent feedback from HN, we should look at improving how we explain PISA, and offer some benchmarks to common systems like Lucene and Tantivy (perhaps).

We also should document some things such as:

- Use cases
- Assumptions (in memory)
- Target audience and why you would want to use it
- Limitations
- Algorithms implemented (in terms of the basics, ie top-k search, Boolean matching, etc)
- Scale (some numbers about the collection sizes we can index and search, and some basic timings/sizes)

This is probably something we can build incrementally into the README or an ABOUT page.

fulmicoton:
Hello. I maintain tantivy. 

We have a simple benchmark that makes it possible to measure your engine against tantivy and lucene.

https://github.com/tantivy-search/search-benchmark-game

elshize:
@fulmicoton Thanks! I noticed that many "union" type of queries are now faster with Lucene. Is that due to their recent implementation of BMW? Do you know? Is that a recent change?

fulmicoton:
Yes this is due to BMWand.

You can see them in the bench if you select the top 10 collector and union queries. 
https://tantivy-search.github.io/bench/

Lucene >=8 is >2x faster than tantivy at those queries. 

It was the other way around for Lucene <8.

JMMackenzie:
Thanks to the awesome work of @amallia and @elshize (and the help of @fulmicoton ) we now feature in the Tantivy benchmark game: https://github.com/tantivy-search/search-benchmark-game

This is one step towards closing off this issue :-)

JMMackenzie:
Hi (@elshize @amallia @gustingonzalez)

Just added this: https://github.com/pisa-engine/pisa/blob/describe-pisa/ABOUT.md

Could you take a look and let me know what you think is missing/needs changing?

elshize:
It looks good. Nothing else to add comes to mind at the moment. You wrote: $50$ million, so you make sure to remove the $ signs, but other than that, it looks solid.

JMMackenzie:
> It looks good. Nothing else to add comes to mind at the moment. You wrote: $50$ million, so you make sure to remove the $ signs, but other than that, it looks solid.

Whoops, too much TeX haha. Fixed, thanks.

gustingonzalez:
Hello everyone! Looks good. Maybe, in the first paragraph, it can be useful define the _inverted index_ concept as the logical representation of a corpus.

[...]

PR #359, opened 2020-04-08
Add ABOUT page 

JMMackenzie:
First iteration for #339. We can iterate on this if we like.


elshize:
What do you think of merging "About" into "Readme"? Do we gain anything by having a separate file? I think there's an advantage to having the description show up right away to someone who's visiting the repo, especially for the first time. But it would be nice to have a little table of contents at the top... What do you think?

JMMackenzie:
Totally agree, I was just being conservative with it to be honest (as I didn't want to just nuke the README without consultation). Let's see what Antonio thinks, if he agrees then I will merge it all together.

amallia:
I am okay with merging the ABOUT.md page with the README.md one. 
I think it would be nice to have a usage example of PISA as a library, but we can add that later.

JMMackenzie:
@amallia @elshize okay, take a look now. I have merged the two pages.

amallia (approved):

PR #361, opened 2020-04-16
Refactoring

amallia:
I did some refactoring in order to have single methods that perform the same tasks that the tools currently perform. 
This is useful for me in order to have python bindings.


tools/parse_collection.cpp
```diff
@@ -1,182 +1,16 @@
...
 #include <fmt/format.h>
 #include <spdlog/spdlog.h>
 #include <tbb/task_scheduler_init.h>
-#include <trecpp/trecpp.hpp>
-#include <wapopp/wapopp.hpp>
-#include <warcpp/warcpp.hpp>
 
 #include "forward_index_builder.hpp"
+#include "query/term_processor.hpp"
+#include "record_parser.hpp"
```
JMMackenzie: Should this be `parser.hpp` or am I missing something here?

tools/compress.cpp
```diff
@@ -9,22 +9,28 @@
...
+#include "compress.hpp"
 #include "index_types.hpp"
 #include "util/index_build_utils.hpp"
 #include "util/util.hpp"
 #include "wand_data.hpp"
 #include "wand_data_raw.hpp"
 
-#include "CLI/CLI.hpp"
-
 int main(int argc, char** argv)
```
JMMackenzie: Could we possible name this file/tool `compress_inverted_index` or maybe `compress_index` perhaps? I feel like `compress` is not quite verbose enough... Though maybe it doesn't matter. @elshize, any preference?

include/pisa/query/term_processor.hpp
```diff
@@ -16,6 +18,30 @@ namespace pisa {
 
 using term_id_type = uint32_t;
 
+std::function<std::string(std::string)> get_term_processor(std::optional<std::string> const& type)
```
JMMackenzie: Some versions of GCC are complaining about multiple definitions of this, could you check Travis?

JMMackenzie (changes requested):

tools/compress.cpp
elshize: Either works for me.

JMMackenzie: > Either works for me.

You mean any of the three options? :-)

elshize: I'd lean towards adding at least "index". Though I don't think it matters much. What I'd like to suggest is to start using hyphens instead of underscores in the command names, such as `compress-index`. That's usually how it goes, and it would be nice to follow the standard way.

tools/parse_collection.cpp
amallia: Thanks!

tools/compress.cpp
amallia: Agreed with `compress_inverted_index`

include/pisa/query/term_processor.hpp
amallia: Will fix

elshize (approved):

PR #363, opened 2020-04-20
Unify document reordering in one tool

elshize:
This is ready for review, but before merging, I want to update the docs.


tools/reorder_docids.hpp
```diff
@@ -0,0 +1,319 @@
...
+        input_collection,
+        input_sizes,
+        output_basename,
+        document_lexicon,
+        reordered_document_lexicon,
+        mapping);
+    return 0;
+}
+
+auto reorder_url(
```
JMMackenzie: This is almost more generic than URL sorting right? You could pass a file containing a string for each document, where the string is the target sort key, and this would still work. Correct? Perhaps we should modify it to be more generic than URL?

JMMackenzie: Of course, users could just do this with the pre-defined mapping component as well.

JMMackenzie (approved):
Looks good, left a minor comment. Nice job!

tools/reorder_docids.hpp
```diff
@@ -0,0 +1,319 @@
...
+    std::optional<std::string> const& document_lexicon,
+    std::optional<std::string> const& reordered_document_lexicon,
+    std::uint64_t seed) -> int
+{
+    spdlog::info("Computing random permutation");
+    binary_freq_collection input_collection(input_basename.c_str());
+    auto num_docs = input_collection.num_docs();
+    std::vector<std::uint32_t> mapping(num_docs);
+    std::iota(mapping.begin(), mapping.end(), uint32_t());
+    std::shuffle(mapping.begin(), mapping.end(), std::mt19937{seed});
```
JMMackenzie: Travis for Clang/XCode:
```
[ 54%] Building CXX object tools/CMakeFiles/reorder-docids.dir/reorder_docids.cpp.o

In file included from /Users/travis/build/pisa-engine/pisa/tools/shards.cpp:15:

/Users/travis/build/pisa-engine/pisa/tools/reorder_docids.hpp:118:63: error: non-constant-expression cannot be narrowed from type 'std::uint64_t' (aka 'unsigned long long') to 'std::__1::mersenne_twister_engine<unsigned int, 32, 624, 397, 31, 2567483615, 11, 4294967295, 7, 2636928640, 15, 4022730752, 18, 1812433253>::result_type' (aka 'unsigned int') in initializer list [-Wc++11-narrowing]

    std::shuffle(mapping.begin(), mapping.end(), std::mt19937{seed});

                                                              ^~~~

/Users/travis/build/pisa-engine/pisa/tools/reorder_docids.hpp:118:63: note: insert an explicit cast to silence this issue

    std::shuffle(mapping.begin(), mapping.end(), std::mt19937{seed});

                                                              ^~~~

                                                              static_cast<result_type>( )

1 error generated.
```

elshize: Yeah, for some reason, `std::mt19937` takes `unsigned int`, which is stupid but whatever. I already fixed it but I know Antonio's got some comments so I'm waiting for that to push.

tools/reorder_docids.hpp
elshize: True, this one and the one with defined mapping is very similar, only this is really a reverse mapping really (or the other one is, depending on the way you look at it). So really, these two could co-exist but maybe it would be confusing which is which? Not sure. I wanted to have an option "by URL" so that it's easier to think about, but the problem currently is that you still have to pass that url file.

Bottom line, it's not a bad idea to make it as generic as it can be. If there's an easier way to just do url without passing a file, then we can do it then. For now, though, I we would need to think hard about naming these options to make it clear what it is. Maybe `--with-mapping` vs. `--by-label`. Something descriptive. Any suggestions?

tools/reorder_docids.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+#include "app.hpp"
+#include "reorder_docids.hpp"
+
+int main(int argc, const char** argv)
+{
+    CLI::App app{"Reassigns the document IDs."};
+    pisa::ReorderDocuments args(&app);
+    CLI11_PARSE(app, argc, argv);
+    pisa::reorder_docids(args);
```
amallia, line 11: cam we expand the arguments here?

```
pisa::reorder_docids(args.input_basename(),
                *args.output_basename(),
                args.document_lexicon(),
                args.reordered_document_lexicon(),
                args.seed()
                ...)
```

tools/reorder_docids.hpp
```diff
@@ -0,0 +1,319 @@
+#pragma once
```
amallia, line 1: Can we have this in the library?

tools/recursive_graph_bisection.hpp
```diff
@@ -50,7 +50,7 @@ inline void run_default_tree(size_t depth, const range_type& initial_range)
     recursive_graph_bisection(initial_range, depth, depth - 6, bp_progress);
 }
 
-[[nodiscard]] auto run(RecursiveGraphBisectionArgs const& args) -> int
+[[nodiscard]] inline auto run(RecursiveGraphBisectionArgs const& args) -> int
```
amallia: Same here, can we move this to the library?

amallia (changes requested):

tools/reorder_docids.cpp
elshize: There is a reason why it's not expanded. Mainly, it's because it's used in two binaries, which means you have to expand twice, and it's easy to make a mistake and make these two places diverge.

I suggest moving the specific functions like `reorder_bp` to the library, but keep that one function in `tools` so that it can be reused between `reorder_docids` and `shards`.

amallia: I agree. Thanks

amallia (changes requested):

tools/reorder_docids.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "pisa/reorder_docids.hpp"
+
+namespace pisa {
+
+auto reorder_docids(ReorderDocuments args) -> int
+{
+    try {
+        if (args.bp()) {
+            return recursive_graph_bisection(RecursiveGraphBisectionOptions{
+                .input_basename = args.input_basename(),
```
JMMackenzie, line 13: I don't think some of our compilers like ths format, could instead use `input_basename : args.input_basename(),` I think?

JMMackenzie (changes requested):
Looks nice, left a comment about some of our tools (Travis). PS, do you still want to update the docs?

tools/reorder_docids.hpp
elshize: Nah, I messed up the order + missed one member, just pushed a fix, it should work fine (let's hope).

elshize:
@JMMackenzie Yeah, I should update the docs before merging.

tools/reorder_docids.hpp
elshize: Yep, fixed it.

amallia (approved):
Great PR! Thanks Michal

elshize:
@JMMackenzie is it okay to merge?

JMMackenzie (approved):

PR #364, opened 2020-04-21
Added stem_queries tool

amallia:


tools/stem_queries.cpp
```diff
@@ -0,0 +1,43 @@
...
+    app.add_option("-o,--output", output_filename, "Query output file")->required();
+    app.add_option("--stemmer", stemmer, "Stemmer")->required();
+
+    CLI11_PARSE(app, argc, argv);
+
+    std::ofstream output_file;
+    output_file.open(output_filename);
+
+    auto input_file = std::ifstream(input_filename);
+    pisa::io::for_each_line(input_file, [&](std::string const& query_string) {
```
elshize: Can we move this lambda somewhere to the lib and write a test? I was thinking of maybe something like:
```cpp
try {
    LineStemmer line_stemmer(stemmer);
    pisa::io::for_each_line(input_file, [&](auto const& line) { line_stemmer.stem_line(line); });
} catch (whatever LineStemmer constructor throws) {
    spdlog::error(...);
    std::exit(1);
}
```
Or at least something like this. I'm not married to the names either.

PR #366, opened 2020-04-24
Implement CLI scorer params

JMMackenzie:
Adds command line options for scorers.


include/pisa/compress.hpp
```diff
@@ -62,7 +62,7 @@ void compress_index(
     bool check,
     std::string const& seq_type,
     std::optional<std::string> const& wand_data_filename,
-    std::optional<std::string> const& scorer_name,
+    std::optional<ScorerParams> const& scorer_params,
```
JMMackenzie, line 65: Not entirely sure if I can use std::optional here, see next comment.

tools/app.hpp
```diff
@@ -128,34 +129,45 @@ namespace arg {
         {
             auto* wand = app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
             auto* scorer =
-                app->add_option("-s,--scorer", m_scorer, "Query processing algorithm")->needs(wand);
+                app->add_option("-s,--scorer", m_params.m_name, "Query processing algorithm")->needs(wand);
```
JMMackenzie, line 132: This doesn't work as it currently stands, as m_params is of type `std::optional<ScorerParams>` and so `m_name` cannot be accessed. Same for the other arguments.

include/pisa/compress.hpp
```diff
@@ -87,8 +87,8 @@ void compress_index(
 
         std::unique_ptr<index_scorer<WandType>> scorer;
 
-        if (scorer_name) {
-            scorer = scorer::from_name(*scorer_name, wdata);
+        if (quantized) {
```
JMMackenzie, line 90: @elshize Attention here: Instead of passing an `optional` I have opted to pass the ScorerParam struct and use the `quantized` flag to check whether we need a ranker.

include/pisa/scorer/bm25.hpp
```diff
@@ -9,24 +9,25 @@ namespace pisa {
 
 template <typename Wand>
 struct bm25: public index_scorer<Wand> {
-    static constexpr float b = 0.4;
-    static constexpr float k1 = 0.9;
-
     using index_scorer<Wand>::index_scorer;
 
-    static float doc_term_weight(uint64_t freq, float norm_len)
+    bm25(const Wand& wdata, const float b, const float k1)
```
elshize, line 20: I was wondering, maybe we can at least put a link reference to some description of BM25 for explanation of `b` and `k1`. I imagine for someone just coming to see the code, it's pretty confusing.

include/pisa/scorer/pl2.hpp
```diff
@@ -16,13 +16,13 @@ template <typename Wand>
 struct pl2: public index_scorer<Wand> {
     using index_scorer<Wand>::index_scorer;
 
-    static constexpr float c = 1;
+    pl2(const Wand& wdata, const float c) : index_scorer<Wand>(wdata), m_c(c) {}
```
elshize, line 23: Same idea as for BM25 only for `c`.

include/pisa/scorer/scorer.hpp
```diff
@@ -11,26 +11,37 @@
 #include "quantized.hpp"
 #include "spdlog/spdlog.h"
 
+struct ScorerParams {
+    ScorerParams(std::string s_name = "default") : name(s_name) {}
```
elshize: A few comments:
1. There shouldn't be a default, `"default"` doesn't mean anything does it? I imagine you did that so you can initialize it in CLI with something. I suggest just initialize it to `ScoreParams("")` **there** and leave this to require a name.
2. The constructor should be `explicit`.
3. You should move the contents of `s_name`: `name(std::move(s_name))`.
4. Maybe let's expand the name of the argument to `scorer_name`.

include/pisa/scorer/qld.hpp
```diff
@@ -10,23 +10,26 @@ namespace pisa {
 
 template <typename Wand>
 struct qld: public index_scorer<Wand> {
-    static constexpr float mu = 1000;
-
     using index_scorer<Wand>::index_scorer;
 
+    qld(const Wand& wdata, const float mu) : index_scorer<Wand>(wdata), m_mu(mu) {}
```
elshize, line 23: Same point about `mu`. Here, I imagine we could even explain that `mu` is a Dirichlet (is this the name of the smoothing method?) smoothing parameter.

tools/app.hpp
```diff
@@ -128,34 +129,43 @@ namespace arg {
...
     struct Scorer {
         explicit Scorer(CLI::App* app)
         {
-            auto* opt =
-                app->add_option("-s,--scorer", m_scorer, "Query processing algorithm")->required();
+            app->add_option("-s,--scorer", m_params.name, "Query processing algorithm")->required();
+            app->add_option("--bm25-k1", m_params.bm25_k1, "BM25 k1 parameter.");
+            app->add_option("--bm25-b", m_params.bm25_b, "BM25 b parameter.");
+            app->add_option("--pl2-c", m_params.pl2_c, "PL2 c parameter.");
+            app->add_option("--qld-mu", m_params.qld_mu, "QLD mu parameter.");
```
elshize: Since these are the same as the ones below, maybe we can extract it to a template function:
```cpp
template<typename T>
void add_scorer_options(CLI::App* app, T& args) {
    app->add_option("-s,--scorer", args.m_params.name, "Scorer function")->required();
    ...
}
```
and call it with `add_scorer_options(app, *this)`.

Having the same thing in two places is bug-prone. Case in point:
```
app->add_option("-s,--scorer", m_params.name, "Query processing algorithm")->required();
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
Happens in one place but not the other.

elshize (changes requested):

include/pisa/scorer/scorer.hpp
JMMackenzie: Okay, thanks. Why do we need the `std::move` in this case? Is there some 'best practice' here?

include/pisa/scorer/pl2.hpp
JMMackenzie: Added some verbose descriptions, maybe we can clean them up.

tools/app.hpp
JMMackenzie: Thanks, good idea. I am looking at this now and I noticed that if I make this a free function, I can't access the `ScorerParams` object since it's a `private` member inside each struct.

I've opted to make the free function a friend of each struct that requires it. Not entirely sure if this is a good solution,  let me know what you think.

include/pisa/scorer/scorer.hpp
elshize: The short answer is that you avoid unnecessary copy.

Your constructor gets `s_name` as value, so it will be destructed after the constructor goes out of scope. If you don't move, then an additional copy will be made for `name` constructor. `std::move` effectively moves out any allocated memory to an rvalue reference, while leaving `s_name` in undetermined but valid "empty" state.

It's important to note that small strings will be stored on stack because of small string optimization, so chances are it will be byte-copied anyway given our short names, but if that happens, `std::move` won't have any disadvantages over a copy.

Another point is that alternatively, we could simply pass by reference to the constructor:

```cpp
ScorerParams(std::string const& s_name) : name(s_name) {}
```

This will avoid making an additional copy as well. However, in this case, we'll always be making one copy, while pass-by-value can avoid making any if the argument is an rvalue itself (like a temporary):

```cpp
ScorerParams params("bm25");
// or
ScorerParams params(std::move(another_variable));
```

Again, in our case this whole thing is not very important simply because any overhead is negligible here. But as you suggested, this is kind of a best practice pattern when holding member values of non-trivially-copyable types.

Also, I believe clang-tidy will complain :)

include/pisa/scorer/bm25.hpp
```diff
@@ -7,26 +7,34 @@
 #include "index_scorer.hpp"
 namespace pisa {
 
+// Implements the Okapi BM25 model. k1 and b are both free parameters which
+// alter the weight given to different aspects of the calculation.
+// We adopt the defaults recommended by the following resource - A. Trotman,
+// X-F. Jia, and M. Crane: "Towards an Efficient and Effective Search Engine,"
+// in Proceedings of the SIGIR 2012 Workshop on Open Source Information
+// Retrieval (OSIR), 2012.
```
elshize: Let's use triple slash `///` to indicate it's a piece of documentation, and remove the empty line between the comment and the struct.

The reason is that if we ever decide to use a API doc generator, such as doxygen, it will be automatically usable.

include/pisa/scorer/dph.hpp
```diff
@@ -12,12 +12,16 @@
 
 namespace pisa {
 
+// Implements the DPH model. This model is parameter free.
+// See the following resource for further information - G. Amati, E. Ambrosi,
+// M Bianchi, C Gaibisso, and G Gambosi: "FUB, IASI-CNR and University of Tor
+// Vergata at TREC 2007 Blog Track," in Proceedings of the 16th Text REtrieval
+// Conference (TREC), 2007.
+
```
elshize: Same comment as bm25.

include/pisa/scorer/pl2.hpp
```diff
@@ -12,17 +12,22 @@
 
 namespace pisa {
 
+// Implements the PL2 model. c is a free parameter.
+// See the following resource for further information - G. Amati: "Probabalistic
+// models for information retrieval based on divergence from randomness." PhD
+// Thesis, University of Glasgow, 2003.
+
```
elshize: Same comment as bm25.

include/pisa/scorer/qld.hpp
```diff
@@ -8,25 +8,37 @@
...
 
+// Implements the Query Liklihood model with Dirichlet smoothing.
+// This model has a smoothing parameter, mu.
+// See the following resources for further information - J. M. Ponte, and
+// W. B. Croft: "A Language Modeling Approach to Information Retrieval," in
+// Proceedings of SIGIR, 1998.
+// Also see: C. Zhai and J. Lafferty: "A Study of Smoothing Methods for Language
+// Models Applied to Ad Hoc Information Retrieval," in Proceedings of SIGIR,
+// 2001.
+
```
elshize: Same comment as bm25.

elshize (changes requested):
Looks good! I have just that one cosmetic comment about the documentation.

elshize (approved):
Fantastic :tada:

Issue #357, opened 2020-04-06
Always check if path exists


elshize:
We have many places where we just assume that a file exists, say, when we create a `mio::mmap_source`, which then fails with simply `No such file exists` or something similar if the file is missing. It's difficult to debug, and more importantly, it's very confusing to the end user. We should at least have a function `resolve_path(str)` that throws a more informative error message. We can also think of having another function `resolve_path(path, msgfmt, args...)` that would use `fmt` to display message, and some specialized functions that repeat many times, say:
```cpp
auto resolve_term_lexicon_path(std::string_view path) {
    return resolve_path(path, "Missing term lexicon file: {}", path);
}
```
But I'm also open to other suggestions.

JMMackenzie:
+1, sounds like a nice idea.

PR #370, opened 2020-04-25
More robust memory mapping

### LINKED ISSUE #367: Remove pessimizing move
```
In file included from /home/michal/pisa/tools/stem_queries.cpp:6:
/home/michal/pisa/include/pisa/query/query_stemmer.hpp:23:37: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
            stemmed_terms.push_back(std::move(m_stemmer(*term_iter)));
                                    ^
/home/michal/pisa/include/pisa/query/query_stemmer.hpp:23:37: note: remove std::move call here
            stemmed_terms.push_back(std::move(m_stemmer(*term_iter)));
                                    ^~~~~~~~~~                     ~
1 warning generated.
```

elshize:
Addresses, in part, #357.

Fixes #367


include/pisa/compress.hpp
```diff
@@ -74,16 +75,14 @@ void compress_index(
...
         pisa::progress progress("Create index", input.size());
         WandType wdata;
-        mio::mmap_source md;
-        if (wand_data_filename) {
-            std::error_code error;
-            md.map(*wand_data_filename, error);
-            if (error) {
-                spdlog::error("error mapping file: {}, exiting...", error.message());
-                std::abort();
+        auto wdata_source = [&] {
```
amallia: why a lambda here? Do we need to re-use it? 
I guess we could have one `load_wand_file` function and share it everywhere. What do you think?

include/pisa/compress.hpp
```diff
@@ -74,16 +75,14 @@ void compress_index(
...
+            if (wand_data_filename) {
+                auto source = MemorySource::mapped_file(*wand_data_filename);
+                mapper::map(wdata, source.data(), mapper::map_flags::warmup);
+                return source;
             }
-            mapper::map(wdata, md, mapper::map_flags::warmup);
-        }
+            return MemorySource();
```
amallia: Why do we return a `MemorySource`? `wdata_source` is never used.

include/pisa/memory_source.hpp
```diff
@@ -0,0 +1,73 @@
...
+    /// Subspan of memory.
+    ///
+    /// \throws std::out_of_range   if offset + size is out of bounds
+    [[nodiscard]] auto subspan(size_type offset, size_type size = gsl::dynamic_extent) const
+        -> gsl::span<value_type const>;
+
+  private:
+    explicit MemorySource(std::unique_ptr<mio::mmap_source> source) : m_source(std::move(source)) {}
+
+    std::unique_ptr<mio::mmap_source> m_source;
```
amallia: Why a pointer? We could use `mio::mmap_source`.

include/pisa/memory_source.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <memory>
+
+#include <boost/filesystem/path.hpp>
+#include <gsl/span>
+#include <mio/mmap.hpp>
+
+namespace pisa {
+
+/// This is an owning memory source for any byte-based structures.
+class MemorySource {
```
amallia, line 12: I am not against this class. Just for the sake of discussion, what is the additional value when compared to a `mio::mmap_source`?

src/memory_source.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+    return MemorySource::mapped_file(io::resolve_path(file));
+}
+
+auto MemorySource::mapped_file(boost::filesystem::path file) -> MemorySource
+{
+    if (not boost::filesystem::exists(file)) {
+        throw io::NoSuchFile(file.string());
+    }
+    return MemorySource(std::make_unique<mio::mmap_source>(file.string().c_str()));
```
amallia: Do we deal with any errors that could be generated from mapping the file?
Something like:
```
std::error_code error;	                
md.map(*wand_data_filename, error);	                
if (error) {
   spdlog::error("error mapping file: {}, exiting...", error.message());
}
```

include/pisa/query/query_stemmer.hpp
```diff
@@ -20,7 +20,7 @@ class QueryStemmer {
         std::vector<std::string> stemmed_terms;
         TermTokenizer tokenizer(raw_query);
         for (auto term_iter = tokenizer.begin(); term_iter != tokenizer.end(); ++term_iter) {
-            stemmed_terms.push_back(std::move(m_stemmer(*term_iter)));
+            stemmed_terms.push_back(m_stemmer(*term_iter));
```
amallia: Thanks :)

src/memory_source.cpp
elshize: Yes, the constructor throws `std::system_error` on failure: https://github.com/mandreyel/mio#how-to-create-a-mapping. But it's a good point that I fail to print the file name, need to fix that.

include/pisa/memory_source.hpp
elshize: I mistakenly thought I couldn't use value but I can, thanks.

include/pisa/memory_source.hpp
elshize: This is a good question. This is a very light wrapper but has the following advantages:
1. Initial motivation was to have something less error-prone, in particular, something that gives a clear error when trying to map a non-existent file, but also something that fails quickly if the memory is empty.
2. It's implementation agnostic, if we ever decide to use another backend or even provide alternative to loading everything in memory, we can easily do that.
3. Now that I made some changes, I actually separated implementation and only a small header from `mio` is necessary to use it, the rest is in a binary. Same could be done for `binary_collection` later on.

include/pisa/memory_source.hpp
elshize: I kept the pointer to separate implementation from the header.

JMMackenzie (approved):
Aside from Antonio's comments it's nice! Good work.

JMMackenzie:
Worth mentioning that this closes #367 right?

amallia (approved):

elshize:
As mentioned before, I think there's more to be done, I think index/wand data need to be decoupled from the way we handle memory. It's not as much about the mapper as it is about `mappable_vector`. Imo it needs to go. Regardless, I'll write down what I mean in more detail separately, and we can discuss.

elshize:
@amallia any updates on this?

amallia:
> @amallia any updates on this?

I will follow up soon

elshize:
@amallia still waiting for comments to the latest push.

amallia:
Sorry. It took a while but I was finally able to check it again and merge it.

PR #375, opened 2020-04-29
Compile Headers workflow

elshize:


amallia:
Is there a way to specify the compiler and OS?

amallia (approved):

elshize:
Yes, here's an example: https://github.com/onqtam/doctest/blob/master/.github/workflows/main.yml

PR #378, opened 2020-05-01
Use TBB thread local storage

elshize:
It fixes the problem with `thread_local` storage that causes reordering to segfault when used with `shards` tool. Essentially, the problem is that the sizes of static arrays are initialized once per thread and never reset. This one should initialize thread local storage at the beginning of each BP execution.

The problem with the former version is that the sizes of `thread_local` vectors are set for one shard, and later when they are used for another, it runs out of bound. Now TBB will create thread local storage each time it creates a thread.

I tested performance on a smaller index, and it was the same as before.


JMMackenzie (approved):
Nice work, seems sensible. :+1:

include/pisa/recursive_graph_bisection.hpp
```diff
@@ -226,18 +245,31 @@ void swap(document_partition<Iterator>& partition, degree_map_pair& degrees)
...
-    right_degree.clear();
+    bool exists = false;
+    auto& left_degree = thread_local_data.left_degrees.local(exists);
+    if (exists) {
+        left_degree.clear();
+    } else {
+        left_degree.resize(partition.left.term_count());
+    }
+    auto right_degree = thread_local_data.right_degrees.local(exists);
+    if (exists) {
```
amallia: Wouldn't be better to use the same logic as above: `if(not exist)`
Also can we have a function for this logic?

include/pisa/recursive_graph_bisection.hpp
```diff
@@ -150,13 +160,19 @@ void compute_move_gains_caching(
...
-    const single_init_vector<size_t>& to_lex)
+    const single_init_vector<size_t>& to_lex,
+    bp::ThreadLocal& thread_local_data)
 {
     const auto logn1 = log2(from_n);
     const auto logn2 = log2(to_n);
 
-    thread_local single_init_vector<double> gain_cache(from_lex.size());
-    gain_cache.clear();
+    bool exists;
```
amallia: Not initialized.

include/pisa/recursive_graph_bisection.hpp
```diff
@@ -226,18 +245,31 @@ void swap(document_partition<Iterator>& partition, degree_map_pair& degrees)
...
+    document_partition<Iterator>& partition,
+    GainF gain_function,
+    bp::ThreadLocal& thread_local_data,
+    int iterations = 20)
 {
-    thread_local single_init_vector<size_t> left_degree(partition.left.term_count());
-    left_degree.clear();
-    thread_local single_init_vector<size_t> right_degree(partition.right.term_count());
-    right_degree.clear();
+    bool exists = false;
```
amallia: Here you initialize to `false`. let's be consistent

include/pisa/recursive_graph_bisection.hpp
```diff
@@ -150,13 +160,19 @@ void compute_move_gains_caching(
...
 
-    thread_local single_init_vector<double> gain_cache(from_lex.size());
-    gain_cache.clear();
+    bool exists;
+    auto& gain_cache = thread_local_data.gains.local(exists);
+    if (not exists) {
+        gain_cache.resize(from_lex.size());
+    } else {
+        gain_cache.clear();
+    }
```
amallia: if the thread never asked for `gain_cache` it will create it, otherwise it will clear it and reuse it. 

There is one small optimization possible. 
```cpp
    } else {
        if(gain_cache.capacity() != from_lex.size()) {
           gain_cache.resize(from_lex.size());
         }
        gain_cache.clear();
    }
```
Lexicon becomes smaller and smaller, so we could potentially free up some memory.

There is another thing I do not understand. If Thread_1 runs at some point while reordering shard X. Later on we reorder shard Y and Thread_1 runs again. `thread_local_data.gains.local(exists);` will set `exist` to `True` as Thread_1 previously ran on Shard X. Consequentially it will only clear `gain_cache`, but the capacity will not necessarily satisfy the lexicon size of Shard Y. Makes sense?
`

amallia (changes requested):
Left some comments

include/pisa/recursive_graph_bisection.hpp
elshize: First, in your proposed optimization, I don't think you need a check for capacity, `resize` will do that for you. But more importantly: (1) the entire point of thread local storage is to not reallocate, and I'm pretty sure you can't release memory without reallocation; (2) `resize` will actually never change the capacity, from the cppreference:
> Vector capacity is never reduced when resizing to smaller size because that would invalidate all iterators, rather than only the ones that would be invalidated by the equivalent sequence of pop_back() calls.

If we want to release memory, we would have to force it. But in order to avoid too many reallocations, the condition would have to be more like `2 * from_lex.size() < gain_cache.capacity()`.

As far as the thread thing: that was exactly the bug I was fixing. Now, because the thread local data is explicitly created each time the algorithm starts and destroyed when it finishes, no two instances of the algorithm will share any data, so the problem is no more.

amallia: >As far as the thread thing: that was exactly the bug I was fixing. Now, because the thread local data is explicitly created each time the algorithm starts and destroyed when it finishes, no two instances of the algorithm will share any data, so the problem is no more.

This is what I was thinking, but I was not sure that the data was actually released.

elshize:
I realized we have no tests for BP, so I'm gonna push changes along with some tests I'm writing now.

amallia (approved):

PR #384, opened 2020-05-07
Streaming index compression for block index

elshize:
Compressing an inverted index currently happens in memory and data is dumped to disk only when it's finished. This takes a large amount of RAM. For instance, on my server with 32 GB, I am unable to compress Clueweb09B. And there's really no reason for this to be the case.

This PR addresses this problem only to a certain extend. In particular, I introduce `stream_builder` next to the original `builder` for `block_freq_index` only. The `compress_inverted_index` tool only uses it when it recognizes that it's the right type of index, and also does it only if quantization is not requested.


JMMackenzie (approved):
Looks good, nice work!

PR #387, opened 2020-05-18
Fix deprecated TBB calls

### LINKED ISSUE #368: TBB task_scheduler_init has been deprecated
```
 #pragma message("TBB Warning: tbb/task_scheduler_init.h is deprecated. For details, please see Deprecated Features appendix in the TBB reference manual.")
```

I think we need to move to using `global_control` - https://software.intel.com/en-us/node/589744

JMMackenzie:
Fixes #368 by using the `tbb::global_control` class.

It seems that at alternative method would have been to use the `task_arena` but the current approach will work on a global basis (whereas task_arenas can limit threads on a per-arena basis) so this seemed like the best approach for now.

### DIFF
--- a/tools/evaluate_queries.cpp
+++ b/tools/evaluate_queries.cpp
@@ -199,8 +199,8 @@ int main(int argc, const char** argv)
 
     CLI11_PARSE(app, argc, argv);
 
-    tbb::global_control control(tbb::global_control::max_allowed_parallelism, app.threads());
-    spdlog::info("Number of threads: {}", app.threads());
+    tbb::global_control control(tbb::global_control::max_allowed_parallelism, app.threads() + 1);
+    spdlog::info("Number of worker threads: {}", app.threads());
 
     if (run_id.empty()) {
         run_id = "PISA";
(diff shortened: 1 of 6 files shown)


tools/evaluate_queries.cpp
```diff
@@ -199,7 +199,7 @@ int main(int argc, const char** argv)
 
     CLI11_PARSE(app, argc, argv);
 
-    tbb::task_scheduler_init init(app.threads());
+    tbb::global_control control(tbb::global_control::max_allowed_parallelism, app.threads());
```
elshize: Looking at the docs, it seems like the number of threads include the main thread:

> Limit total number of worker threads that can be active in the task scheduler to parameter value - 1

So I'm wondering whether we should increase it to `app.threads() + 1` since our main thread only waits for the others, so effectively you'd get one fewer worker than you'd expect. Does this make sense?

JMMackenzie: Yep agreed. I have also explicitly stated "worker threads" in the printout (which is the passed thread count, ignoring the +1).

elshize (approved):

elshize:
You have a clang-fmt error in one of the files: https://travis-ci.com/github/pisa-engine/pisa/jobs/336410476

PR #388, opened 2020-05-18
Clean up wand requirements

### LINKED ISSUE #38: Tests fail when run in parallel

### LINKED ISSUE #381: WAND data should be required?
Right now, the CLI doesn't require `--wand` for `compute_intersection` but it fails if I don't pass it. But more generally, is any querying have any chance of working without wand data? Does it make sense for it to be optional?

Edit: in fact, if you look at `evaluate_queries`, all algorithms require wand data, so no reason not to require it at CLI. As far as I can tell, only the not ranked algorithms don't require it (in `tools/queries.cpp`). We should definitely require it for any command that won't work otherwise.

JMMackenzie:
Fixes #381 using the template approach as discussed. We can now specify wand data being required or optional on a per-binary basis.

### DIFF
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -39,17 +39,15 @@ namespace arg {
     struct WandData {
         explicit WandData(CLI::App* app)
         {
+            auto* wand = app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
+            app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file")
+                ->needs(wand);
+
             if constexpr (Mode == WandMode::Required) {
-                auto* wand =
-                    app->add_option("-w,--wand", m_wand_data_path, "WAND data filename")->required();
-                app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file")
-                    ->needs(wand);
-            } else {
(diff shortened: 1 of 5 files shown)


tools/app.hpp
```diff
@@ -33,16 +33,30 @@ namespace arg {
...
-                ->needs(wand);
+            if constexpr (Mode == WandMode::Required) {
+                auto* wand =
+                    app->add_option("-w,--wand", m_wand_data_path, "WAND data filename")->required();
+                app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file")
+                    ->needs(wand);
+            } else {
+                auto* wand = app->add_option("-w,--wand", m_wand_data_path, "WAND data filename");
+                app->add_flag("--compressed-wand", m_wand_compressed, "Compressed WAND data file")
+                    ->needs(wand);
```
elshize: We can take inside of `else` out to the top of the function, and then only add
```cpp
if constexpr (Mode == WandMode::Required) {
    wand->required();
}
```
to avoid code duplication.

elshize (changes requested):
One comment, looks good otherwise :+1:

elshize (approved):

PR #395, opened 2020-05-31
Maxscore perf

elshize:
This is a refactored MaxScore algorithm that brings some significant performance improvements.

**Disclaimer**: Results below were obtained from another branch I'm working on, but the code within maxscore itself is identical.

It would be good **if someone could double-checked the results** if it's consistent across different systems.

I wasn't able to figure out which part exactly makes it faster. I think it's a combination of the additional inner loop and more cache friendly cursor loops, these things mostly seemed to affect the performance but if some is up for a bit more investigation, please go ahead. But first I'd like to check if others get the same times.

# Benchmarks

I performed before and after benchmarks for four cases (all SIMDBP-128 on Clueweb09-B with 1k of Trec05 efficiency track queries, k=1,000):
- `MaxScore`: using BM25 live scoring
- `MaxScore-T:` same as above but with `qk3` thresholds (explained below).
- `MaxScore-Q`: same as `MaxScore` but with quantized scores.
- `MaxScore-QT`: same as `MaxScore-T` but with quantized scores.

| Commit     | MaxScore  | MaxScore-T  | MaxScore-Q  | MaxScore-QT |
| ---------- | ---------:| -----------:| -----------:| -----------:|
| master     | 35.95     | 30.38       | 20.66       | 16.86       |
| this PR    | 29.40     | 24.54       | 15.38       | 12.30       |

# Thresholds

The thresholds are safely estimated from precomputed top-k thresholds for single terms, pairs, and triples.

# Edit

I've pushed some additional commits since the initial push. Here's some explanation:
1. I first noticed that the test for range query failed because I moved cursors out of the original vector, forgetting that it can be used after the function finishes.
2. So I reverted to sorting pointers, as was done before.
3. I ran the benchmark, however, and it slowed down (apparently this was also a factor)
4. So to fix it, I tried sorting "in place" (sort of, but the important thing that the function modified the original vector in a function).
5. This turned out to help but it wasn't as fast as originally.
6. What eventually helped was to move out of the cursor, as I did at the beginning, and only at the end swap the contents with the cursor container passed as an argument. I think it might have been due to the fact that the compiler wasn't able to do some optimizations because it couldn't assume anything about the passed reference. Once I moved content out of it, it had all the information about the local vector to generate more optimal code. I didn't verify that, though, so it's just a conjecture.


elshize:
Forgot to mention, this is based on `cursors` branch. I changed the base branch so that changes related to this PR alone are visible clearly but this should be rather merged to `master` once the other is merged as well.

JMMackenzie:
Awesome! 

I'm running with some different settings here: Gov2, BP index, SIMDBP compression, a bunch of random queries from MQT. I am comparing off of master.

Mean Latency
-------------------
k | Master | This |
|:---:|:---:|:---:|
100 | 14.5 | 11.4 | 
1000 | 24.8 | 19.2 |

Median Latency
------------------
k | Master | This |
|:---:|:---:|:---:|
100 | 8.2 | 6.7 | 
1000 | 17.6| 14.1|

95th Percentile Latency
------------------
k | Master | This |
|:---:|:---:|:---:|
100 | 48.0 | 36.9 | 
1000 | 73.7| 56.1 |

elshize:
@JMMackenzie Nice, can you also double-check precision just in case?

JMMackenzie:
> @JMMackenzie Nice, can you also double-check precision just in case?

I checked the runs on the 150 GOV2 topics and found no diff except for inversions of tied documents, no score differences. So it seems to be a safe optimization :-)

include/pisa/query/algorithm/maxscore_query.hpp
```diff
@@ -1,83 +1,140 @@
...
+                return;
             }
+        }
+    }
 
-            cur_doc = next_doc;
+    template <typename Cursors>
+    void operator()(Cursors&& cursors_, uint64_t max_docid)
+    {
+        using cursor_type = typename std::decay_t<Cursors>::value_type;
```
amallia: What for?

include/pisa/query/algorithm/maxscore_query.hpp
```diff
@@ -1,83 +1,140 @@
...
+        if (cursors_.empty()) {
+            return;
         }
+        auto cursors = sorted(cursors_);
+        run_sorted(cursors, max_docid);
+        std::swap(cursors, cursors_);
```
amallia, line 136: Why do we need this? 
This should prevent from calling `operator()` with `cursors_` being  `const &`. Right?

amallia (approved):
Just a couple of comments, but the code looks good to me.

include/pisa/query/algorithm/maxscore_query.hpp
elshize: not needed here, I'll remove it.

include/pisa/query/algorithm/maxscore_query.hpp
elshize: Yes, it will prevent passing `const&` but we can't pass const anyway because we need to use the cursors. For performance reasons I mentioned before, I move out the cursors of the original container. Because I do that, I'm giving it back at the end. Otherwise range queries won't work.

JMMackenzie (approved):

Issue #402, opened 2020-06-02
Decrease number of test queries


elshize:
Some of the most expensive tests to run are the ones performing queries. These tests are pretty slow because they are run in Debug mode. Right now, we have 500 test queries. I think that for the purpose of unit/integration tests, we don't need as many. I think about 100 would be just fine, and it would save us a few minutes on each run.

PR #404, opened 2020-06-03
Reduce test query count

JMMackenzie:
Back off to the first 100 queries for #402


JMMackenzie:
Going to open a new one without the conflicts, forgot to sync first.

PR #406, opened 2020-06-04
Added kth_threshold tool

amallia:
A tool to predict the threshold of a query similarly to what is proposed in:
> Erman Yafay and Ismail Sengor Altingovde. 2019. Caching Scores for Faster Query Processing with Dynamic Pruning in Search Engines. In Proceedings of the 28th ACM International Conference on Information and Knowledge Management (CIKM ’19). Association for Computing Machinery, New York, NY, USA, 2457–2460. DOI:https://doi.org/10.1145/3357384.3358154

### DIFF
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -62,7 +62,7 @@ struct topk_queue {
 
     void set_threshold(Threshold t) noexcept { m_threshold = t; }
 
-    Threshold threshold() noexcept { return m_threshold; }
+    Threshold threshold() const noexcept { return m_threshold; }
 
     void clear() noexcept
     {
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -70,7 +70,9 @@ struct topk_queue {
         m_threshold = 0;
     }
 
-    [[nodiscard]] uint64_t size() const noexcept { return m_k; }
+    [[nodiscard]] size_t capacity() const noexcept { return m_k; }
+
+    [[nodiscard]] size_t size() const noexcept { return m_q.size(); }
 
   private:
     float m_threshold;
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -62,6 +62,8 @@ struct topk_queue {
 
     void set_threshold(Threshold t) noexcept { m_threshold = t; }
 
+    Threshold threshold() noexcept { return m_threshold; }
+
     void clear() noexcept
     {
         m_q.clear();
--- a/tools/kth_threshold.cpp
+++ b/tools/kth_threshold.cpp
@@ -41,10 +41,7 @@ std::set<uint32_t> parse_tuple(std::string const& line, size_t k)
     for (auto&& term_id: term_ids) {
         try {
             term_ids_int.insert(std::stoi(term_id));
-        } catch (const std::invalid_argument& e) {
-            throw std::runtime_error(
-                fmt::format("Cannot convert {} to int in line: {}", term_id, line));
-        } catch (const std::out_of_range& e) {
+        } catch (...) {
             throw std::runtime_error(
                 fmt::format("Cannot convert {} to int in line: {}", term_id, line));
         }
@@ -123,7 +120,7 @@ void kt_thresholds(
             for (size_t j = i + 1; j < terms.size(); ++j) {
                 if (pairs_set.count({terms[i], terms[j]}) > 0) {
                     Query query;
-                    query.terms = {terms[i], terms[j]}
+                    query.terms = {terms[i], terms[j]};
                     wand_q(make_max_scored_cursors(index, wdata, *scorer, query), index.num_docs());
                     threshold = std::max(threshold, topk.size() == k ? topk.threshold() : 0.0F);
                 }
@@ -134,7 +131,7 @@ void kt_thresholds(
                 for (size_t s = j + 1; s < terms.size(); ++s) {
                     if (triples_set.count({terms[i], terms[j], terms[s]}) > 0) {
                         Query query;
-                        query.terms = {terms[i], terms[j], terms[s]}
+                        query.terms = {terms[i], terms[j], terms[s]};
                         wand_q(
                             make_max_scored_cursors(index, wdata, *scorer, query), index.num_docs());
                         threshold = std::max(threshold, topk.size() == k ? topk.threshold() : 0.0F);
--- a/tools/kth_threshold.cpp
+++ b/tools/kth_threshold.cpp
@@ -59,7 +59,9 @@ void kt_thresholds(
     uint64_t k,
     bool quantized,
     std::optional<std::string> pairs_filename,
-    std::optional<std::string> triples_filename)
+    std::optional<std::string> triples_filename,
+    bool all_pairs,
+    bool all_triples)
 {
     IndexType index;
     mio::mmap_source m(index_filename.c_str());
@@ -116,7 +118,7 @@ void kt_thresholds(
         }
         for (size_t i = 0; i < terms.size(); ++i) {
             for (size_t j = i + 1; j < terms.size(); ++j) {
-                if (pairs_set.count({terms[i], terms[j]}) > 0) {
+                if (pairs_set.count({terms[i], terms[j]}) > 0 or all_pairs) {
                     Query query;
                     query.terms = {terms[i], terms[j]};
                     wand_q(make_max_scored_cursors(index, wdata, *scorer, query), index.num_docs());
@@ -127,7 +129,7 @@ void kt_thresholds(
         for (size_t i = 0; i < terms.size(); ++i) {
             for (size_t j = i + 1; j < terms.size(); ++j) {
                 for (size_t s = j + 1; s < terms.size(); ++s) {
-                    if (triples_set.count({terms[i], terms[j], terms[s]}) > 0) {
+                    if (triples_set.count({terms[i], terms[j], terms[s]}) > 0 or all_triples) {
                         Query query;
                         query.terms = {terms[i], terms[j], terms[s]};
                         wand_q(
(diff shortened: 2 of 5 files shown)


tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    mapper::map(wdata, md, mapper::map_flags::warmup);
+
+    using Pair = std::set<uint32_t>;
+    std::unordered_set<Pair, boost::hash<Pair>> pairs_set;
+
+    using Triple = std::set<uint32_t>;
+    std::unordered_set<Triple, boost::hash<Triple>> triples_set;
+
+    std::string t;
```
elshize: Maybe let's name it `line` so it's clear what it's for.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (pairs_filename) {
+        std::ifstream pin(*pairs_filename);
+        while (std::getline(pin, t)) {
+            std::vector<std::string> p;
```
elshize: What is `p`? Is that a pair? Are the elements supposed to be term IDs? maybe let's call it accordingly.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+            boost::algorithm::split(p, t, boost::is_any_of(" \t"));
+            pairs_set.insert({(uint32_t)std::stoi(p[0]), (uint32_t)std::stoi(p[1])});
```
elshize: This will cause headaches sooner or later. I suggest the following:
- Extract this logic to a function and parameterize it with the number of terms expected (such that we can use it below for triples).
- Introduce one exception class that will report invalid format (could be just `std::runtime_error` with appropriate messages).
- Throw if the size of the vector isn't what we expect.
- Catch both `std::invalid_argument` and `std::out_of_range` that `std::stoi` throws, and re-throw same type as above with a good message.
- Make sure in both cases to include what went wrong (type of error) and the line at which it failed, so that a user can find what went wrong easily.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    for (auto const& query: queries) {
+        float threshold = 0;
+
+        auto terms = query.terms;
+        topk_queue topk(k);
+        wand_query wand_q(topk);
+
+        for(auto&& term : terms) {
+            Query q;
```
elshize: How about just `query`?

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+            q.terms.push_back(term);
+            wand_q(make_max_scored_cursors(index, wdata, *scorer, q), index.num_docs());
+            topk.finalize();
+            auto results = topk.topk();
+            topk.clear();
+            float t = 0.0;
+            if (results.size() == k) {
+                t = results.back().first;
+            }
```
elshize: it could be simplified with `results.size() == k ? results.back().first : 0.0`. We can put that directly into `max` function to avoid awkward variable naming.
```cpp
threshold = std::max(threshold, results.size() == k ? results.back().first : 0.0);
```

elshize: Actually, now that I think about it, we should be able to access both the threshold and the actual size of the `topk_queue` (if not, we should add that). Then, we avoid copying and make it much cleaner:
```cpp
...
wand_q(make_max_scored_cursors(index, wdata, *scorer, q), index.num_docs());
threshold = std::max(threshold, topk.size() == k ? topk.threshold() : 0.0);
```

elshize: Same thing for the loops below.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+            threshold = std::max(threshold, t);
+        }
+        for (size_t i = 0; i < terms.size(); ++i) {
+            for (size_t j = i + 1; j < terms.size(); ++j) {
+                if (pairs_set.count({terms[i], terms[j]})) {
```
elshize: We should avoid implicit conversions, I'm surprised this is not rejected by clang-tidy. Let's use either `count(...) > 0` or `find(...) != pairs_set.end()`. I know it's a pain `contains` isn't in c++17 but still.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+                        t = results.back().first;
+                    }
+                    threshold = std::max(threshold, t);
+                }
+            }
+        }
+        for (size_t i = 0; i < terms.size(); ++i) {
+            for (size_t j = i + 1; j < terms.size(); ++j) {
+                for (size_t s = j + 1; s < terms.size(); ++s) {
+                    if (triples_set.count({terms[i], terms[j], terms[s]})) {
```
elshize: Same comment about count.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,211 @@
...
+    size_t k;
+    std::string index_filename;
+    std::string wand_data_filename;
+    bool quantized = false;
+
+    App<arg::Index, arg::WandData<arg::WandMode::Required>, arg::Query<arg::QueryMode::Ranked>, arg::Scorer>
+        app{"A tool for performing threshold estimation using k-th score informations."};
+
+    app.add_option("-p,--pairs", pairs_filename, "Pairs filename");
+    app.add_option("-t,--triples", triples_filename, "Triples filename");
```
elshize: First, it's not clear what these files are for, so let's document.

Second, do I understand correctly that only these passed tuples will be included in the calculation? This is very complex to use if I just want all of them. Can we have a param `-k` and use these as optional filters (and maybe name them as such, say, `--filter-pairs`? For example, I have a bunch of queries and I want thresholds including pairs and triples, then I have to prepare two additional files just to say "use all".

elshize (changes requested):

tools/kth_threshold.cpp
amallia: I have changed it. Please check this is what you mean

tools/kth_threshold.cpp
amallia: Actually `readability-implicit-bool-conversion` is in place. Weird!

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+    mapper::map(wdata, md, mapper::map_flags::warmup);
+
+    using Pair = std::set<uint32_t>;
+    std::unordered_set<Pair, boost::hash<Pair>> pairs_set;
+
+    using Triple = std::set<uint32_t>;
+    std::unordered_set<Triple, boost::hash<Triple>> triples_set;
+
+    auto to_int = [](std::vector<std::string> const& term_ids) {
```
elshize: I don't really see any reason to create lambda over a free function, it doesn't capture anything and makes the main function longer and more difficult to decode.

But more importantly, why not just pass the line and split inside function? It's exactly the same in both cases. The `while` loop should be something like:

```cpp
while (std::getline(pin, line)) {
    pairs_set.insert(parse_tuple(line, 2)); // substitute 2 with 3 for triples
}
```
The reason why I'm passing 2 is that you want to make sure that you got the right number of terms, because if you pass a file that's four terms on each line, then nobody's gonna notice, there will be no error but unexpected behavior.

So below, you want to do:
```cpp
if (term_ids.size() != k) {
    throw std::runtime_error(fmt::format("Wrong number of terms in line: {} (expected {} but found {})", ...));
}
```

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,204 @@
...
+        std::set<uint32_t> term_ids_int;
+        for (auto&& term_id: term_ids) {
+            try {
+                term_ids_int.insert(std::stoi(term_id));
+            } catch (const std::invalid_argument& e) {
+                throw std::runtime_error(fmt::format("Cannot convert {} to int.", term_id));
```
elshize: Let's also report the line:
```cpp
fmt::format("Cannot convert {} to int in line: {}", term_id, line)
```

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,204 @@
...
+            threshold = std::max(threshold, topk.size() == k ? topk.threshold() : 0.0f);
+        }
+        for (size_t i = 0; i < terms.size(); ++i) {
+            for (size_t j = i + 1; j < terms.size(); ++j) {
+                if (pairs_set.count({terms[i], terms[j]}) > 0) {
+                    Query query;
+                    query.terms.push_back(terms[i]);
+                    query.terms.push_back(terms[j]);
+                    wand_q(make_max_scored_cursors(index, wdata, *scorer, query), index.num_docs());
+                    threshold = std::max(threshold, topk.size() == k ? topk.threshold() : 0.0f);
```
elshize: Just double-checking: are you sure that the function `size` does what it is supposed to and doesn't return the capacity instead? I remember something like that in the past but don't remember if was fixed.

elshize (changes requested):

include/pisa/topk_queue.hpp
```diff
@@ -62,13 +62,17 @@ struct topk_queue {
 
     void set_threshold(Threshold t) noexcept { m_threshold = t; }
 
+    Threshold threshold() noexcept { return m_threshold; }
```
elshize: Function should be `const`.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (term_ids.size() != k) {
+        throw std::runtime_error(fmt::format(
+            "Wrong number of terms in line: {} (expected {} but found {})", line, k, term_ids.size()));
+    }
+
+    std::set<uint32_t> term_ids_int;
+    for (auto&& term_id: term_ids) {
+        try {
+            term_ids_int.insert(std::stoi(term_id));
+        } catch (const std::invalid_argument& e) {
```
elshize: Since we're doing the same thing in both and there's no confusion where the potential exception is coming from, maybe we can just do `catch (const std::exception&) {` and  have one catch block.

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+        for (size_t i = 0; i < terms.size(); ++i) {
+            for (size_t j = i + 1; j < terms.size(); ++j) {
+                for (size_t s = j + 1; s < terms.size(); ++s) {
+                    if (triples_set.count({terms[i], terms[j], terms[s]}) > 0) {
+                        Query query;
+                        query.terms.push_back(terms[i]);
+                        query.terms.push_back(terms[j]);
+                        query.terms.push_back(terms[s]);
```
elshize: We can do `query.terms = {terms[i], terms[j], terms[s]}`.

elshize: Same for pair above.

elshize (changes requested):
Just some nitpicks + the outstanding comment about pairs and triples. Rest looks fine.

include/pisa/topk_queue.hpp
```diff
@@ -62,13 +62,17 @@ struct topk_queue {
...
+    Threshold threshold() const noexcept { return m_threshold; }
+
     void clear() noexcept
     {
         m_q.clear();
         m_threshold = 0;
     }
 
-    [[nodiscard]] uint64_t size() const noexcept { return m_k; }
+    [[nodiscard]] size_t capacity() const noexcept { return m_k; }
```
elshize, line 73: :+1: `[[nodiscard]]` :rocket:

tools/kth_threshold.cpp
```diff
@@ -0,0 +1,206 @@
...
+            pairs_set.insert(parse_tuple(line, 2));
+        }
+        spdlog::info("Number of pairs loaded: {}", pairs_set.size());
+    }
+
+    if (triples_filename) {
+        std::ifstream trin(*triples_filename);
+        while (std::getline(trin, line)) {
+            std::vector<std::string> term_ids;
+            boost::algorithm::split(term_ids, line, boost::is_any_of(" \t"));
```
elshize: These aren't used, are they?

elshize (changes requested):

tools/kth_threshold.cpp
amallia: Good catch

elshize (approved):
Nice.

PR #409, opened 2020-06-07
clang-6 & code coverage

elshize:


.github/workflows/ci.yml
```diff
@@ -49,6 +53,10 @@ jobs:
           if [ "${cc}" = "clang-10" ]; then
             sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"
           fi
+          if [ "${cc}" = "clang-6.0" ]; then
+            # Problems compiling with Clang 6 against libstdc++-9
+            sudo apt-get remove libstdc++-9-dev libgcc-9-dev cpp-9
```
amallia, line 58: Would it be enough to update system includes? I am okay with this solution anyway.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: Test
+      shell: bash
+      working-directory: ${{runner.workspace}}/build
+      run: ctest -j 4
+
+    - name: Upload results
+      shell: bash
+      working-directory: ${{runner.workspace}}/build
+      run: bash <(curl -s https://codecov.io/bash);
```
amallia: You are missing the codecov token to submit new results. I will send it over once I find it :)

amallia: CODECOV_TOKEN="e3f86b6b-0770-4816-9b7d-19b0c550229f"

.github/workflows/ci.yml
elshize: I don't know how to do that tbh. clang finds all installations and always seems to select the latest version. I couldn't figure out how to force it to use another.

amallia:
Can we merge now?

elshize:
@amallia can we exclude `external` and `test` directories from the report?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,47 @@
...
+
+
+    - name: Create Build Environment
+      shell: bash
+      run: cmake -E make_directory ${{runner.workspace}}/build
+
+    - name: Configure
+      shell: bash
+      working-directory: ${{runner.workspace}}/build
+      run: cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=Debug -DPISA_BUILD_TOOLS=OFF -DENABLE_COVERAGE=ON ..
```
amallia, line 29: What's `GITHUB_WORKSPACE` here?

elshize: I think this should point to the root git directory, so the same as `..`. I think I used this at some point, and then changed the working directory and used `..` instead. So I think this is unnecessary. I'm actually a bit surprised it doesn't just fail, because it gets two paths but maybe it just ignores one of them.

`GITHUB_WORKSPACE` seems to be the same as `${{runner.workspace}}` but it's set by the `checkout` action. Not sure if there's a case in which it can be different.

amallia:
Please check now @elshize

elshize:
Awesome. Let's marge it, shall we?

PR #412, opened 2020-06-09
Remove optimal_hybrid_index

elshize:
Removes `optimal_hybrid_index` and removes `STXXL` :tada:

@amallia @JMMackenzie I do, however, have a question regarding two things: `mixed_block` and `semiasync_queue` -- are these used anywhere else? If not, should we drop them? I didn't find any other uses but wanted to check with you.


amallia:
`semiasync_queue` is used in https://github.com/pisa-engine/pisa/blob/16a9c3383ae44f333dec39b19e6b42cda064c7dc/include/pisa/sequence_collection.hpp#L7

It would be great if you removed `semiasync_queue` and replaced in `sequence_collection` using TBB instead. It is not easy. 
EDIT: https://github.com/pisa-engine/pisa/commit/1a27771953720eabd142cf6c1f977c73b16a23cb

amallia:
You can also remove: https://github.com/pisa-engine/pisa/blob/04bba3a251a05ea8a03b2d2ce5f546b24b289cbb/tools/profile_decoding.cpp

amallia:
And yes `mixed_block` can be removed.

elshize:
> `semiasync_queue` is used in
> 
> https://github.com/pisa-engine/pisa/blob/16a9c3383ae44f333dec39b19e6b42cda064c7dc/include/pisa/sequence_collection.hpp#L7
> 
> It would be great if you removed `semiasync_queue` and replaced in `sequence_collection` using TBB instead. It is not easy.
> EDIT: [1a27771](https://github.com/pisa-engine/pisa/commit/1a27771953720eabd142cf6c1f977c73b16a23cb)

I guess in that case, I'll leave it in this PR, we can remove it later. What exactly is a point of this queue, does it just serve as a thread pool? Or does it do something special I'm missing?

JMMackenzie (approved):
Looks fine to me, I guess we can remove the other stuff in a separate PR?

PR #414, opened 2020-06-11
Update badge

amallia:


README.md
```diff
@@ -2,9 +2,14 @@
 
 # PISA: Performant Indexes and Search for Academia v0.8.2
 
-[![Build Status](https://travis-ci.com/pisa-engine/pisa.svg?branch=master)](https://travis-ci.com/pisa-engine/pisa)
-[![codecov](https://codecov.io/gh/pisa-engine/pisa/branch/master/graph/badge.svg)](https://codecov.io/gh/pisa-engine/pisa)
+![Build and test](https://github.com/pisa-engine/pisa/workflows/Build%20and%20test/badge.svg)
+![Compile Headers](https://github.com/pisa-engine/pisa/workflows/Compile%20Headers/badge.svg)
```
elshize: I think this one might be confusing for people, I don't think we need this one. I would just remove it and rename "Build and test" to "Tests".

Also I don't see the point to advertise that `clang-format` passed; this is more of an internal test for developers, not really important to the users overall imo.

Now, `clang-tidy`, that makes sense. But I think we can limit badges to Tests, clang-tidy, and coverage. We might want to add a docker badge (not so much for the test run as just with the link to the image on docker hub).

amallia: Ok, but if we do `Tests` are we showing the badge for `Build and test` only? `Compile Headers` is on a separate workflow

elshize: Yes, I think that's fine, the headers one is not that important, I think.

amallia: Removed `Compile Headers` and `clang-format `.

I cannot rename `Build and test` to `Tests` because the badge is generated based on the workflow name.

elshize (approved):

PR #417, opened 2020-06-22
Update progress.hpp

amallia:
Progress is printing a lot of logs and it is very difficult to debug failing tests in Github actions.

### DIFF
--- a/include/pisa/util/progress.hpp
+++ b/include/pisa/util/progress.hpp
@@ -52,7 +52,7 @@ class progress {
         size_t progress = (100 * m_count) / m_goal;
         std::chrono::seconds elapsed = std::chrono::duration_cast<std::chrono::seconds>(
             std::chrono::steady_clock::now() - m_start);
-        if (progress != m_progress and elapsed != m_elapsed) {
+        if (progress != m_progress or elapsed != m_elapsed) {
             m_progress = progress;
             m_elapsed = elapsed;
             std::cerr << '\r' << m_name << ": " << progress << "% [";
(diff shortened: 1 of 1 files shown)


include/pisa/util/progress.hpp
```diff
@@ -39,6 +39,8 @@ class progress {
     std::string m_name;
     size_t m_count = 0;
     size_t m_goal = 0;
+    size_t m_progress = 0;
+    std::chrono::seconds m_elapsed;
```
elshize: Let's do `std::chrono::seconds m_elapsed{};` for consistency and remove the `, m_elapsed()` from the constructor.

include/pisa/util/progress.hpp
```diff
@@ -9,7 +9,7 @@ namespace pisa {
 
 class progress {
   public:
-    progress(const std::string& name, size_t goal) : m_name(name)
+    progress(const std::string& name, size_t goal) : m_name(name), m_elapsed()
     {
         if (goal == 0) {
             throw std::runtime_error("goal must be positive");
```
elshize: As long as we're making changes to this file, can we also update this message to something more informative? Say, "Progress bar must have a positive goal but 0 given".

elshize (changes requested):
This should be good enough solution for CI but in general, we'll still have a problem if run a longer processing and pipe the logs to a file (a line per second for hours will be too much, maybe printing a line per percent point or something, I'm not sure).

The usual solution to that is to detect whether we're speaking to the TTY or a file. Here's an example for POSIX systems (we'd have to replace `stdin` with `stdout`):

https://stackoverflow.com/a/7601564

We should define somewhere a `is_tty` function that will implement this for all platforms, and use that in our code.

Feel free to defer it to a separate PR but if so, make sure to open an issue.

elshize (approved):

PR #424, opened 2020-07-13
Fix BP memory issue

### LINKED ISSUE #423: BP Reordering using large amounts of memory
- BP seems to be using much more memory (up to 4x more from my calculations) than it did before, and can even end up crashing if the depth is too deep or if the collection is large. 

- My working hypothesis is that the introduction of the `thread_local` storage (see https://github.com/pisa-engine/pisa/commit/2c459738fc23393c5f5009dc5b71e4b57d4f058d#diff-51da00fc61477e7d898e77928ab18333) is causing the algorithm to duplicate memory as it recurses (as it should), but the memory of the partition it recursed from remains in-use. 

- If my hypothesis is correct, I think we should be able to free the memory of the 'parent' partition once we recurse? So each time a partition is "done", we split and call two new threads which build their local storage and set up, and then we free the caller's memory.

- Related to #422

- I admit that this requires further investigation. However, correctness still seems good, as I can reproduce earlier results with similar indexes (eg, a CIFF gov2 index gives similar BPI to the ECIR paper).

elshize:
This most likely will fix #423 It should both reduce the memory consumption and speed it up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,12 @@
 
 # PISA: Performant Indexes and Search for Academia v0.8.2
 
-[![Build Status](https://travis-ci.com/pisa-engine/pisa.svg?branch=master)](https://travis-ci.com/pisa-engine/pisa)
-[![codecov](https://codecov.io/gh/pisa-engine/pisa/branch/master/graph/badge.svg)](https://codecov.io/gh/pisa-engine/pisa)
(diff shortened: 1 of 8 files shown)


JMMackenzie (approved):
Looks good to me.

PR #429, opened 2020-09-08
Fix threshold logic

### LINKED ISSUE #415: Heap condition slows down queries on quantized index
Because `would_enter` method of the top-k heap checks if the score is **greater or equal**, it means that sometimes a document will be entered into the heap just to be removed. This rarely happens when scores are decimals, but it's much more frequent when the scores are quantized. It especially affects VBMW algorithm, and can slow it down in some cases such that a query on the quantized index is several times slower than one on the non-quantized index.

JMMackenzie:
Fixes #415

- Heap threshold check is less liberal
- Thresholds program will remove an epsilon from the threshold before outputting it.


amallia:
We need to fix all the tools first

amallia:
Sorry, I mean instead of outputting a threshold lower than the k-th score, I prefer to remove the epsilon at query time.

elshize:
If we follow Antonio's definition of threshold "the score of the k-th document", I think then we stick to it everywhere, and ideally hide this logic in the heap implementation. Would that work? If so, we should then make sure to use the public definition of threshold in the public interface of the class. This could be actually nice, because all that trickery would be contained.

One thing to test would be if decreasing the quantized threshold with `std::nextafter` would work properly? I think it should be just fine unless I'm missing something... Should test it though for sure.

JMMackenzie:
I agree with you both, but I thought we'd decided in #415 that the threshold should be independent of the k-th score?

If we want to hide this functionality as you're suggesting, then topk_queue will need two threshold values:

1) `heap_entry_threshold`: The value a document needs to exceed to get into the heap
2) `score_threshold`: The value a document needs to exceed to be considered for scoring

The `score_threshold` is important because it can be set by any tool (and as originally discussed, can come from anywhere, not necessarily the PISA tools). Note that `heap_entry_threshold` may initially be lower than `score_threshold`, but they should eventually converge to the same value (and indeed, as the `heap_entry_threshold` rises above `score_threshold`, `score_threshold` must be updated at the same time, otherwise we will score more documents than necessary). There is also the case where the (predicted) `score_threshold` is larger than the final value of `heap_entry_threshold` - in this case, we know the prediction was wrong, and need to recompute the answer set (similar to what is currently done now with the `--safe` flag).

The reason I pushed this PR was that I think it's the most simple technique which satisfies our requirements, but I am happy to revisit this original discussion if need be.

elshize:
Let me think this over tomorrow morning and come back to you.

The reason I want to get it right is that I really want to avoid confusion by having thresholds as different things in different places. If we have it in `queries` than it's fine now, but what if someone writes their own program to query using the heap, and then they need to know that this threshold is different from the other. It just is inconsistent and bug prone.

I'll look into it tomorrow, including what you just commented and try to see if I have an idea how to approach it so it's both clear and efficient.

JMMackenzie:
After a discussion with @elshize we have worked out a new potential approach which I have just drafted.

1) Inside the heap, when passing in a predicted threshold, we store this threshold (as `m_predicted_threshold`) and we set `m_threshold` to the predicted value minus epsilon.

2) Now, we are able to check correctness of processing by establishing whether `m_threshold == m_predicted_threshold` after traversal. If this is not true, it means that either A) there isn't k documents in the heap or B) we overpredicted. 

This also addresses the current issue of having `>=` on the heap entry check, and covers the case where we have a quantized score and perfectly predict the threshold.

NOTE: This is just a first take for discussion. I may have missed some other functions etc, and we may want to change the function names and so on. But we should discuss this approach.

include/pisa/topk_queue.hpp
```diff
@@ -10,7 +10,7 @@ using Threshold = float;
 struct topk_queue {
     using entry_type = std::pair<float, uint64_t>;
```
elshize: This is not related to the issue but it just occurred to me: there is no reason for the entry to hold `uint64_t`, is there? Notice that because of this, there will be some padding after the float and the entire entry is now 16 bytes when it could be just 8 (we use 4 byte integers for documents elsewhere anyway, so we don't support any bigger anyway). This is of course a relatively small structure but I'm thinking that cutting the size in half could maybe improve caching. And I see no downsides, either. So I suggest changing it to `std::pair<float, uint32_t>`.

elshize: Then, we should change the type of `k` as well for good measure, to be consistent.

include/pisa/topk_queue.hpp
```diff
@@ -60,14 +60,21 @@ struct topk_queue {
 
     [[nodiscard]] std::vector<entry_type> const& topk() const noexcept { return m_q; }
 
-    void set_threshold(Threshold t) noexcept { m_threshold = t; }
+    void set_threshold(Threshold t) noexcept
```
elshize: Name and docs suggestion:
```cpp
/// Override the threshold to be `threshold`.
///
/// This is meant to be used at the beginning of query processing if a threshold is available beforehand,
/// e.g., by means of estimation. Note that if `threshold` is higher than it would be at the end of query processing,
/// it could result in incomplete document list.
void force_threshold(Threshold threshold) noexcept
```

include/pisa/topk_queue.hpp
```diff
@@ -60,14 +60,21 @@ struct topk_queue {
...
+    {
+        m_threshold = std::max(std::nextafter(t, 0.0), 0.0);
+        m_predicted_threshold = t;
+    }
 
     Threshold threshold() const noexcept { return m_threshold; }
```
elshize:
```cpp
/// Returns the threshold based on the heap state. The threshold is defined as the score of the `k`-th document
/// or 0.0 if the heap is not full.
Threshold threshold() const noexcept { return capacity() == size() ? m_q.front().first : 0.0; }
```

include/pisa/topk_queue.hpp
```diff
@@ -60,14 +60,21 @@ struct topk_queue {
...
 
+    bool is_correct() noexcept { return (m_threshold == m_predicted_threshold); }
```
elshize: Unnecessary parens and:
```cpp
/// Returns `true` if no documents have been missed up to this point.
/// The reason why document could be missed is forcing a threshold that is too high (overestimated).
bool complete_results() noexcept { return m_threshold == m_predicted_threshold; }
```

include/pisa/topk_queue.hpp
```diff
@@ -76,6 +83,7 @@ struct topk_queue {
 
   private:
     float m_threshold;
+    float m_predicted_threshold;
```
elshize: I'd use initializer here instead of constructor:
```cpp
    float m_threshold = 0;
    float m_predicted_threshold = 0;
```

elshize (commented):
I added some comments and suggestions for naming and docs, I'm open to ideas though.

amallia:
> After a discussion with @elshize we have worked out a new potential approach which I have just drafted.
> 
> 1. Inside the heap, when passing in a predicted threshold, we store this threshold (as `m_predicted_threshold`) and we set `m_threshold` to the predicted value minus epsilon.
> 2. Now, we are able to check correctness of processing by establishing whether `m_threshold == m_predicted_threshold` after traversal. If this is not true, it means that either A) there isn't k documents in the heap or B) we overpredicted.
> 
> This also addresses the current issue of having `>=` on the heap entry check, and covers the case where we have a quantized score and perfectly predict the threshold.
> 
> NOTE: This is just a first take for discussion. I may have missed some other functions etc, and we may want to change the function names and so on. But we should discuss this approach.

I agree with this solution. I think it is the best option we have right now!

include/pisa/topk_queue.hpp
amallia: Good point. We should have a type definition of DocID somewhere and use always that one.
We can do it in a separate PR if you prefer.

include/pisa/topk_queue.hpp
amallia: Unrelated to this PR, but just a curiosity. Can we use `complete_results` to early terminate? :)

include/pisa/topk_queue.hpp
```diff
@@ -10,7 +10,7 @@ using Threshold = float;
 struct topk_queue {
     using entry_type = std::pair<float, uint64_t>;
 
-    explicit topk_queue(uint64_t k) : m_threshold(0), m_k(k) { m_q.reserve(m_k + 1); }
+    explicit topk_queue(uint64_t k) : m_threshold(0), m_predicted_threshold(0), m_k(k) { m_q.reserve(m_k + 1); }
```
amallia: Minor: 
`topk_queue(uint64_t k, Threshold t)` can we have a constructor where we pass an initial threshold so we dont need to set it later?

include/pisa/topk_queue.hpp
amallia: Instead of `m_predicted_threshold` can we call it `m_initial_threshold`?
It might not be "predicted" and I guess it can generate some confusion. The name `initial` abstracts out the way it was obtained, instead it indicates the heap is initialized to a value. 

EDIT: maybe also `initial` is confusing as it may indicate that it will change. Maybe `m_initialized_threshold` or `m_expected_threshold`. Idk, suggestions for better naming?

amallia (changes requested):
Minor changes requested, but everything else looks good to me.

include/pisa/topk_queue.hpp
elshize: We might as well change it to `uint32_t` here, but since the alias will touch a lot of files, maybe that part should be done separately.

include/pisa/topk_queue.hpp
elshize: I don't think so. Now that I'm thinking about it, maybe we should call it `is_safe`.

BTW, shouldn't it be `m_threshold >= m_predicted_threshold`? Or am I missing something?

include/pisa/topk_queue.hpp
elshize: Aggreed.

include/pisa/topk_queue.hpp
elshize: I think I like "initialized". I think also we should document these things in the code to make it clear.

include/pisa/topk_queue.hpp
JMMackenzie: Agreed, but yes, let's do this in a separate PR.

include/pisa/topk_queue.hpp
JMMackenzie: You're right, it should be `>=` - I will fix that.

include/pisa/topk_queue.hpp
```diff
@@ -10,7 +10,8 @@ using Threshold = float;
 struct topk_queue {
     using entry_type = std::pair<float, uint64_t>;
 
-    explicit topk_queue(uint64_t k) : m_threshold(0), m_k(k) { m_q.reserve(m_k + 1); }
+    explicit topk_queue(uint64_t k) : m_threshold(0), m_initialized_threshold(0), m_k(k) { m_q.reserve(m_k + 1); }
+    explicit topk_queue(uint64_t k, Threshold t) : m_threshold(0), m_initialized_threshold(t), m_k(k) { m_q.reserve(m_k + 1); }
```
elshize: We can now remove the zero-initializers (`m_threshold` in both and `m_initialized_threshold` in the first constructor) now that we set it below.

elshize (changes requested):

include/pisa/topk_queue.hpp
JMMackenzie: Yes, I just noticed this - fixing a couple of other issues too. Thanks!

JMMackenzie:
I've just made a WIP push to reflect some of the changes discussed.

Might be nice to document some of the other things going on too... 

In any case, I *think* this *should* work now, but we can implement some tests and maybe run a regression vs master to double check.

elshize:
We do have some regression tests for query processing. But since this is a crucial part of the system, it would be a good idea to have some unit tests for the queue itself; maybe even some property tests with rapidcheck. For example, generating some postings, then sorting, taking the k-th score, then shuffling, and running: (1) without initial threshold, (2) with lower threshold, (3) with exact threshold, and (4) higher threshold. Then we check some properties, like whether `is_safe` is correct, etc.

JMMackenzie:
We can merge this one, but we'll need to make a new issue for the unit tests etc.

elshize:
Hey, here's something I realized. First, minor detail, `std::max(std::nextafter(t, 0.0), 0.0)` is equivalent to `std::nextafter(t, 0.0)` as per https://en.cppreference.com/w/cpp/numeric/math/nextafter

> 1-3) If from equals to, to is returned.

But more importantly, I noticed that even if we perform that operation, it doesn't necessarily ensure correctness. If the threshold is calculated with a different algorithm and arithmetic operations are performed in a different order, the lowered threshold could be within float error. I'm not sure how much we want to care about that for floats, but I'm concerned that in some rare cases it could affect quantized scores in not insignificant way. (That's why I think we should test this one thoroughly; in particular, we should test that algorithms work correctly with quantized threshold equal to the real threshold.)

The simplest solution I can think of is just calling `nextafter` a few times to be on the safe side, say, 10 times. This should not affect performance, the difference will still be very small, but we might avoid any precision issues.

Any thoughts?

JMMackenzie:
Yeah that's a good point.

I wonder if we could analytically determine the smallest legitimate difference between two document scores and then build that into some epsilon value that we remove? I agree that calling nextafter a few times would be one approach but possibly confusing, I think we'd need to document it carefully.

I agree though that the quantisation case is more interesting (difficult) so we should definitely get some thorough tests in for that.

elshize:
I haven't spent much time thinking this through, but I think the epsilon depends on the number of query terms. The scoring formula for a term should be executed the same way no mater what (right?) so I think it's only the order of summing scores could change something, so the precision error should be bounded by the equivalent of executing `nextafter` `term_count` times. I'm not sure if that's sound logic, though. Also, the heap doesn't know the number of terms, so that's something to think about.

amallia:
What is the status of this PR? @elshize @JMMackenzie 
I have the feeling we would like to merge it considering the impact it has on quantized scores (DeepImpact for example).

JMMackenzie:
Yes agreed. At very worst case we should apply:

`bool would_enter(float score) const { return score > m_threshold; }` instead of the `>=` that's there now.

What do you think @elshize @amallia ?

amallia:
Yes, I would go with 
```
bool would_enter(float score) const { return score > m_threshold; }
```

elshize:
Agreed.

elshize (approved):

elshize:
I've done some small changes to the queue (mostly minor) and added test + comments for public functions. Let me know what you guys think of this.

One thing I did is to remove `set_threshold` altogether, and instead, you gotta set it at construction, or with `clear` -- so that you can't manually set it in the middle of the algorithm.

Good thing I wrote tests, because the `nextafter` we were calling was actually incorrect: because 0.0 was interpreted as `double`, then the result was `double`, and then immediately turned to `float` again, so it ended up being the same value. That would have totally brake quantized. It checks out: I remember when I was hacking it for the paper, I ended up subtracting 1 for quantized instead, because nextafter wouldn't work -- now I know why!

Gotta love c++ and its implicit type casting...

include/pisa/topk_queue.hpp
```diff
@@ -34,18 +51,21 @@ struct topk_queue {
         if (PISA_UNLIKELY(m_q.size() <= m_k)) {
             std::push_heap(m_q.begin(), m_q.end(), min_heap_order);
             if (PISA_UNLIKELY(m_q.size() == m_k)) {
-                m_threshold = m_q.front().first;
+                m_effective_threshold = m_q.front().first;
```
JMMackenzie, line 55: Wouldn't this create slowdowns where a good initial threshold is provided?

As in, assume the `initial_threshold = 10` and then once the heap has *k* elements the *k* th score is `5.0` - this is probably not what is desired.

include/pisa/topk_queue.hpp
```diff
@@ -34,18 +51,21 @@ struct topk_queue {
...
             }
         } else {
             std::pop_heap(m_q.begin(), m_q.end(), min_heap_order);
             m_q.pop_back();
-            m_threshold = m_q.front().first;
+            m_effective_threshold = m_q.front().first;
```
JMMackenzie, line 60: And same issue here, right?

include/pisa/topk_queue.hpp
```diff
@@ -34,18 +51,21 @@ struct topk_queue {
...
         }
         return true;
     }
 
-    bool would_enter(float score) const { return score >= m_threshold; }
+    /// Checks if an entry with the given score would be inserted to the queue, according
+    /// to the current threshold.
+    bool would_enter(float score) const { return score > m_effective_threshold; }
```
JMMackenzie, line 67: I think this is fine depending on how `m_effective_threshold` is handled, see comments above...

include/pisa/topk_queue.hpp
```diff
@@ -58,26 +78,52 @@ struct topk_queue {
...
+    Threshold true_threshold() const noexcept
+    {
+        return capacity() == size() ? m_q.front().first : 0.0;
+    }
+
+    /// Returns the threshold set at the start (by default 0.0).
+    Threshold initial_threshold() const noexcept { return m_initial_threshold; }
+
+    /// Returns the maximum of `true_threshold()` and `initial_threshold()`.
+    Threshold effective_threshold() const noexcept { return m_effective_threshold; }
```
JMMackenzie: Again, I think this might be OK, or we might actually want to be using the max here depending on how the previous cases are handled.

JMMackenzie (commented):
I think it's generally pretty good but I believe I have identified a corner-case which needs to be addressed. I'll wait for confirmation first, though.

include/pisa/topk_queue.hpp
elshize: I'm not sure I understand. Do you mean that the threshold would decrease? I don't think it can, because only higher scores would go past `would_enter` at the top of the function, right? Or am I missing something?

JMMackenzie: Ah, yes, I think you are right; we can only get a change in `m_effective_threshold` once k elements enter, but those elements can only enter if they exceed that value in the first place. My mistake!

But hey, at least we have a test for that now (thanks!)

JMMackenzie:
I think we're good to go? @elshize @amallia - let's merge?

elshize:
> But hey, at least we have a test for that now (thanks!)

Yep, I actually think these ones are very good to have.

> I think we're good to go?

Yes, ok with me.

PR #430, opened 2020-09-17
Taily stats

elshize:
This PR adds a set of tools related to [Taily](https://github.com/pisa-engine/taily):
- [x] `taily-stats` tool for storing term statistics necessary for Taily computation (including a subcommand for `shards`)
- [x] `ToolStats` class for reading stored statistics from memory (mapped file)
- [x] `shards taily-score` tool for scoring shards for queries passed to the program
- [x] Taily threshold estimation tool
- [x] CLI sanity tests


external/CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/GSL EXCLUDE_FROM_ALL)
 
 # Add Boost
 add_subdirectory(boost-cmake)
+set(BOOST_INCLUDEDIR ${CMAKE_CURRENT_SOURCE_DIR}/boost-cmake CACHE PATH "Change boost lookup path")
+set(BOOST_LIBRARYDIR ${CMAKE_BINARY_DIR}/boost-cmake CACHE PATH "Change boost lookup path")
```
amallia, line 72: Can I ask why we need these two?

amallia (approved):

external/CMakeLists.txt
elshize: Yeah, taily library uses boost but doesn't pull it as submodule, so in order for it to find the one we pull, you need this path. Otherwise, it will either find the system copy (like in my env) or nothing and fail (like on CI).

PR #434, opened 2020-09-24
Fix lexicon issue in term processor and refactor binary search

elshize:
The issue was that the binary search didn't check if the iterator is `end()` which made things crash when a word was being looked up which is larger lexicographically than the last element in the lexicon.

The problem wasn't there in `lexicon` tool but was in the term processor. So I extracted the logic to a common function to avoid a situation like that in the future. Also added some tests.


amallia (approved):
Damn! Makes sense. Thanks

PR #437, opened 2020-09-27
Read collection with lexicon

elshize:
This modifies the `read_collection` tool so that we can pass a lexicon instead of a text file as int to string mapping.


tools/read_collection.cpp
```diff
@@ -4,54 +4,79 @@
...
 
-    CLI::App app{"read_collection - read collections."};
+    CLI::App app{"Reads binary collection to stdout."};
     app.add_option("-c,--collection", collection_file, "Collection file name")->required();
-    app.add_option("-m,--map", map_file, "String map file name");
-    app.add_option("first", first, "Element number")->required();
-    app.add_option("last", last, "Element number");
+
+    auto maptext =
+        app.add_option("--maptext", map_file, "Text file mapping each line number (ID) to a string");
```
JMMackenzie: Maybe we could say "DocId" here just to be a bit more clear?

tools/read_collection.cpp
```diff
@@ -4,54 +4,79 @@
...
+    auto maplex = app.add_option("--maplex", lex_file, "Lexicon file mapping IDs to strings");
+    maptext->excludes(maplex);
+    maplex->excludes(maptext);
+
+    app.add_option("first", first, "First element number")->required();
+    app.add_option(
+        "last",
+        last,
+        "Open-ended end of range of elements to read. "
```
JMMackenzie: Maybe standardize as "Last element number; [First, Last] provides a range of elements to read. If Last is not defined, only one element (First) will be read" -- Or something similar?

JMMackenzie (commented):
Looks good to me, just a few minor wording issues which were slightly confusing IMO.

tools/read_collection.cpp
elshize: Actually, this is not always DocId, as it works with both fwd and inv indexes, and in general with any "binary collection". But I agree, maybe we should be more verbose here:
```
ID to string mapping in text file format. Line n is the string associated with ID n.
E.g., if used to read a document from a forward index, this would be the `.terms`
file, which maps term IDs to their string reperesentations.
```

Same for lexicon:
```
ID to string mapping in lexicon binary file format.
E.g., if used to read a document from a forward index, this would be the `.termlex`
file, which maps term IDs to their string reperesentations.
```

tools/read_collection.cpp
elshize: Noted. Actually, I think I want to change this logic. First, the only reason why it makes sense (and probably why I did it this way) to have range [first, last) instead of [first,last] is if you want to print all, and you say: 0 <len>, which is considerably more convenient for scripting than `<len> - 1`. But instead, I think what's better is to have these options:
1. `n` read `n`-th element.
2. `n m` read between `n` and `m` inclusively.
3. `n -` read from `n` to the end.

The alternative is to have users state `n n` when they want the `n`-th and `n` if they want all.

Yet another solution would be to introduce subcommands: `entry`, `range`, `all` or something like that.

Let me know what you think about any of these.

tools/read_collection.cpp
```diff
@@ -4,54 +4,79 @@
...
     app.add_option("-c,--collection", collection_file, "Collection file name")->required();
-    app.add_option("-m,--map", map_file, "String map file name");
-    app.add_option("first", first, "Element number")->required();
-    app.add_option("last", last, "Element number");
+
+    auto maptext =
+        app.add_option("--maptext", map_file, "Text file mapping each line number (ID) to a string");
+    auto maplex = app.add_option("--maplex", lex_file, "Lexicon file mapping IDs to strings");
+    maptext->excludes(maplex);
+    maplex->excludes(maptext);
```
amallia, line 57: I think having both is redundant, but it does not hurt too.

tools/read_collection.cpp
amallia: I like `entry, range, all` if I have to be honest :)

amallia (approved):

tools/read_collection.cpp
elshize: It's good for when the help message is printed, whichever you look at, you immediately see it excludes the other.

elshize:
I made some changes. I went for `entry` for single entry and `range` for range with optional second argument, where missing value indicates read till the end.

PR #438, opened 2020-09-27
Update documentation

elshize:
Updating documentation to reflect current `master` version.


elshize:
The diagram can be found here (as it won't appear in source or on github preview): https://github.com/pisa-engine/pisa/blob/docs-update/docs/source/_static/img/pipeline.png

I'm not sure if I'm making those inter-links properly, but I think so.

JMMackenzie (approved):
Assuming the links work, this looks great. Nice diagram btw :-)

elshize:
@amallia can we verify the links before merging? E.g., can I build it locally?

amallia:
I think we shoudl rename `WAND data` in the diagram to reflect the naming we used in the OSIRRC paper (which was `Index Metadata`).

amallia (approved):
LGTM. just a small comment

elshize:
I added an issue for changing the name of wand data: https://github.com/pisa-engine/pisa/issues/442 at this point let's leave it what it is in the diagram and docs, I don't want to create confusion. We should change the name everywhere at once.

PR #443, opened 2020-10-28
Fix reading bytes from bit vector

elshize:
This fix addresses two issues in the interpolative coding. First, we reinterpreted an arbitrary buffer of bytes as 32-bit ints, which introduced undefined behavior. We can do that safely with `std::memcpy`. However, there's another issue that actually is worse: if I understand this right, even though the bit writer uses 32-bit ints as a buffer, the block interpolative codec actually truncates the memory to the exact number of bytes. But in the reader, we still assumed we can just read another 4 bytes each time we run out of buffered bits, which is incorrect and causes reads after buffer.

This might not be a problem most of the time, because we usually just own that memory anyway. But not if that's the last posting list. Also, I was working on another memory source that reads from disk posting lists, and in that case this problem is very bad.

The solution from this PR has possible performance issues. But I don't imagine it's very important, since it's in the already slow interpolative encoding anyway, and in other block codecs we only use it to encode (possibly) the last block (which is the one that's the most problematic). The advantage of this fix is that it is actually correct.

That said, I'm investigating strange segfaults in my experiments, and this is just one of many issues I've noticed. So until I solve my problem, I might either fix or report a few more of these.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
           if [ "${cc}" = "clang-10" ]; then
             wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
             sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"
+            sudo apt-get remove libstdc++-10-dev libgcc-10-dev cpp-10
           fi
           if [ "${cc}" = "clang-6.0" ]; then
             # Problems compiling with Clang 6 against libstdc++-9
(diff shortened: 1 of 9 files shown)


JMMackenzie (approved):
Okay, nice pick up. Can you check the clang issues?

elshize:
> Okay, nice pick up. Can you check the clang issues?

It can't download clang for some reason. I'll check tomorrow on which side the problem is.

elshize:
Seems like there was some change to CI image. I fixed some key problems, and opened #445

amallia:
I am merging this. I would be interested in understanding what's the performance penalty we pay here, but I guess it will be acceptable because of the type of codec used here, as you already said. 
Anyway, this should never fail in our current scenario, but I am glad it fixes your code. Thanks for the thorough PR.

PR #450, opened 2021-02-04
Modify count-postings tool

elshize:
This is a more useful posting counting tool, which counts postings from queries (sums up terms). You can easily count all postings by piping all terms or range of IDs.

It's only implemented for a compressed index because at this moment the uncompressed binary format doesn't support random access, so this is kind of pointless.


JMMackenzie (approved):
Looks good!

PR #453, opened 2021-02-17
Enable counting postings by term

elshize:
`count-postings` so far always summed up all posting counts across all query terms. With this PR, this behavior is achieved with `--sum` flag, and by default, a list of counts is printed for each query. I also added `--query-id` and `--sep` options, similar to those in `map_queries` program.


JMMackenzie (approved):
Looks good to me! This makes it easier to add new plug-in features too, if any ever arise.

PR #457, opened 2021-05-01
Explicitly use Ubuntu 18.04 in CI

elshize:
`ubuntu-latest` is now 20.4, and our CI setup fails on that.


JMMackenzie (approved):
LGTM, but why are we failing on `latest` and should we be looking to fix it longer term? Thanks for taking care of this!

elshize:
There are a few reasons it was failing. First, we use 18.04-specific LLVM repos, but that could be easily changed. The second reason is that the default gcc and clang versions are different, so some changes into installing/removing toolchains need to be done.

I figured we shouldn't be using `latest` anyway, because this will happen in another two years again, when they upgrade to 22.04. We should probably upgrade to `20.04` at some point, but there is no rush.

PR #460, opened 2021-05-24
Add CMake options for fPIC

### LINKED ISSUE #459: Library Builds
As discussed on PyPISA, we need to alter some compilation details for correctly linking. See: https://github.com/pisa-engine/pypisa/issues/1

However, we need to decide whether we'll include these all the time, or whether we want to have two different compilation processes (one for lib builds, for use in PyPISA for example, and one for binaries?)

Any thoughts on this? We might need to double check that we're not introducing any performance regressions when we include these flags, as I know that -fPIC can disrupt inlining for example.

JMMackenzie:
Fixes #459 

- Globally set -fPIC but allow avoidance via `-DNO_GLOBAL_PIC=1` (etc)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,20 +5,19 @@ project(PISA CXX C)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-set(GUMBO_CFLAGS "-fPIC")
-if (NO_GLOBAL_PIC)
-    set(CMAKE_POSITION_INDEPENDENT_CODE OFF)
-    set(GUMBO_CFLAGS "")
-    MESSAGE("Global Position-Independent Code Disabled")
-endif()
 
 option(PISA_BUILD_TOOLS "Build command line tools." ON)
 option(PISA_ENABLE_TESTING "Enable testing of the library." ON)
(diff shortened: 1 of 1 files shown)


elshize:
Some builds are failing. I wonder if it's the flag or it's something else...
Master built fine but that was 22 days ago.

elshize:
Regardless of the failures, maybe we should stick to our convention and define the variable as `option` below? Plus we can call it `PISA_USE_PIC` or something like that, and set the default to `ON`, instead of doing "no-something"

JMMackenzie:
Agreed. And yeah, I think the build issues must be from this, not immediately clear why though...

elshize:
Ok, it seems that `master` fails now, too, I just rerun it: https://github.com/pisa-engine/pisa/runs/2660795628 So maybe something changed in the actions image again? Tbh, this is super annoying and unreliable, I'm starting to think we were better off with travis...

JMMackenzie:
> Ok, it seems that `master` fails now, too, I just rerun it: https://github.com/pisa-engine/pisa/runs/2660795628 So maybe something changed in the actions image again? Tbh, this is super annoying and unreliable, I'm starting to think we were better off with travis...

That's really annoying, yeah :-(

Just pushed a new attempt based on our prior conventions.

elshize (approved):
Looks good. I opened another PR #461 to fix CI. The problem most likely was that they added gcc11 to actions image and something didn't work right. I'm trying now to just build with clang's own std library.

amallia:
LGTM

amallia (approved):

PR #461, opened 2021-05-25
Fix CI problems

elshize:


elshize:
@JMMackenzie I have done it!

I'm just using a docker image for clang-tidy to be able to reproduce when fails. I think we should do that to all tests eventually (#462).

Take another look at this and merge if looks good. Then we can finally take care of that other PR hopefully.

JMMackenzie (approved):
Yep, good idea, this looks good to me. I'll merge now and backtrack to the other PR.

PR #467, opened 2021-07-29
Obey query term weights

JMMackenzie:

- This PR ensures that a query terms repeated _n_ times are given a weight of _n_.
- Previously, we just folded up repeat terms and ignored the weighting.
- Perhaps we should have a flag to specify whether the weighting is desired or now -- it often depends on the context of the search (algo, ranker, etc) as to whether this behavior is desired.


include/pisa/cursor/scored_cursor.hpp
```diff
@@ -33,7 +33,10 @@ class ScoredCursor {
         return m_base_cursor.docid();
     }
     [[nodiscard]] PISA_ALWAYSINLINE auto freq() -> std::uint32_t { return m_base_cursor.freq(); }
-    [[nodiscard]] PISA_ALWAYSINLINE auto score() -> float { return m_term_scorer(docid(), freq()); }
+    [[nodiscard]] PISA_ALWAYSINLINE auto score() -> float
+    {
+        return m_query_weight * m_term_scorer(docid(), freq());
```
elshize: I'm wondering how much this will impact performance. For any non-repeated word in a query, we'll  have to do an additional floating point multiplication. I don't have a good idea at the moment about the overhead. Maybe it's negligible... I wonder if here we could use the fact that we use dynamic dispatch for scorers, and when resolving a scorer, we'd do something like (pseudocode):
```cpp
if (count == 1) { return scorer(); }
else { return weighted_scorer(scorer(), count); }
```
I haven't thought this through though. What's your opinion?

JMMackenzie: Interesting thought. Let me do a quick benchmark to see if there's much penalty; I'll report back soon.

JMMackenzie: Over ClueWeb09B with MaxScore k = 10 (5k queries):
```
Master
{"type": "block_simdbp", "query": "maxscore", "avg": 24908.6, "q50": 15178, "q90": 61869, "q95": 81351, "q99": 125953}

This
{"type": "block_simdbp", "query": "maxscore", "avg": 24669, "q50": 14160, "q90": 61683, "q95": 82117, "q99": 134078}

```
What do you think? Seems negligible. Perhaps I will hand craft some queries with repeat terms? Although I think the floating point op is happening regardless.

elshize: Out of curiosity, could you run it for `k=1000`? I think for simplicity's sake, at least for now, we could use this, provided that it doesn't slow down any more for larger k.

But I have another thought now: regardless of performance, maybe we should have an option to not use a weight? Not sure. Is either one a "canonical" approach? It's not immediately cleart to me that using weights is something we can assume everybody wants. Is there any research about it? I'm not familiar with any, and don't have a good intuition whether or not applying weights would improve precision. What are your thoughts?

JMMackenzie:
```
k = 1000

Master
{"type": "block_simdbp", "query": "maxscore", "avg": 42579.1, "q50": 32520, "q90": 92757, "q95": 118030, "q99": 181862}

This
{"type": "block_simdbp", "query": "maxscore", "avg": 38439, "q50": 27553, "q90": 86294, "q95": 113846, "q99": 180904}

```

Of course, these latency differences are possibly due to differences in pruning (since queries like "cost of beer beer" will up-weight "beer" making query processing potentially faster). So I also ran `ranked_or` to avoid pruning effects; For convenience, I used the `evaluate_queries` binary with 32 threads.

```
Master
[info] Time taken to process queries: 66279ms

This
[info] Time taken to process queries: 67768ms
```

I agree 100% about the option. I did mention it in the original PR too. I think neither is canonical, and we need to be careful with this. Either alerting the user that weights are (or are not) being applied and providing a command line toggle (for example) is probably best.

For example, we could have `--weight-query` which specifies to use the counts. We'd have to pass it through anywhere cursors are created.

On the other hand, we also have:
```
struct Query {
    std::optional<std::string> id;
    std::vector<term_id_type> terms;
    std::vector<float> term_weights;
};
```

So perhaps it's better to totally do away with "counting" terms in queries, and having either "weighted queries" or "unweighted queries" from input? EG, you either specify a weighted query file of some format, or you specify an unweighted query file (like currently in master, if a term appears multiple times it just gets folded into one occurrence).

What do you think about that?

elshize: Thanks, good point about pruning. I think we should be able to just use the multiplication and handle the option of whether or not to use weights at query parsing stage. This would be definitely the easiest, and I think the overhead is minor.

Though I do hate paying for multiplications with no-weights option... Just doesn't feel right, you know? :joy: I would want to look into that at some point, though I'm fine with merging it with the overhead.

Of course, one option would be to have statically dispatched weighted and unweighted cursors, but I think that type of thinking will get us nowhere. We need to think of complexity of the code, and compilation speed.

The only way I see we could do it is, as I mentioned, parameterize the scorer so that we only have overhead for when we use the weights. But then, we'd need to apply the weight at max-score and block-max-score but not at the posting... not sure I like it.

As I said, I'm personally fine with going with the simple approach. We can consider any performance improvements later on, I'd have to look at the code more closely to see if it's worth it.

JMMackenzie: Okay I totally agree.

It would be nice to avoid the multiplication if we can. One option is to template the cursors but of course we'll pay for compilation time.

Let's park it for now and think about the solutions. Once we agree on the "right" way, we'll fix it here and then merge.

elshize:
I think it looks good. How do you feel about this approach?

elshize (approved):

JMMackenzie:
> I think it looks good. How do you feel about this approach?

Yeah, I think this is the most elegant way to go for now at least. Might be good to add some extra test cases for weighted queries too though before it gets merged.

elshize:
Sure, extra tests are always good. The important thing is that the old logic works, which should be the case because there's a bunch of tests for that. But if you find some time to add tests for weighted queries, it's even better.

include/pisa/cursor/block_max_scored_cursor.hpp
```diff
@@ -30,10 +30,12 @@ class BlockMaxScoredCursor: public MaxScoredCursor<Cursor> {
     ~BlockMaxScoredCursor() = default;
 
     [[nodiscard]] PISA_ALWAYSINLINE auto block_max_score() -> float { return m_wdata.score(); }
```
elshize: @JMMackenzie Don't we need to multiply this one by the weight as well?

JMMackenzie: I don't think so. If you have a look at the BMW algorithm you see, for example, the following:

```
              block_upper_bound +=
                    ordered_cursors[i]->block_max_score() * ordered_cursors[i]->query_weight();
```

So the `query_weight` component is being included here. I guess its up to interpretation as to whether this should be "silently" handled by propogating that weight back into the cursor itself (so then you do what you suggested and return m_wdata.score() * query_weight, and just use that single value in the algorithms).

Note that the `query_weight()` call is inherited from the `ScoredCursor` class, and this PR propagates (or at least, should propagate) the correct term weight back if `weighted` is set to true.

Hope this makes sense!

elshize: Ah, yeah, I thought it was implemented somewhere but wasn't sure. I suppose it's a bit inconsistent but I don't want to refactor this now, seems like a lot of work. I'll leave a comment explaining it instead.

JMMackenzie: Agreed, it's kind of weird. Once we merge here, we can make a new issue noting that a refactor is required. I could probably do it soon while this stuff is still fresh, but we'd need to consider whether there's ever a case where it would make sense to decouple the query term weight and the block weights during traversal.

elshize:
> Once we merge here, we can make a new issue noting that a refactor is required.

I'd appreciate it if you could open an issue describing this, so we
could discuss details.

> we'd need to consider whether there's
> ever a case where it would make sense to decouple the query term weight
> and the block weights during traversal.

Right. Should none of them be decupled, or maybe all of them? Not sure,
honestly. Intuitively, decoupling this adds to the complexity of the
algorithms, and may lead to some bugs. On the other hand, it's limiting.
That said, I don't think we actually use it in a context where it would
matter, but still something to consider.

elshize:
@amallia Can you give this a quick look when you get a chance? I don't want to forget about it again...

amallia (approved):

amallia:
This looks good. Sorry for the delay in getting back.
Thanks for the great work both of you have done.

JMMackenzie:
NOTE: This was force merged since the actions for MacOS are seemingly broken. This will be a separate issue.

PR #468, opened 2021-08-27
Add patch for accepting non-trec identifiers from WARC files

### LINKED ISSUE #397: Support modern WARC/1.0 inputs in parse_collection
**Describe the bug**
Downloading a WARC file from the common crawl and piping it into `parse_collection` results in an empty forward index.

**To Reproduce**
1. `wget https://commoncrawl.s3.amazonaws.com/crawl-data/CC-NEWS/2018/02/CC-NEWS-20180201022700-00187.warc.gz`
2. `zcat ./CC-NEWS-20180201022700-00187.warc.gz | ./parse_collection -o ./forward -j 5 -f warc --stemmer porter2 --content-parser html`
3. Check file size of the forward index

**Error message**
```
-rw-rw-r-- 1 ubuntu ubuntu    8 Jun  1 01:51 forward
-rw-rw-r-- 1 ubuntu ubuntu   16 Jun  1 01:51 forward.doclex
-rw-rw-r-- 1 ubuntu ubuntu    0 Jun  1 01:51 forward.documents
-rw-rw-r-- 1 ubuntu ubuntu   16 Jun  1 01:51 forward.termlex
-rw-rw-r-- 1 ubuntu ubuntu    0 Jun  1 01:51 forward.terms
-rw-rw-r-- 1 ubuntu ubuntu    0 Jun  1 01:51 forward.urls
```

**Expected behavior**
The warc records should have been extracted/parsed/added to the forward index.

JMMackenzie:
Fixes #397 

- Uses WARC-Record-ID instead of WARC-TREC-ID where the latter is not available.
- Depends on pending PR here: https://github.com/pisa-engine/warcpp/pull/10
- Will require the submodule update prior to merge

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -89,7 +89,8 @@ record_parser(std::string const& type, std::istream& is)
                         if (rec.has_trecid()) {
                             return std::make_optional<Document_Record>(
                                 rec.trecid(), rec.content(), rec.url());
-                        } else if (rec.has_recordid()) {
+                        }
+                        if (rec.has_recordid()) {
                             return std::make_optional<Document_Record>(
                                 rec.recordid(), rec.content(), rec.url());
                         } 
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -89,9 +89,13 @@ record_parser(std::string const& type, std::istream& is)
                         if (rec.has_trecid()) {
                             return std::make_optional<Document_Record>(
                                 rec.trecid(), rec.content(), rec.url());
-                        } else {
+                        } else if (rec.has_recordid()) {
                             return std::make_optional<Document_Record>(
                                 rec.recordid(), rec.content(), rec.url());
+                        } else {
+                            // This should be unreachable
+                            spdlog::warn("Skipped invalid record: No warc-trec-id or warc-record-id...");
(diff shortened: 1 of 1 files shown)


src/parser.cpp
```diff
@@ -86,8 +86,13 @@ record_parser(std::string const& type, std::istream& is)
                             return std::optional<Document_Record>{};
                         }
                         // TODO(michal): use std::move
-                        return std::make_optional<Document_Record>(
-                            rec.trecid(), rec.content(), rec.url());
+                        if (rec.has_trecid()) {
+                            return std::make_optional<Document_Record>(
+                                rec.trecid(), rec.content(), rec.url());
+                        } else {
```
amallia: Could we do `if (rec.has_warcid()) {` instead?
Then have a final `else` case that throws an exception.

amallia (approved):
Some checks are failing, other than that we can merge it.

src/parser.cpp
JMMackenzie: No problem. I have just updated this and also updated the `warcpp` submodule too.

amallia (approved):

elshize:
:rocket:

PR #472, opened 2021-10-28
Remove PSTL

### LINKED ISSUE #458: Compilation issues on libstdc++10
It seems like there are some issues regarding PSTL when compiling against libstdc++10.

I'm not entirely sure what the problem is. It might be related to the fact that this version of the library has its own parallel algorithms implementation, but not sure. In general, PSTL has been moved to the LLVM project, so I'm not sure what we want to do with that in the long run.

I don't have any more info at the moment, I can add some error messages later, but I mostly wanted to point this out for later investigation, and for the reference for anyone who could have the same problem.

Note that I compiled it with clang11/libc++11 just fine, and this version I think actually has the standard parallel algorithms implemented using TBB & PSTL under the hood. So there is some logic in PSTL that considers this.

Again, the problem is that we can't really upgrade (easily?) PSTL because it seems to be now part of LLVM repository.

### LINKED ISSUE #462: Containers for CI?
Here's an idea: what if we compile & run tests in a docker container instead of directly on the actions image?

We've had problems when the image changes on us, things fail, there's a bunch of preinstalled software that can affect us, it's becoming not reproducible. But can't we just use a docker image, say one for gcc or clang, and just roll with that?

Is there any reason not to do it?

elshize:
At this stage, all algorithms that used PSTL are turned into sequential.
Further effort is needed to implement parallelization directly via TBB.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Build and test
+name: Test MacOS
 
 on:
   push:
@@ -11,23 +11,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        name: [linux-gcc-9, linux-gcc-7, linux-clang-6, linux-clang-10, macos-gcc-9, macos-xcode-11.3]
+        name: [macos-gcc-9, macos-xcode-11.3]
         include:
-          - name: linux-gcc-9
-            os: ubuntu-18.04
-            compiler: "gcc"
-            version: "9"
-          - name: linux-gcc-7
-            os: ubuntu-18.04
-            compiler: "gcc" # default on bionic: gcc-7
-          - name: linux-clang-6
-            os: ubuntu-18.04
-            compiler: "clang"
-            version: "6.0"
-          - name: linux-clang-10
-            os: ubuntu-18.04
-            compiler: "clang"
-            version: "10"
           - name: macos-gcc-9
             os: macos-latest
             compiler: "gcc"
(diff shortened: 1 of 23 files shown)


amallia:
Great thanks. Did you ensure the algorithm still works correctly?

elshize:
I haven't yet done any testing beyond automated tests. That said, I don't expect anything to fail; the calls to PSTL-supported algorithms were just replaced with calls to sequential ones. Still, I'd like to run some larger test before merging.

Right now, I'm working on implementing CI runs for newer compilers, and at the same time making all linux builds run from docker containers. The github action images are not very stable, they keep changing, and have a lot of software already installed by default, which makes it a bit tricky to set up different compilers. I had already implemented the clang tidy test with docker before, now doing a similar thing to unit tests.

elshize:
@JMMackenzie @amallia would either of you be able to run some correctness test on some larger collection than the integration tests? Basically, we'd like to test if the index is still building correctly (all steps), including BP reordering.

Notice that some tests are still pending because I removed them but they are still set up as required. We'll have to change the settings to replace these with the new tests.

elshize:
I added `algorithm.hpp` header with a bunch of algorithm wrappers that will use parallel execution if it is supported. I think this is a reasonable approach. It doesn't complicate the code like tbb would, and to get the faster code you only need libstdc++ 9+ (latest stable is 11).

amallia:
Sounds good to me and, yes, I can run it on a larger collection and let you know

JMMackenzie (approved):
Sorry, finally got around to doing this.

I could successfully parse, index gov2 including BP:

```
[2022-01-28 14:34:46.912] [info] Computing statistics about document ID space
[2022-01-28 14:35:04.317] [info] Average LogGap of documents: 4.46804

VS

[2022-01-28 14:35:54.618] [info] Computing statistics about document ID space
[2022-01-28 14:36:10.770] [info] Average LogGap of documents: 1.62642
```

amallia (approved):

amallia:
Thanks @JMMackenzie for testing it out. Shall we merge it then?

elshize:
@amallia @JMMackenzie I just realized there are still some places we can use our `pisa::execution` to work faster on compatible compilers. But seeing as this has been now tested, if you both are okay with this, I say let's merge it, and I can push a separate PR with the remaining changes.

elshize:
Fixes #462

elshize:
I added a PR branched out from this one: #473 We can either merge that one, or this one and then the other later. Let me know what you guys think.

PR #476, opened 2022-02-28
Use DocId and Score aliases

### LINKED ISSUE #435: Replace topk_queue's entry type
As discussed here:
https://github.com/pisa-engine/pisa/pull/429#discussion_r486323978
we should replace `uint64_t` in the entry type with a custom-defined alias `using DocId = std::uint32_t`. See comments for more info.

elshize:
Fixes https://github.com/pisa-engine/pisa/issues/435
This also includes some additional doc comments for the topk queue.

### DIFF
--- a/include/pisa/type_alias.hpp
+++ b/include/pisa/type_alias.hpp
@@ -1,0 +1,10 @@
+#pragma once
+
+#include <cstdint>
+
+namespace pisa {
+
+using DocId = std::uint32_t;
+using Score = float;
+
+}  // namespace pisa
--- a/tools/queries.cpp
+++ b/tools/queries.cpp
@@ -25,6 +25,7 @@
 #include "scorer/scorer.hpp"
 #include "timer.hpp"
 #include "topk_queue.hpp"
+#include "type_alias.hpp"
 #include "util/util.hpp"
 #include "wand_data_compressed.hpp"
 #include "wand_data_raw.hpp"
@@ -36,7 +37,7 @@ template <typename Fn>
 void extract_times(
     Fn fn,
     std::vector<Query> const& queries,
-    std::vector<Threshold> const& thresholds,
+    std::vector<Score> const& thresholds,
     std::string const& index_type,
     std::string const& query_type,
     size_t runs,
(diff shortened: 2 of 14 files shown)


include/pisa/type_alias.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <cstdint>
+
+namespace pisa {
+
+using DocId = std::uint32_t;
```
amallia, line 7: Isn't there something similar here: https://github.com/pisa-engine/pisa/blob/0efb7926d4928c8aae672ba4d7a1419891f2676d/include/pisa/type_safe.hpp#L91

BTW it seems to be wrong in `type_safe.hpp` as it is defined as `int32_t` instead of `uint32_t`

tools/queries.cpp
```diff
@@ -258,7 +259,7 @@ void perftest(
             Simple_Accumulator accumulator(index.num_docs());
             topk_queue topk(k);
             ranked_or_taat_query ranked_or_taat_q(topk);
-            query_fun = [&, ranked_or_taat_q, accumulator](Query query, Threshold t) mutable {
+            query_fun = [&, ranked_or_taat_q, accumulator](Query query, Score t) mutable {
```
amallia: Since we are on it can we rename `t` with `threshold`? I have the feeling that before since the type was `Threshold` we had an idea of what it is about, but now that is a `Score` type we will lose it.

amallia: Alternatively, a `Threshold` type should be an alias of a `Score` type.

elshize: I'll update the variable, I think `Score` is a better abstraction. Threshold is not a type, it's more of a function, say, you could have time threshold, or distance threshold, and you'd still want those logically to be seconds or kilometers or whatever.

include/pisa/type_alias.hpp
elshize: Yeah, this is something I'm not sure about. So this one is **not** the same as the underlying type, so if we use the type-safe alias, then this will affect a big chunk of the code base. So I was thinking we can use a regular C++ alias, at least for now.

But I can go back and see what trade-offs this would incur. Honestly, I didn't look into that, so maybe I'll do that. I expect for some significant refactoring to be needed if int and DocId are not interchangable in the type system, but I should check.

tools/queries.cpp
amallia: Agreed. thanks

include/pisa/type_alias.hpp
amallia: Let's stick to what we have here for now. We will get back to `type_safe` separately.

amallia (approved):
Looks good. Feel free to merge this.

PR #478, opened 2022-03-07
Move forward index builder implementation to cpp

elshize:
Moves the implementation to a separate compilation unit for faster
compilation. It also includes some minor changes to eliminate most of
the compilation warnings.


JMMackenzie (approved):
All good assuming the tests pass, not sure what's holding them up. Thanks, this looks like a lot of tedious work...

PR #480, opened 2022-03-17
Move implementation to cpp files

elshize:


JMMackenzie (approved):
Thanks! Looks good.

I wonder if we're noticing any compile time memory reductions or speedups yet?

elshize:
> Thanks! Looks good.
> 
> I wonder if we're noticing any compile time memory reductions or speedups yet?

I didn't check memory, but time-wise we're about the same, it seems. But would have to actually test that.

I think these particular changes are quite small; I'm starting with them because they're the low-hanging fruits that are easier to extract.

We might see difference for these if we iteratively recomplie. So say, `invert` tool should be faster, but that's actually not the bulk of time. I'll keep hacking bit by bit, and hopefully I can get to a place where it's easier to tackle the slower parts.

PR #481, opened 2022-04-03
Separates bit_vector_builder from bit_vector

elshize:
This can slightly improve compilation time down the road (more code will
need to get the same type of treatment first), but also cleans it up a
little by making compilation units and header files smaller and easier
to navigate.


JMMackenzie (approved):
Thanks, looks good!

Issue #483, opened 2022-06-05
Disk-streamed indexing can result in non-deterministic indexes if there is insufficient /tmp storage


JMMackenzie:
**Describe the bug**
As the title suggests; I had some issues with segfaults upon mmapping an index. It turned out that the index was being built (but incorrectly) because there was insufficient storage space on `/tmp`. No errors were thrown though.

Running the indexing command with the `--check` flag threw an error (demonstrating that the index built was wrong).

I will look into this in more detail soon, but am just marking it here for now.

cadurosar:
+1, just had the same error for a while, thanks @JMMackenzie for pointing this out. Going for a machine with more /tmp space solved the problem but without that I would have no idea that was the problem.

JMMackenzie:
@elshize I think we'd need to run a check somewhere here: https://github.com/pisa-engine/pisa/blob/196a6e6e0b2312fad5befa58e613be4485294aa6/include/pisa/block_freq_index.hpp#L154

I can never remember the correct way to do it, but I think we need to flush and then check for `.bad()` or something like that?

elshize:
@JMMackenzie yeah, something like this. Didn't have time to have a look today. I'll have to refresh my memory. There are many places like this where we should do those kind of checks tbh. We weren't very good about this. In our defense, this is way more complicated to do than it should be in C++. But it is what it is, we need to address that.

elshize:
@JMMackenzie I think this should be actually resolved, because the issue is about the disk-streamed indexing, which was addressed in #484 

There are more places where checks should be done, but I don't see much sense in tracking all of them here. Let's close, and we'll open specific ones (or one tracking issue) instead.

PR #484, opened 2022-08-12
Throw exception when write fails for block_freq_index

elshize:
This addresses #483, though we have more places to fix, so this won't close the issue.

### DIFF
--- a/.github/workflows/tidy.yml
+++ b/.github/workflows/tidy.yml
@@ -19,9 +19,8 @@ jobs:
       working-directory: ${{runner.workspace}}/pisa/test/docker/tidy
       shell: bash
       run: |
-          docker build . -t tidy-image
-          docker run --name tidy -v ${{runner.workspace}}/pisa:/pisa -d --rm -it tidy-image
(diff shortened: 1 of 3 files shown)


JMMackenzie (approved):

JMMackenzie (approved):

elshize:
There is an issue with tidy build, seems to be caused by some changes in ubuntu image and probably new git version.

But the underlying problem is that docker isn't run as the same user as the one that owns submodule directories. Not sure what the best solution for this is. One could be to copy files to image and change owner. We can also change owner for the files in the volume, but idk if that would affect anything else... Certainly wouldn't want that being run on my local, so would rather not do that. Maybe using podman instead of docker is better? Would have to look it up.

elshize:
@JMMackenzie I finally fixed tidy build! Effectively, I just switched to debian-based clang12 image.

This was clearly caused by some updates in the Ubuntu image. I'm still a bit confused as to whether it's git or cmake. Either way, this problem could at some point come back. But I say let's cross that bridge when we get there. The bullseye image is unlikely to move much in the near future :)

There's a codacy error saying I should fix package version in the docker image, but I elected to ignore that.

PR #489, opened 2022-11-10
Move TemporaryDirectory to pisa namespace (#419)

### LINKED ISSUE #419: Temporary_Directory is not in pisa namespace
We should define it in `pisa` namespace, and while we're at it, remove the underscore from the name.

elshize:
Temporary_Directory is renamed to TemporaryDirectory and moved inside pisa namespace. Implementations are moved to a separate cpp file.

Fixes #419


JMMackenzie (approved):

JMMackenzie:
Wtf is with these Mac builds :joy:

elshize:
> Wtf is with these Mac builds

No idea. Not having access to a mac makes it difficult to say if it's mac issue or github actions issue.

PR #490, opened 2022-11-11
project readme in Chinese and ADD the company Algmon tech SUPPORT in Asia

ghost:
For the Company Algmon Intro, please SEE
https://algmon.com/
a formal NYU student created company


elshize:
We are not affiliated with nor do we endorse any companies, regardless of whether they use PISA. While you are welcome to fork and use the project in compliance with the Apache2 license, we are not accepting any 3rd party promotional material, such as their names, support contact, etc.

As far as the translations go, we do not have specific policy or approach, as there has never been a request for one. That said, I think a more appropriate would be a separate translated document, to which we can link from the main readme.

PR #492, opened 2022-11-20
Improve block codecs

### LINKED ISSUE #407: Improve block codecs
Hi everyone! I bring up for discussion some possible improvements that could be introduced in the `block codecs`.

### Interpolative
Starting here:
https://github.com/pisa-engine/pisa/blob/35dd03b73506d2089686a4a211e38898671b425c/include/pisa/codec/block_codecs.hpp#L151
Why not just use `in` instead of `inbuf`?

On the other hand, in the next line:
https://github.com/pisa-engine/pisa/blob/35dd03b73506d2089686a4a211e38898671b425c/include/pisa/codec/block_codecs.hpp#L152
Could there be an improvement using `numeric_limits<uint32_t>::max()` instead of `uint32_t(-1)`?

### Simple16
Given that 28 is the maximum `batch size`, I think that the `buf` size could be reduced to `4 * block_size + block_size % 28` instead of `2 * 8 * block_size`:
https://github.com/pisa-engine/pisa/blob/35dd03b73506d2089686a4a211e38898671b425c/include/pisa/codec/simple16.hpp#L14
Although the buffer size is not a problem for the current memory sizes, if you want to use a `block_size` greater than 128, the memory saving is more substantial. On the other hand, the `Simple8b` codec could be improved in a similar way. 

Again on `Simple16`, on the decoding could be used a `memcpy` instruction instead a `raw loop`:
https://github.com/pisa-engine/pisa/blob/35dd03b73506d2089686a4a211e38898671b425c/include/pisa/codec/simple16.hpp#L30
Alternatively, it could be used the `std::copy` function, but it should be slower than `memcpy` as with https://lemire.me/blog/2020/01/20/filling-large-arrays-with-zeroes-quickly-in-c/.

### MaskedVByte
Although `4 bytes` is the maximum size of a raw integer, in the `MaskedVByte` codec is required a continuation bit for each byte. Then, the maximum size of an integer encoded with this one will be `5 bytes`. Then, `buf` could be redefined as `5 * block_size`:
https://github.com/pisa-engine/pisa/blob/35dd03b73506d2089686a4a211e38898671b425c/include/pisa/codec/maskedvbyte.hpp#L21

In same way, the buffers of `VarintGB` and `VarintG8IU` codecs could be resized as `4 + 1 bytes`, assuming (for simplicity) `1 byte descriptor` for each integer. Currently the buffer size for these codecs is `2 * 4 * block_size`, too.

elshize:
Fixes #407

### DIFF
--- a/include/pisa/codec/block_codecs.hpp
+++ b/include/pisa/codec/block_codecs.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <array>
+
 #include "FastPFor/headers/optpfor.h"
 #include "FastPFor/headers/variablebyte.h"
 
@@ -121,13 +123,14 @@ class TightVariableByte {
 };
 
 struct interpolative_block {
-    static const uint64_t block_size = 128;
+    static constexpr std::uint64_t block_size = 128;
 
     static void encode(uint32_t const* in, uint32_t sum_of_values, size_t n, std::vector<uint8_t>& out)
     {
         assert(n <= block_size);
-        thread_local std::vector<uint32_t> inbuf(block_size);
-        thread_local std::vector<uint32_t> outbuf;
+        thread_local std::array<std::uint32_t, block_size> inbuf{};
+        thread_local std::vector<uint32_t> outbuf;  // TODO: Can we use array? How long does it need
+                                                    // to be?
         inbuf[0] = *in;
         for (size_t i = 1; i < n; ++i) {
             inbuf[i] = inbuf[i - 1] + in[i];
@@ -213,7 +216,7 @@ struct optpfor_block {
         uint8_t const* b = nullptr)  // if non-null forces b
     {
         thread_local codec_type optpfor_codec;
-        thread_local std::vector<uint8_t> buf(2 * 4 * block_size);
+        thread_local std::array<std::uint8_t, 2 * 4 * block_size> buf{};
         assert(n <= block_size);
 
         if (n < block_size) {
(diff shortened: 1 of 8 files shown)


JMMackenzie (approved):
I'm happy for this to be approved given the related discussion in #407 but do we want to extend to all codecs in this PR, or address later?

elshize:
> do we want to extend to all codecs in this PR, or address later?

I was thinking of extending it; I should have some time next week to do it, and run some tests just to make sure all is good. I'll ask you to have another look when I push more changes.

elshize:
I ran test for all modified codecs, looks like it's at least not worse, but mostly better performance now:

image

image

Raw data: [results.tar.gz](https://github.com/pisa-engine/pisa/files/10093758/results.tar.gz)

So I'm pretty comfortable merging that.

include/pisa/codec/block_codecs.hpp
```diff
@@ -121,13 +123,14 @@ class TightVariableByte {
...
-    static const uint64_t block_size = 128;
+    static constexpr std::uint64_t block_size = 128;
 
     static void encode(uint32_t const* in, uint32_t sum_of_values, size_t n, std::vector<uint8_t>& out)
     {
         assert(n <= block_size);
-        thread_local std::vector<uint32_t> inbuf(block_size);
-        thread_local std::vector<uint32_t> outbuf;
+        thread_local std::array<std::uint32_t, block_size> inbuf{};
+        thread_local std::vector<uint32_t> outbuf;  // TODO: Can we use array? How long does it need
```
JMMackenzie, line 132: Shall we convert this TODO to an issue?

JMMackenzie (approved):

include/pisa/codec/block_codecs.hpp
elshize: Will do.

PR #497, opened 2022-11-27
CLI option to configure log level

### LINKED ISSUE #495: Control log level
We should add a CLI argument to control the log level. For example, `queries` shows a lot of warnings about missing terms. While this may be good, safe default, I found this super annoying when scripting lots of executions. For cases like that, I'd like to set it to ERROR, so we only get errors. The only current way to silence it is by silencing all of the logs, but then errors won't show, which may not be the best.

elshize:
All programs using CLI11 get a new option -L,--log-level to configure log level at runtime.

Fixes #495

### DIFF
--- a/include/pisa/util/progress.hpp
+++ b/include/pisa/util/progress.hpp
@@ -5,35 +5,33 @@
 #include <iostream>
 #include <thread>
 
+#include <spdlog/spdlog.h>
+
 namespace pisa {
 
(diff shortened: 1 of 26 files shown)


JMMackenzie (commented):
Nice, a productive weekend :-)

JMMackenzie (approved):

PR #499, opened 2022-12-01
Selecting tokenizer at query time

elshize:
Note that the code around parsing is a bit of a mess generally. I'd like to address that but I want to do this step by step.

For now, I just enhanced the solution we have now for query parsing, so that we can chose between tokenizers. The old English tokenizer is still the default if nothing is defined.

I want to then implement it on the collection parsing side, and only after that I'd like to unify it a little by introducing some kind of parsing "pipeline". I don't have the details nailed down yet, I'll share them when I get a chance to think about it.

### DIFF
--- a/test/test_queries.cpp
+++ b/test/test_queries.cpp
@@ -41,7 +41,8 @@ TEST_CASE("Compute parsing function")
 
     WHEN("No stopwords, terms, or stemmer")
     {
-        auto parse = resolve_query_parser(queries, std::nullopt, std::nullopt, std::nullopt);
+        // Note we don't need a tokenizer because ID parsing does not use it
+        auto parse = resolve_query_parser(queries, nullptr, std::nullopt, std::nullopt, std::nullopt);
         THEN("Parse query IDs")
         {
             parse("1:0 2 4");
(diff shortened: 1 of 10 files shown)


test/test_queries.cpp
```diff
@@ -53,7 +54,11 @@ TEST_CASE("Compute parsing function")
     WHEN("With terms and stopwords. No stemmer")
     {
         auto parse = resolve_query_parser(
-            queries, lexfile.string(), stopwords_filename.string(), std::nullopt);
+            queries,
+            std::make_unique<WhitespaceTokenizer>(),
```
JMMackenzie, line 58: Just checking that you actually want `<WhitespaceTokenizer>` here?

test/test_queries.cpp
```diff
@@ -64,8 +69,12 @@ TEST_CASE("Compute parsing function")
     }
     WHEN("With terms, stopwords, and stemmer")
     {
-        auto parse =
-            resolve_query_parser(queries, lexfile.string(), stopwords_filename.string(), "porter2");
+        auto parse = resolve_query_parser(
+            queries,
+            std::make_unique<WhitespaceTokenizer>(),
```
JMMackenzie, line 74: And again: Just checking that you actually want `<WhitespaceTokenizer>` here?

JMMackenzie: Upon further checking I think you do, so feel free to ignore these :-)

JMMackenzie (approved):
Looks good, thanks! I left a minor comment but I think it's correct so please ignore if so.

test/test_queries.cpp
elshize: Yeah, the truth is it doesn't really matter here, this is not what's mainly being tested. Both tokenizers will work the same way. I added a test that does check the tokenizer works as expected, but for the rest I just opted for the simple one. But it makes no difference, really.

PR #503, opened 2022-12-24
Implement TextAnalyzer

### LINKED ISSUE #494: BERT tokens

elshize:
A _text analyzer_ consists of:
 1. zero or more text filters (e.g., strip HTML),
 2. a tokenizer,
 3. zero or more token filters (e.g., stemming).

For example, we a common text analyzer for the document content would consist of (1) strip HTML filter, (2) standard English tokenizer, (3) lowercase filter, stemmer, and stopword remover.

A _text filter_ takes a string input and returns the transformed string. The only implemented text filter at the moment is the one stripping HTML markup.

A _tokenizer_ takes a string and returns a token stream. See `TokenStream` for the details. We currently implement English and whitespace tokenizers.

A _token filter_ takes a single token and returns a token stream. The stream can always return a stream containing a single token (1-1 transformation), potentially return no tokens (stopword removing), or return multiple words. None of our currently implemented filters returns multiple tokens, but in the future we can consider implementing filters that do some term expansion, like synonyms.

A text analyzer is used for parsing both queries and documents, after the content part is already extracted (either from colon-delimited query string or from document input format, such as TREC).

---

Fixes #494

---

There are still some outstanding items to do before merging:
- [x] Update documentation
- [x] Write the one missing CLI test
- [x] Fix CLI tests


JMMackenzie (approved):
Nice work, looks good to me!

PR #507, opened 2022-12-31
Custom sift down for binary heap

### LINKED ISSUE #504: `topk_queue` rapidcheck tests fail on Clang 15
We have some property tests implemented that check `topk_queue` thresholds, and they mysteriously fail on Clang 15.

elshize:
Fixes #504

The previous solution where we swap first and last elements of the heap
and call `pop_heap` is incorrect according to the standard, and fails
when compiled with libc++ 15. Calling `push_heap` in addition to
`pop_heap` is less efficient; in fact, the custom sift down solution is
more efficient than the previous implementation (according to the
microbenchmarks).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -70,3 +70,6 @@
 [submodule "external/taily"]
 	path = external/taily
 	url = https://github.com/pisa-engine/taily.git
+[submodule "external/benchmark"]
+	path = external/benchmark
+	url = https://github.com/google/benchmark.git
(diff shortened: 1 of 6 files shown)


elshize:
FYI, the tests will fail; I fixed them in the other PR, so we should merge that one first...

JMMackenzie:
@elshize I'll check this out once we merge #503 - can you re-request my review please?

JMMackenzie:
@elshize ready for review? How does it compare in the benchmark?

elshize:
> @elshize ready for review? How does it compare in the benchmark?

Ah, sorry, forgot to tag you. Yes, it's ready for review. As far as benchmarks go, I only did the microbenchmarks but no query-on-index type of benchmark. It is actually ever-so-slightly faster than the baseline on some data, and within the margin of error on other.

It boils down to the fact that the recent versions of both LLVM and GNU standard libs implement this optimization that assumes the sifted-down element is a leaf, which ends up being a pessimisation for a lot of inserts for us.

Anyway, looks like it performs the same or better, and this is actually fixing a bug, because the current implementation assumes a certain way the algorithm works that's not part of the standard (and now started failing on LLVM).

JMMackenzie (approved):

PR #510, opened 2023-01-18
Fix log of zero

elshize:


JMMackenzie (approved):
Nice pick up.

I guess we never called log2(0) since the application of the pre-computed logarithms was on d-gaps {d |d > 0}?

elshize:
Yeah, I don't think we were ever calling `log2(0)` but Codacy picked it up as a warning, and I can't be bothered to check in the standard if this is defined behavior (even though cppreference says it just should return -inf, but then again, it was wrong about the heap stuff before). So just to be on the safe side, I decided to fill this one myself, so we avoid any potential UB.

PR #513, opened 2023-01-21
C++20 Support

elshize:
This includes three things:
1. Upgraded range-v3 to the latest version
2. Fixed a namespace issue in our own code base
3. Fixed CI (and added one C++20 build)

The last one is not really related to any of the changes, but it seems to be related to a cmake bug. Not sure what happened, I suspect the system cmake was updated in the base image. For some reason Debian doesn't have the last point version of 3.13 that was released in 2019 but stopped at 3.13.4, which had that bug that caused failures.

Anyhow, I thought it might be best to simply get cmake from upstream, and saw no reason not to use the latest version.


JMMackenzie (approved):
Well, that was easier than I thought it would be :grin:

test/docker/cmake-3.25.2-SHA-256.txt
```diff
@@ -0,0 +1,17 @@
+82ef7846809a6cfe282a6196b300b081ff7878e40065c1d530352f8f7a2a0c2d  cmake-3.25.2-files-v1.json
```
amallia, line 1: do we need this?

amallia: yes we do

amallia (approved):

test/docker/cmake-3.25.2-SHA-256.txt
elshize: Yeah, I want to check sums, I'm paranoid like this :)

elshize:
> Well, that was easier than I thought it would be

This how much faith you have in me? :grin: But seriously, it was just a `using namespace` bug on our side really. C++ standards are famously backwards-compatible to a point (this is actually big point of contention, because it also means you can't change bad old stuff much).

PR #515, opened 2023-01-25
topk_queue::finalize sorts by both score and ID (#508)

### LINKED ISSUE #508: stability of document order when tied
**Describe the bug**
Since [`sort_heap` is not stable](https://stdcxx.apache.org/doc/stdlibref/sort-heap.html) and [`topk_queue` only sorts on score](https://github.com/pisa-engine/pisa/blob/379d94fa93b3e6811e21bc44256822d8f78cc214/include/pisa/topk_queue.hpp#L132), the order of results that have the same scores can differ between runs. This is [harmful for the repeatbility](https://dl.acm.org/doi/10.1145/3331184.3331339) of results.

**To Reproduce**
1. Follow the steps to index and execute queries. Be sure to include a query that will result in a score tie for the ranking model. For instance, the query "chemical reactions" over the [vaswani](https://ir-datasets.com/vaswani) dataset.
2. Observe the order of documents with tied scores.
3. Repeat and re-observe several times.

**Error message**
No error message.

**Expected behavior**
I expect results to be ordered ascending by DocId as a secondary sort when scores result in ties.

**Environment info**
Operating System: Ubuntu 22.04.1
Compiler name: gcc
Compiler version: 9

elshize:
The final sorting order is now by score (descending) and docid (ascending). Furthermore, `std::push_heap` is replaced with our own implementation to maintain consistency across standard libraries.

Fixes #508

### DIFF
--- a/include/pisa/topk_queue.hpp
+++ b/include/pisa/topk_queue.hpp
@@ -16,6 +16,13 @@ namespace pisa {
 /// min element. Because it is a binary heap, the elements are not sorted;
 /// use `finalize()` member function to sort it before accessing it with
 /// `topk()`.
+///
+/// Note that `finalize()` breaks ties between entries with equal scores by
+/// sorting them by document ID. However, this is not true for insertion, and
+/// thus it may happen that a document with identical score as the k-th
+/// document in the heap but with a lower document ID is _not_ in the top-k
+/// results. In such case, which entry makes it to top k is determined by the
+/// order in which elements are pushed to the heap.
 struct topk_queue {
     using entry_type = std::pair<Score, DocId>;
 
@@ -50,7 +57,7 @@ struct topk_queue {
         }
         m_q.emplace_back(score, docid);
         if (PISA_UNLIKELY(m_q.size() <= m_k)) {
-            std::push_heap(m_q.begin(), m_q.end(), min_heap_order);
+            push_heap(m_q.begin(), m_q.end());
             if (PISA_UNLIKELY(m_q.size() == m_k)) {
                 m_effective_threshold = m_q.front().first;
             }
@@ -69,17 +76,25 @@ struct topk_queue {
 
     /// Sorts the results in the heap container in the descending score order.
     ///
-    /// After calling this function, the heap should be no longer modified, as
-    /// the heap order will not be preserved.
+    /// If multiple entries have equal score, they are sorted by document ID. Notice that this only
+    /// happens in the finalization step; due to performance considerations, inserting is done with
+    /// score-only order. After calling this function, the heap should be no longer modified, as the
+    /// heap order will not be preserved.
     void finalize()
     {
-        std::sort_heap(m_q.begin(), m_q.end(), min_heap_order);
-        size_t size = std::lower_bound(
-                          m_q.begin(),
-                          m_q.end(),
-                          0,
-                          [](std::pair<Score, DocId> l, Score r) { return l.first > r; })
-            - m_q.begin();
+        auto sort_order = [](auto const& lhs, auto const& rhs) {
+            if (lhs.first == rhs.first) {
+                return lhs.second < rhs.second;
+            }
+            return lhs.first > rhs.first;
+        };
+        // We have to do a full sort because it is not the exact same ordering as when pushing
+        // elements to the heap.
+        std::sort(m_q.begin(), m_q.end(), sort_order);
+
+        auto search_order = [](auto entry, auto score) { return entry.first > score; };
+        auto first_zero_score = std::lower_bound(m_q.begin(), m_q.end(), 0, search_order);
+        std::size_t size = std::distance(m_q.begin(), first_zero_score);
         m_q.resize(size);
     }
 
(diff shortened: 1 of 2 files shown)


seanmacavaney (approved):
Thanks! These changes look reasonable to me.

include/pisa/topk_queue.hpp
```diff
@@ -69,17 +76,25 @@ struct topk_queue {
...
+            if (lhs.first == rhs.first) {
+                return lhs.second < rhs.second;
+            }
+            return lhs.first > rhs.first;
+        };
+        // We have to do a full sort because it is not the exact same ordering as when pushing
+        // elements to the heap.
+        std::sort(m_q.begin(), m_q.end(), sort_order);
+
+        auto search_order = [](auto entry, auto score) { return entry.first > score; };
```
amallia, line 95: Could add a comment here to explain what this is doing? I know this is unmodified behaviour, but since you are on it would be better to document the code.

include/pisa/topk_queue.hpp
```diff
@@ -69,17 +76,25 @@ struct topk_queue {
...
-            - m_q.begin();
+        auto sort_order = [](auto const& lhs, auto const& rhs) {
+            if (lhs.first == rhs.first) {
+                return lhs.second < rhs.second;
+            }
+            return lhs.first > rhs.first;
+        };
+        // We have to do a full sort because it is not the exact same ordering as when pushing
+        // elements to the heap.
+        std::sort(m_q.begin(), m_q.end(), sort_order);
```
amallia, line 93: If we are switching to `sort` then why don't we just use `stable_sort` on the scores such that we maintain as second sorting condition the insertion time?

include/pisa/topk_queue.hpp
```diff
@@ -172,6 +181,26 @@ struct topk_queue {
         }
     }
 
+    // We use our own function (as opposed to `std::heap_push`), to ensure that
+    // heap implementation is consistent across standard libraries.
```
amallia, line 185: What does `to ensure that heap implementation is consistent across standard libraries` mean?

elshize: Means that if I compile on libstdc++ 12 and you compile on libc++ 10, then both of us have the same logic, and thus if you run the same exact algorithm on the same exact collection, etc., you will get the same results.

With `std::push`, the implementation may change from version to version and implementation to implementation (LLVM/GNU/MS/Apple) so it's not necessarily reproducible across environments.

Using pair ordering (score, doc) woulds of course solve all of that, but that's not efficient.

include/pisa/topk_queue.hpp
elshize: I'm not sure `stable_sort` will preserve that. I actually haven't analyzed this in detail, but are you saying that given our implementation, we can be sure that the more recently inserted entry is always above in the heap or at the same level but on the left side?

amallia: The issue is that `heap_sort` creates some issues with reproducibility since two documents with equal score have no guarantees on the order they will appear in the list of results.
Stable sort will keep the relative order of elements with equivalent values, fixing the reproducibility issue.

I prefer this solution because:
(1) it is easier and less error prone
(2) it resolves the issue with docID ordering not maintained while inserting and only done at the end, which could cause the problem described in the code comment.

include/pisa/topk_queue.hpp
amallia: Understand! Makes sense. Right now we are not aware of differences in the implementation  of libc++ 10 and libstdc++ 12, right?

My major objection to this approach is that if we stick to this paradigm, we would need to reimplement the entire STL.

include/pisa/topk_queue.hpp
elshize: So what you're suggesting is to change the semantic: instead of ensuring that within the top-k results ties are broken by the docID order, you're proposing to not give any such guarantee, but rather say that the order of elements of with the same score is undefined, but deterministic, i.e., reproducible across multiple runs.

Do I understand correctly?

include/pisa/topk_queue.hpp
elshize: > My major objection to this approach is that if we stick to this paradigm, we would need to reimplement the entire STL

Normally, I would agree, but this is somewhat special circumstances. Technically speaking, we probably would want to strictly define the order (score + docid) but we're not because of efficiency. But also we are already implementing the more complex pop operation, so we might as well finish the job. It's very little effort for possibly a significant improvement for someone down the line.

include/pisa/topk_queue.hpp
elshize: > it resolves the issue with docID ordering not maintained while inserting and only done at the end, which could cause the problem described in the code comment

You'll have to be more explicit here, not sure which issue we're resolving. From my understanding, you're solution simply changes the semantic of what we do, but maybe I'm missing something, so if you could be a bit more clear about what you mean and how it solves it, that would help.

include/pisa/topk_queue.hpp
elshize: Let's put it this way: you're doing research, and have some results, and then you recompile after system upgrade or something, and it doesn't match; it's not life changing, but that's very annoying imo, and I personally see no reason not to fix that with one little function. It could save someone a headache down the line.

PR #516, opened 2023-01-26
Rank start from 1 instead of 0

amallia:
Rank in TREC files should start from 1. This is probably not very important when evaluating with `trec_eval`, but it is crucial when the TREC file is converted to a TSV file for `msmarco_passage_eval.py`


elshize (approved):

amallia:
Thanks for approving. What is the CI waiting for?

elshize:
> What is the CI waiting for?
Ignore that. I added more jobs here: https://github.com/pisa-engine/pisa/pull/513 and I changed the settings to wait for new jobs instead of old ones. Just merge; once the other PR is merged, it should work fine.

PR #519, opened 2023-02-05
Accumulator concept

elshize:
Implements PartialScoreAccumulator concept (only under C++20), and introduces a slight refactoring of the accumulator API, including renaming methods and making some methods private.

These changes are part of the effort to stabilize library API. Concepts not only help with debugging and compilation time using C+20-compatible compiler, but also serve as a programmatically verifiable API documentation. Thus, even if not always used, they are extremely useful.


JMMackenzie (approved):
Nice, this looks interesting/useful

PR #522, opened 2023-02-09
Fix TREC run format

amallia:
Switch from tabs to spaces in TREC run output file


JMMackenzie (approved):
Fine by me, I thought most eval tools accepted any type of whitespace by default though?

elshize:
> Fine by me, I thought most eval tools accepted any type of whitespace by default though?

That's what I thought, too, but I suppose then a space is as good as a tab.

elshize (approved):

PR #523, opened 2023-02-10
Split CI into push, trunk, and PR flows

### LINKED ISSUE #521: Run some CI tests only on merge
We are currently running a lot of CI builds and we'll likely add more. But all of those don't necessarily have to be executed for each commit. Let's instead select a subset of those that we want to keep, and configure the rest to only run when a PR is merged.

elshize:
All jobs are now split into three workflows:
 * push: run on any commit/tag push,
 * trunk: run only on a push to master,
 * pr: run only on a PR push.

Many tests were moved to trunk, including many unit test with different compiler combinations, MacOS build, and clang-tidy build. On a PR push, we run the same unit tests but for smaller number of combinations. On any push, we run some faster builds, like format, CLI tests, etc.

Fixes #521


elshize:
@JMMackenzie @amallia I'm debating whether to keep coverage here or move it out to only run after merge. It's by far the slowest of these, at 38 min. That's almost twice as long as the next slowest, GCC test build.

By the way, there's big difference in build time between GCC and Clang, the latter being almost twice as fast! Not sure why that is.

JMMackenzie (approved):

JMMackenzie:
> @JMMackenzie @amallia I'm debating whether to keep coverage here or move it out to only run after merge. It's by far the slowest of these, at 38 min. That's almost twice as long as the next slowest, GCC test build.
> 
> By the way, there's big difference in build time between GCC and Clang, the latter being almost twice as fast! Not sure why that is.

Interesting, I've also noticed that issue with Clang compilation locally before. Not really sure what the deal with that is. While we're discussing it, have we ever compared Clang and GCC to see if the code generated has any impact on latency?

elshize:
> By the way, there's big difference in build time between GCC and
> Clang, the latter being almost twice as fast! Not sure why that is.

I don't believe we have ever made such a comparison. We've always
historically used GCC as far as I'm aware of. My typical use of Clang is
for clangd in my language server for nvim, clang-tidy, and sometimes I
build it with Clang if GCC error messages don't give me much information
(Clang can be sometimes much clearer on those).

It would be definitely interesting to do some benchmarks in this
capacity.

Do you have an opinion on coverage? Leave it as is or move it to
master-only?

elshize:
Ok, I haven't heard from anyone, I'm going to merge, we can always tweak later.

PR #535, opened 2023-03-26
Use rapidcheck to test codecs

elshize:
Rapidcheck makes failures reproducible, as opposed to what we had, which was using `rand` to generate some values.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        compiler: [gcc12, clang12]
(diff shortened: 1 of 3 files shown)


JMMackenzie (approved):
Looks good to me.

For posterity, I am also noting here the existence of UB if we don't have `T dst{0}` for Clang in the `include/pisa/memory.hpp` file which caused immense pain for @elshize in particular.

elshize:
> For posterity, I am also noting here the existence of UB if we don't have T dst{0} for Clang in the include/pisa/memory.hpp file which caused immense pain for @elshize in particular.

Ah, thanks for immortalizing this low moment for me :joy:

PR #540, opened 2023-08-19
Allow for system Boost installation

elshize:
Together with `BOOST_NO_CXX98_FUNCTION_BASE`, newer version of Boost fix builds on newer Clang versions. Currently, the way we pull Boost automatically only allows 1.67, and this one will fail to compile. This needs to be addressed in the near feature as well.

---

Replace `boost::filesystem` with `std::filesystem`

`std::filesystem` is now supported in all of our supported platforms,
and Boost crashes on newer clang compilers.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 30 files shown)


JMMackenzie (approved):
Sorry for the long delay, looks good to me!

PR #549, opened 2023-10-08
Migrate to mdBook documentation

### LINKED ISSUE #505: Migrate from Sphinx to mdBook
This is just a suggestion at the moment, but mdBook seems a bit more cleaner + easier to use. It's built around markdown and config is done in TOML, both very simple and well-known formats.

I will look into it a bit more to confirm it actually is as simple as I think.

elshize:
Fixes #505 

---

The intention is to host docs at pisa-engine.github.io/book

---
Most of "CLI Reference" description is missing, should be filled up with more details in the future.


elshize:
Note that `docs/book` are generated files, no need to review.

JMMackenzie (approved):
Neato! Fixing documentation is a big "want" for me but I am too time poor :crying_cat_face:

elshize:
> Neato! Fixing documentation is a big "want" for me but I am too time poor

Yeah, this still will need more work but we need to be better about it.

PR #556, opened 2023-11-18
Remove configuration.hpp

### LINKED ISSUE #322: Get rid of configuration.hpp
Get rid of `configuration.hpp`

elshize:
Removes the configuration file that defined how many bits to use for quantization. Instead, this is now configured by CLI options.

Fixes #322


JMMackenzie (approved):
Cool, this is good. And no default bits specified in the CLI, right? I couldn't see anything in `app.*` but just double checking.

elshize:
> Cool, this is good. And no default bits specified in the CLI, right? I couldn't see anything in `app.*` but just double checking.

Correct, no default because the option also says "do quantization", which means the default is no quantization bits.

PR #561, opened 2023-12-30
Query refactoring

### LINKED ISSUE #501: `Query::term_weights` is not assigned
`Query::term_weights` seems to be never assigned. The `--weighted` flag in `queries` and `evaluate_queries` programs ignore those, and instead resolve weights when creating cursors.

elshize:
Weights are now stored together with term IDs and resolved at construction time according to one of the policies. In our tools, we use the default policy that removes duplicates and sets the weight to the number of occurrences of the term in a query. Other policies are, for the time being, only available programmatically via the library API.

Some legacy code used to parse and process queries has been removed in favor of the text analyzer and the new query parser.

Because weights are resolved when a query object is created, I also refactored creating the cursors: now the weight is simply taken from the query.

Fixes #501


include/pisa/query.hpp
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Tells `Query` how to process the terms passed to the constructor.
+     *
+     * By default, duplicate terms will be removed, and the weight of each term will be equal to
+     * the number of occurrences of that term in the query. Furthermore, the order of the terms
+     * will be preserved (if there are duplicates, the term will be at the position of its first
+     * occurrence.
+     *
+     * This policy can be modified with the following options:
+     *  - `keep_duplicates`: duplicates will be preserved, each with weight 1.0;
```
JMMackenzie: Should we note that this is *very bad* in terms of efficiency?

JMMackenzie (approved):
This looks great, nice work.

One quick thought - do we need to worry about the `Thresholds` tool/use here? I assume not, since thresholds are applied on a per-query basis, but I am just trying to think of any corner cases we might have missed. It might also be worth running some regression testing on this since it's quite a sensitive (but very positive) change to the inner processing logic? I'd be happy to run some tests comparing to master in the coming week or two if that would be useful.

include/pisa/query.hpp
elshize: Good idea

elshize:
Right, I don't think thresholds should be any different than queries, but I'll give it another look.

Good idea about regression tests. There's a test docker image that I created for that, but haven't finished. Maybe it's a good idea to continue with it to make it easier to repeat in the future (or even automate).

include/pisa/query.hpp
elshize: Check out my edit.

JMMackenzie (approved):

include/pisa/query.hpp
JMMackenzie: Great, thanks a lot!

elshize:
I need to fix the conflicts, and after that, I'll run the docker that was just merged in the other PR to evaluate and see if there's any regression.

elshize:
@JMMackenzie The regression test was successful. Are you ok merging it?

JMMackenzie:
Great, let's merge!

PR #566, opened 2024-01-16
Fix ranked_or_taat and ranked_or_taat_lazy

amallia:
Currently `ranked_or_taat` and `ranked_or_taat_lazy` fail with segfault, this change should fix them.


elshize (approved):
Looks good to me, if you could just rebase on master and fix the formatting (the test looks to be failing) it would be perfect.

PR #568, opened 2024-01-18
Append sourcing setvars.sh to .bashrc

elshize:
This is not only needed for compilation but also for running the tests.


JMMackenzie (approved):
Sorry, I didn't realize when I approved the last PR - so I am equally guilty :-)

PR #571, opened 2024-01-21
Concepts

elshize:
This change set contains commits that introduce C++20 concepts for:
* mappings,
* posting lists,
* inverted index.

Concepts are only used when compiled in C++20 mode, ignored otherwise.

Assertions are added to known structures that satisfy these concepts.

Some `require` constraints are also added, most notably for the algorithm execution that expect certain posting lists.

## Rationale

Concepts are introduced for faster constraint checking and explicit interface modeling when templates are used.

### DIFF
--- a/include/pisa/term_map.hpp
+++ b/include/pisa/term_map.hpp
@@ -4,6 +4,9 @@
 #include <string>
 #include <string_view>
 
+#include "concepts.hpp"
+#include "concepts/container.hpp"
+#include "concepts/mapping.hpp"
 #include "payload_vector.hpp"
 
 namespace pisa {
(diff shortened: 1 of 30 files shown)


include/pisa/term_map.hpp
```diff
@@ -17,22 +20,26 @@ class TermMap {
     TermMap& operator=(TermMap&&);
     virtual ~TermMap();
 
-    [[nodiscard]] virtual auto operator()(std::string_view term) -> std::optional<std::uint32_t> = 0;
-    [[nodiscard]] virtual auto operator()(std::string const& term)
+    [[nodiscard]] virtual auto find(std::string_view term) const -> std::optional<std::uint32_t> = 0;
```
JMMackenzie, line 23: We should note this API change explicitly for anyone using the lib.

include/pisa/term_map.hpp
```diff
@@ -41,10 +48,16 @@ class LexiconMap final: public TermMap {
     explicit LexiconMap(std::string const& file);
     explicit LexiconMap(Payload_Vector<std::string_view> lexicon);
 
-    [[nodiscard]] virtual auto operator()(std::string_view term)
+    [[nodiscard]] auto operator[](std::uint32_t term_id) const -> std::string_view;
+
+    [[nodiscard]] virtual auto find(std::string_view term) const
         -> std::optional<std::uint32_t> override;
-    [[nodiscard]] virtual auto operator()(std::string const& term)
+    [[nodiscard]] virtual auto find(std::string const& term) const
```
JMMackenzie, line 55: As above!

test/in_memory_index.cpp
```diff
@@ -8,7 +8,7 @@ auto VectorCursor::docid() const noexcept -> std::uint32_t {
     return documents[0];
 }
 
-auto VectorCursor::freq() const noexcept -> float {
+auto VectorCursor::freq() const noexcept -> std::uint32_t {
```
JMMackenzie, line 11: This was a float before? :D

JMMackenzie (approved):
This looks great! A few repeated comments about API changes throughout.

test/in_memory_index.cpp
elshize: :shrug:

include/pisa/term_map.hpp
elshize: Added a `Changelog-changed:` trailer to the git message.

PR #585, opened 2024-07-18
Feature/fix test

oxnz:
Fixes # .

-
-
-

### DIFF
--- a/include/pisa/mappable/mappable_vector.hpp
+++ b/include/pisa/mappable/mappable_vector.hpp
@@ -2,13 +2,9 @@
 
 #include <algorithm>
 #include <vector>
-
-#include "boost/function.hpp"
-#include "boost/lambda/bind.hpp"
 #include "boost/lambda/construct.hpp"
 #include "boost/range.hpp"
 #include "boost/utility.hpp"
-
 #include "util/intrinsics.hpp"
 
 namespace pisa { namespace mapper {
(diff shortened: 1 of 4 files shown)


elshize:
Hi, could you clarify in the PR description what this change is addressing?

oxnz:
> Hi, could you clarify in the PR description what this change is addressing?

sry for the disturb, I think I have opened the pull request too soon before it's ready for review.

to be short, I want to make PISA runnable on other platforms. and this might take some time. the PR is at your disposal. if this objective align with you, then I'll update the PR soon.

currently I'm having trouble in building this repo on my Laptop (Mac OS, Arm64 CPU). 
a lot of errors pops up.

just list some majors ones:
1. intrinsics header missing (x86 vs arm64) (I've commented out some of the unsupported codecs, got compiled finally with a lot of functionality missing)
2. CXX standard problems, I have to change `boost::function` to `std::function` to get this work, seems boost is a bit out of date or sth.
3. some unit tests failure. e.g.: `quantize(max) == (1<<bits) - 1` stuff, seams different behavior for x86 cpu v.s. arm cpu, pending further investigation

elshize:
@oxnz sounds great, we're stretched a bit thin, so these type of things often are left unresolved, so this kind of effort is always appreciated.

Re: boost, I am aware that the version we bring from a submodule is old, and that's because the repo we've been using is not updated. Try installing boost system-wide and configuring cmake with `-DPISA_SYSTEM_BOOST=ON`, and recompiling.

Re: bit stuff, there _might_ be some places that assume x86 architecture in certain places, mostly legacy code. If that's the case, it may take some effort to do that. If so, I would appreciate if you could maybe create a separate PR for each problem instead of a general "fixes" PR. E.g., submit missing header issue separately from x86 bit stuff. I'd like to keep it manageable but also I'm trying to be more mindful of keeping git history cleaner.

Issue #242, opened 2019-08-23
libtbb.so.2 is not installed by make install


binshengliu:
**Describe the bug**
`libtbb.so.2` is not installed by `make install`

**To Reproduce**
1. `cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME/.local/ `
2. `make && make install`
3. `parse_collection --help`

**Error message**
```
$ parse_collection --help
parse_collection: error while loading shared libraries: libtbb.so.2: cannot open shared object file: No such file or directory
```
Manually cp libtbb.so.2 solves it.
```
$ cp ./external/tbb_cmake_build/tbb_cmake_build_subdir_release/libtbb.so.2 ~/.local/lib/libtbb.so.2
$ parse_collection --help
parse_collection - parse collection and store as forward index.
Usage: parse_collection [OPTIONS] [SUBCOMMAND]
```
**Expected behavior**
Program runs.

**Environment info**
Operating System: Red Hat
Compiler name: Clang
Compiler version: `clang version 8.0.0 (tags/RELEASE_800/final)`

elshize:
The problem is we don't define `install` at all in our `CMakeLists.txt`. There are two things I see missing that we must do:

- define library installation rule(s) to have source/lib/tools installed,
- install `cmake` targets so that they can be found by `find_package`.

Our `wapopp` most likely does it right: https://github.com/pisa-engine/wapopp

Also as a reference: https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html

lgrz:
Can confirm this when installing into a "non-system" location (a path not listed in `/etc/ld.so.conf.d/*` or similar)

Maybe a first (or easy) step is to add an RPATH entry to the binaries when cmake's `make install` is run, but only when a non-standard install path is given.

So something like the following?

```
if(NOT CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
endif()
```

Here is the output of ldd without CMAKE_INSTALL_RPATH:

```
$ ldd bin/parse_collection 
        linux-vdso.so.1 (0x00007ffc33bc6000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f98dab8e000)
==>     libtbb.so.12 => not found
[...]

PR #588, opened 2024-11-29
Set RPATH for install target

lgrz:
Incremental fix for #242, set the runtime path so that the cli tools can find libtbb when building PISA with the libtbb submodule which is the default option.

- Set RPATH
- Add install step to the cli test build
- Typo fix for benchmark option


elshize (approved):
Great stuff, thanks!

PR #590, opened 2024-12-09
Remove C++17 support

elshize:
The CMake parameters for C++ standard and concepts are removed, and the
code is updated accordingly. We use `[[likely]]` and `[[unlikely]]`
instead of a macro. We also remove the macros for concepts, and use
unconditional `static_assert` and `require`.


JMMackenzie (approved):
:two: :zero: :rocket:

PR #593, opened 2024-12-11
Remove GSL in favor of stdlib

elshize:
We mostly used GSL for its `span` class. After moving to the C++20 standard, we now have `std::span` available to us. It is more explicit in how it works because it's driven by the standard, and it allows us not to rely on a third party library.

`std::span` does no bound checks, so indexed access had to be guarded with explicit checks when necessary. Some helper functions were introduced, including equality operator.

`Expect` macros were replaced with `if` statements that throw `std::invalid_argument` if the contract is not upheld.

Changelog-changed: GSL is removed as dependency
Changelog-changed: gsl::span is replaced with std::span

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,6 @@ target_link_libraries(pisa
         Boost::boost
         mio
         mio_base
-        GSL
         spdlog
         fmt::fmt
         range-v3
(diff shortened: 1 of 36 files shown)


JMMackenzie (approved):
Nice, this is great!

elshize:
Before merging this, I'll add missing tests for the code I added.

PR #596, opened 2024-12-18
Lookup table

elshize:
This lookup table implementation is meant to replace the current lexicon
data structure. The overall concept is the same; the aim here is to
improve the binary format to allow for extensions, and to improve the
interaction with the table from the code, as well as naming convention.

This change includes the data structure implementing the mapping
concept, a CLI tool to build a table and read data from it, and a unit
test suite, as well as CLI tests.

In this particular change, we do not introduce any breaking changes. The
new code is not used in the already existing tools and workflows. This
work will be done in the future.


JMMackenzie (approved):
Nice, this looks great!

PR #600, opened 2025-02-15
Use tiny collection for compression test

### LINKED ISSUE #577: Clang 16 (c++20) build in CI fails
Clang 16 (c++20) build in CI fails:

https://github.com/pisa-engine/pisa/actions/runs/7865449374/job/21458386783

```
65.16 -- Compiling and running to test HAVE_POSIX_REGEX
65.20 -- Performing Test HAVE_POSIX_REGEX -- failed to compile
```

elshize:
The compression test runs for each encoding to verify that collection builds without an error. Additionally, we run a test with sanitizers enabled, which can find memory issues, etc. This tests is very slow, however, so making it a smaller collection helps with the timeouts.

Fixes: https://github.com/pisa-engine/pisa/issues/577


lgrz (approved):
Thanks, the changes are good to go.

PR #601, opened 2025-02-21
Update compiler version support

elshize:
Build CI with GCC 14 (in addition to 10-13), and Clang 16-19 (dropping
15).


docs/book/cli/compress_inverted_index.html
```diff
@@ -173,34 +173,7 @@ <h1 class="menu-title">PISA Guide</h1>
                     <main>
                         <h1 id="compress_inverted_index"><a class="header" href="#compress_inverted_index"><code>compress_inverted_index</code></a></h1>
 <h2 id="usage"><a class="header" href="#usage">Usage</a></h2>
-<pre><code>Compresses an inverted index
```
lgrz: The preprocess command did not run on the CLI tools during mdbook build for some reason.
https://github.com/pisa-engine/pisa/blob/5dab3286af24c7665ba67a2c3491589a6120a325/docs/src/cli/compress_inverted_index.md?plain=1#L6

elshize: Good catch, maybe some misconfiguration  in my local env. I'll look into it.

elshize: Should be good now.

lgrz (approved):
Thanks, the changes are good

PR #602, opened 2025-02-23
CI: upgrade compiler for Clang Tidy

elshize:
This should fix CI check for clang-tidy on `main`.


lgrz (approved):
These are good, thanks!

PR #603, opened 2025-03-08
Feature: JSON document format

elshize:
Support JSONL collection format. This should make it easy to prepare a script that transforms any format into JSON and streams it to the parsing tool.


JMMackenzie (approved):
This is great, nice work!

docs/book/guide/parsing.html
```diff
@@ -262,6 +262,8 @@ <h2 id="supported-formats"><a class="header" href="#supported-formats">Supported
 <li><code>plaintext</code>: every line contains the document's title first, then any
 number of whitespaces, followed by the content delimited by a new line
 character.</li>
+<li><code>jsonl</code>: every line is a JSON document with three fields: <code>title</code>,
```
amallia, line 265: Should we make title optional?

docs/book/print.html
```diff
@@ -420,6 +420,8 @@ <h2 id="supported-formats"><a class="header" href="#supported-formats">Supported
 <li><code>plaintext</code>: every line contains the document's title first, then any
 number of whitespaces, followed by the content delimited by a new line
 character.</li>
+<li><code>jsonl</code>: every line is a JSON document with three fields: <code>title</code>,
```
amallia, line 423: same here

script/ir-datasets.py
```diff
@@ -0,0 +1,35 @@
...
+def _parser():
+    parser = argparse.ArgumentParser("ir-datasets")
+    parser.add_argument("dataset")
+    return parser
+
+
+def parse_dataset(dataset_name):
+    dataset = ir.load(dataset_name)
+    for doc in dataset.docs_iter():
+        print(json.dumps({"title": doc.doc_id, "content": doc.text}))
```
amallia, line 20: why doc_id as title?
For example 
```    
import ir_datasets
dataset = ir_datasets.load("beir/climate-fever")
for doc in dataset.docs_iter():
    doc # namedtuple<doc_id, text, title>
```

```
import ir_datasets
dataset = ir_datasets.load("msmarco-passage")
for doc in dataset.docs_iter():
    doc # namedtuple<doc_id, text>
```


Two examples: one has title and the other one does not

docs/book/guide/parsing.html
amallia: Can we clarify how we treat title and content? Do we concatenate?

elshize: "title" what in PISA we call a string document ID. We must have one. This has nothing to do with a title that could be part of content.

elshize: I'll try to clarify

PR #616, opened 2025-07-04
Tutorial improvements

elshize:


test/docker/tutorial/instructions.md
```diff
@@ -1,22 +1,54 @@
...
+## Prerequisites
+
+### Download input data
+
+TODO
+
+### Download container image
+
+First,
+[download the image](https://sigir2025.it-mil-1.linodeobjects.com/pisa-tutorial.tar.gz).
```
JMMackenzie, line 12: Where is this hosted currently? I need to update this once the image is finalized, correct?

JMMackenzie (approved):
Looks good to me!

test/docker/tutorial/instructions.md
elshize: This is in my S3 bucket. You'll need to probably build and host it somewhere else. I'll make sure to upload the newest version now, please test it. If you're not making any changes to the code, then you should be able to get away with not re-uploading. Note that instructions won't be included in the image anyway, so you only need to rebuild if you're changing code (or image itself, say, you want to pre-install something extra)

PR #620, opened 2025-08-04
Scorer option for `compute_intersection`

### LINKED ISSUE #392: Use --scorer for compute_intersection
Right now, the scorer is hard-coded to `bm25`.

lgrz:
Add CLI option to customize scorer params for BM25 and beyond.
Fixes #392

- Add `--scorer` option to tools/compute_intersection.cpp
- Update tests/test_intersection.cpp
- Fix logic error when filtering queries by min/max number of terms

### DIFF
--- a/tools/app.hpp
+++ b/tools/app.hpp
@@ -8,6 +8,8 @@
 #include <thread>
 
 #include <CLI/CLI.hpp>
+#include <range/v3/range/conversion.hpp>
+#include <range/v3/view/filter.hpp>
 #include <range/v3/view/getlines.hpp>
 #include <range/v3/view/transform.hpp>
 #include <spdlog/spdlog.h>
@@ -134,6 +136,7 @@ namespace arg {
             } else {
                 io::for_each_line(std::cin, parse_query);
             }
+
             return qs;
         }
 
@@ -153,6 +156,42 @@ namespace arg {
         bool m_weighted = false;
         std::optional<std::string> m_term_lexicon{std::nullopt};
         CLI::Option* m_terms_option{};
+    };
+
+    /**
+     * Post query parsing filter. There is only one required filter at present
+     * in `tools/compute_intersection`. Therefore the implementation is
+     * specific to that particular use case (filter by query length).
+     */
+    struct QueryFilter {
+        explicit QueryFilter(CLI::App* app)
+            : m_min_length(1), m_max_length(std::numeric_limits<int>::max()) {
+            app->add_option("--min-query-len", m_min_length, "Minimum query length");
+            app->add_option("--max-query-len", m_max_length, "Maximum query length");
+        }
+
+        [[nodiscard]] auto query_filter_min_length() const -> int { return m_min_length; }
+        [[nodiscard]] auto query_filter_max_length() const -> int { return m_max_length; }
+
+        /**
+         * Takes the (parsed) queries and filters them according to the length
+         * constraints.
+         */
+        [[nodiscard]] auto query_filter_apply(std::vector<::pisa::Query> const& qs) const
+            -> std::vector<::pisa::Query> {
+            return ranges::views::filter(
+                       qs,
+                       [&](auto&& query) {
+                           auto n_terms = query.terms().size();
+                           return m_min_length <= n_terms && n_terms <= m_max_length;
+                       }
+                   )
+                | ranges::to<std::vector>();
+        }
+
+      private:
+        int m_min_length;
+        int m_max_length;
     };
 
     struct Algorithm {
--- a/tools/compute_intersection.cpp
+++ b/tools/compute_intersection.cpp
@@ -80,12 +80,15 @@ int main(int argc, const char** argv) {
     spdlog::set_default_logger(spdlog::stderr_color_mt(""));
 
     std::optional<std::uint8_t> max_term_count;
-    std::size_t min_query_len = 0;
-    std::size_t max_query_len = std::numeric_limits<std::size_t>::max();
     bool combinations = false;
     bool header = false;
 
-    App<arg::Index, arg::WandData<arg::WandMode::Required>, arg::Query<arg::QueryMode::Unranked>, arg::Scorer, arg::LogLevel>
+    App<arg::Index,
+        arg::WandData<arg::WandMode::Required>,
+        arg::Query<arg::QueryMode::Unranked>,
+        arg::QueryFilter,
+        arg::Scorer,
+        arg::LogLevel>
         app{"Computes intersections of posting lists."};
     auto* combinations_flag = app.add_flag(
         "--combinations", combinations, "Compute intersections for combinations of terms in query"
@@ -96,18 +99,12 @@ int main(int argc, const char** argv) {
            "Max number of terms when computing combinations"
     )
         ->needs(combinations_flag);
-    app.add_option("--min-query-len", min_query_len, "Minimum query length");
-    app.add_option("--max-query-len", max_query_len, "Maximum query length");
     app.add_flag("--header", header, "Write TSV header");
     CLI11_PARSE(app, argc, argv);
 
     spdlog::set_level(app.log_level());
 
-    auto queries = app.queries();
-    auto filtered_queries = ranges::views::filter(queries, [&](auto&& query) {
-        auto size = query.terms().size();
-        return min_query_len <= size && size <= max_query_len;
-    });
+    auto filtered_queries = app.query_filter_apply(app.queries());
 
     if (header) {
         if (combinations) {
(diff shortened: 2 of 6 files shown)


tools/compute_intersection.cpp
```diff
@@ -104,7 +106,7 @@ int main(int argc, const char** argv) {
     auto queries = app.queries();
     auto filtered_queries = ranges::views::filter(queries, [&](auto&& query) {
         auto size = query.terms().size();
-        return size < min_query_len || size > max_query_len;
+        return size > min_query_len || size < max_query_len;
```
elshize: Shouldn't it be `min_query_len <= size && size <= max_query_len`? :thinking:

lgrz: Yes, thanks that is a more readable way to put it and that it should include the boundary min/max. Thanks!

elshize: Well, that plus your initial fix still used `&&`. Just confirming: this function should return true if size is in `[min_query_len, max_query_len]` range (inclusive), right? I'm double checking because it's weird it was the opposite before, seems like a very big oversight, so I'm second guessing myself on this... But yeah, it should be right now I think.

lgrz: Yes, it was odd that it was the opposite before. And I second guessed myself a couple of times, but then verified by way of that the change in this PR was the only way to get the intersections printed out when running the the tool. We don't have tests in place for this tool similar to `test/cli/*`, but maybe we (or I) should have some there.

lgrz:
Also, just to add some output for reference. This output is with the new min/max change (in this PR):

```
== without min, max arguments
$ ./build/debug/bin/compute_intersection --encoding block_simdbp --index 100k_block_simdbp.idx --wand 100kwand_fixed40_bm25 --tokenizer english --token-filters lowercase porter2 --terms 100kfwd.termlex --queries dl20.txt --header --scorer dph
qid     length  max_score
1108939 0       0
1112389 0       0
792752  4000    6.984559
...

== with arguments min: 0, max: 3
$ ./build/debug/bin/compute_intersection --encoding block_simdbp --index 100k_block_simdbp.idx --wand 100kwand_fixed40_bm25 --tokenizer english --token-filters lowercase porter2 --terms 100kfwd.termlex --queries dl20.txt --header --scorer dph --min-query-len 0 --max-query-len 3
qid     length  max_score
792752  4000    6.984559
1128373 2402    5.70508
1124979 1       7.8003674
...

== with arguments min: 3, max: 3
$ ./build/debug/bin/compute_intersection --encoding block_simdbp --index 100k_block_simdbp.idx --wand 100kwand_fixed40_bm25 --tokenizer english --token-filters lowercase porter2 --terms 100kfwd.termlex --queries dl20.txt --header --scorer dph --min-query-len 3 --max-query-len 3
qid     length  max_score
1124979 1       7.8003674
156493  1       13.767251
1124464 0       0
...

== with invalid arguments min: 1, max: 0
$ ./build/debug/bin/compute_intersection --encoding block_simdbp --index 100k_block_simdbp.idx --wand 100kwand_fixed40_bm25 --tokenizer english --token-filters lowercase porter2 --terms 100kfwd.termlex --queries dl20.txt --header --scorer dph --min-query-len 1 --max-query-len 0
qid     length  max_score
```

tools/compute_intersection.cpp
elshize: Ideally, this wouldn't be implemented directly in the tool source file but rather in the library and then unit tested. Perhaps subtype `arg::Query` and have it deal with filtering query? Not sure to be honest. I also wouldn't say it's a hard requirement.

If you have a spare moment to write CLI tests, it would be great. But I won't block this for that, it's good to have this fix in.

elshize (approved):

tools/compute_intersection.cpp
lgrz: Yes, I agree. While working on this PR I did try to look at moving the
min/max query length filter into `include/pisa/intersect.hpp`. But I didn't
get to a good solution becuase of the way `for_all_subsets` works for the
`--combinations` flag and the `--max-term-count` filter.

To me a good solution would result in a reasonable design that works well with
`for_all_subsets` and for `Intersection::compute`, where either of those
intersection modes _may_ have the `min/max` query length filter.
Then in `tool/compute_intersection` there would be no need to pass a
function argument `print_intersection` which does more than only printing
the intersection (it does the intersection as well as print to stdout).

Also, I did *not* consider other ways of solving this outside of
`pisa/include/intersection.hpp`. Thanks, this is a good suggestion and
`Args:Query` that sounds like a good place to start looking (because it is
more along the lines of a `TermPolicy` but at the query level).

This could make sense for generating sub-queries too as it is a query
filter in the opposite direction. But then again if it is not used outside
of `compute_intersection` then probably best to keep that local to the
intersection implementation as it may have additional complications for
example the query ID and so forth.

I would prefer to try and resolve the `min/max` query filter with unit
tests rather than the CLI tests. So I won't add the CLI tests for now.

lgrz:
Updated the PR and am looking for feedback on the new changes introduced.
To summarize, added `arg::QueryFilter` to handle the filtering of queries
for `compute_intersection`; opted to _not_ inherit from `arg::Query` since
it is already a child of `arg::Analyzer`. This is why when using
`QueryFilter` the "user" must explicitly call `query_filter_add_options`
and then apply the filter using the output of `app.queries`.

Added tests in `tools/tests/test_app.cpp`, but I realized that they are
outdated and not used in builds or gh actions.

Any suggestions on whether to continue along this path using
`tools/tests/test_app.cpp`?

Another option might be if they are useful is to rework them under the main
test suite.

tools/app.hpp
```diff
@@ -155,6 +158,44 @@ namespace arg {
...
+                           return m_min_length <= n_terms && n_terms <= m_max_length;
+                       }
+                   )
+                | ranges::to<std::vector>();
+        }
+
+      private:
+        int m_min_length = 1;
+        int m_max_length = std::numeric_limits<int>::max();
+        CLI::App* m_app = nullptr;
```
elshize: I think I'd rather not set it to `nullptr`. The reason is that it may suggest that it may not have a set value, but it's not true because the only constructor sets it.

Are there any consequences of not having one variable set here? I don't think so, but I might be wrong. For consistency, we can also not set any of the values here and set the min/max in the constructor as well. WDYT?

elshize (approved):
I think the app tests _could_ be implemented as part of the regular test directory, but they need to be still separate to an extent from the rest. We probably don't want to link app to the lib tests and vice versa. It doesn't need to be done here though.

tools/app.hpp
lgrz: Thanks for the feedback, sure can move these to the constructor.

The member initializer way is sort of a habbit as as code gets changed later on (for example [C.48](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer)). But then again the usecase here is quite specific and `arg` doesn't change that often. And probably more importantly it moving to the constructor will make it consistent with existing code. (Perhaps I will also move `query_filter_add_options` to the constructor too; then there is no need for a member `m_app`.

lgrz:
> I think the app tests could be implemented as part of the regular test
> directory, but they need to be still separate to an extent from the rest.
> We probably don't want to link app to the lib tests and vice versa. It
> doesn't need to be done here though.

Ok yes, thanks for the tips on how it should work if it ends up in `/tests`.

The reason I was asking the above is because currently there is no CI/Build
that has both `PISA_BUILD_TOOLS=ON` and `PISA_ENABLE_TESTING=ON`, which
at present is required for the tests in `tools/tests` to be automated
on build. And I _think_ that possibly the tests under tools only depend on
`CLI.hpp, app.cpp, app.h` and do not depend on the individual tools under
`tools` directory. So it may be useful to lift the app tests out and into
the main test directory.

I will try leaving the app tests where they are for now and maybe it is
just a matter of having an extra build config in the CI for the app tests.
Does that sound OK?

tools/app.hpp
elshize: I have no strong preference as to where to initialize, I just prefer not to initialize anything with a value if it's always overridden in the constructor, because I would want to avoid a situation where we accidentally remove the initialization from the constructor and we end up with explicit nullptr, and static analysis won't detect an issue there.

elshize:
This sounds good. I'm fine moving it to test directory as well. Whatever is easier here, and we can address it in a later PR. We don't have to fix everything here.

tools/app.hpp
lgrz: Sure, thanks for the context on asan; will update the patch

lgrz:
> This sounds good. I'm fine moving it to test directory as well. Whatever is easier here, and we can address it in a later PR. We don't have to fix everything here.

Thanks, have left the tests under `tools/test` for now and noted that they can be resolved at a later stage.

elshize:
Sorry for the late reply, I've been traveling. Just double-checking: this is ready to merge, correct? Or did you plan any more changes?

lgrz:
Yes, this should be good to merge now. No further changes for this one. Thanks!

PR #621, opened 2025-08-05
`tools/compute_intersection`: do not print empty intersections

### LINKED ISSUE #290: Don't print out empty intersections
Currently, `compute_intersection` with `--combinations` prints out all combinations, even empty ones. We should add a condition that doesn't print out an intersection if `intersection.length == 0` because it means effectively it doesn't exist, and it could create problems later on.

lgrz:
This changes `tools/compute_intersection` so that only those queries (with or without the `--combinations` flag, and with or without a `Mask` in `intersection::compute`) with an intersection size greater than zero will be output.

Note, fix typo in `include/intersect.hpp`

Fixes #290


lgrz:
This patch will be updated with the changes from #620 when that one is merged all OK.

lgrz:
Updated to latest changes on `main` and ready for review/merge

elshize (approved):

PR #623, opened 2025-10-24
Update `sample_inverted_index.md`

gustingonzalez:
Add a concise description of `sample_inverted_index`, including its purpose, sampling strategies, and examples.

### DIFF
--- a/docs/src/cli/sample_inverted_index.md
+++ b/docs/src/cli/sample_inverted_index.md
@@ -5,3 +5,22 @@
 ```
 <!-- cmdrun ../../../build/bin/sample_inverted_index --help -->
 ```
+
+## Description
+Creates a smaller inverted index from an existing one by sampling postings or documents. The purpose of this tool is to reduce time and space requirements while preserving the main statistical properties of the original collection, making it useful for faster experiments and debugging.
+
+### Sampling strategy (`-t, --type`)
+- `random_postings`: keep random occurrences per posting list (not whole posting lists).
+- `random_docids`: keep all postings belonging to a random subset of documents.
+
+## Examples
+
+### Keep ~10% of postings
+```bash
+sample_inverted_index -c path/to/inverted -o path/to/inverted.sampled -r 0.1 -t random_postings
+```
+
+### Keep ~10% of the documents
+```bash
+sample_inverted_index -c path/to/inverted -o path/to/inverted.sampled -r 0.25 -t random_docids
+```
--- a/docs/src/cli/sample_inverted_index.md
+++ b/docs/src/cli/sample_inverted_index.md
@@ -7,21 +7,36 @@
 ```
 
 ## Description
-Creates a smaller inverted index from an existing one by sampling postings or documents. The purpose of this tool is to reduce time and space requirements while preserving the main statistical properties of the original collection, making it useful for faster experiments and debugging.
+
+Creates a smaller inverted index from an existing one by sampling postings or
+documents. The purpose of this tool is to reduce time and space requirements
+while preserving the main statistical properties of the original collection,
+making it useful for faster experiments and debugging.
 
 ### Sampling strategy (`-t, --type`)
 
-- `random_postings`: keep random occurrences per posting list (not whole posting lists).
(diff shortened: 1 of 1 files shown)


docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
 ```
 <!-- cmdrun ../../../build/bin/sample_inverted_index --help -->
 ```
+
+## Description
```
elshize, line 9: Let's add an empty line between the header and the paragraph.

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+Creates a smaller inverted index from an existing one by sampling postings or documents. The purpose of this tool is to reduce time and space requirements while preserving the main statistical properties of the original collection, making it useful for faster experiments and debugging.
```
elshize: Let's keep the text width to 80 characters. I suggest using some kind of markdown plugin in your editor. For example, in vim you can use `gww` to format the current line or `gwip` to format an entire paragraph (if it's already broken down into lines but you edited some text and want to fix the formatting without rearranging all of it by hand).

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+
+### Sampling strategy (`-t, --type`)
+
+- `random_postings`: keep random occurrences per posting list (not whole posting lists).
+- `random_docids`: keep all postings belonging to a random subset of documents.
```
elshize: Let's keep the width to 80 here as well.

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+
+## Examples
+
+### Keep ~25% of postings
```
elshize, line 24: Newline below.

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+```bash
+sample_inverted_index -c path/to/inverted -o path/to/inverted.sampled -r 0.25 -t random_postings
+```
+
+### Keep ~25% of the documents
```
elshize, line 34: Newline below.

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+### Sampling strategy (`-t, --type`)
+
+- `random_postings`: keep random occurrences per posting list (not whole posting lists).
+- `random_docids`: keep all postings belonging to a random subset of documents.
+
+## Examples
+
+### Keep ~25% of postings
+```bash
+sample_inverted_index -c path/to/inverted -o path/to/inverted.sampled -r 0.25 -t random_postings
```
elshize: In a way, it is even more important to keep the width small in code because it won't be formatted, so you'll need to scroll horizontally and that's never a good UX.

So I would break it down like so:

```bash
first line that is under 80 characters \
    continue after backslash
```

By adding backslash, you make it possible to copy and paste the command.

But seeing this is documentation, we may want to highlight each parameter or at least the important ones, perhaps even with a comment if necessary (within reason).

```bash
sample_inverted_index \
    -c path/to/inverted \
    -o path/to/inverted.sampled \
    -r 0.25 \
    -t random_postings
```

Now we can see each argument clearly without having to parse a long line of text.

docs/src/cli/sample_inverted_index.md
```diff
@@ -5,3 +5,23 @@
...
+```
+
+### Keep ~25% of the documents
+```bash
+sample_inverted_index -c path/to/inverted -o path/to/inverted.sampled -r 0.25 -t random_docids
```
elshize: Same as the above command example.

elshize (changes requested):
Nice, just have some formatting nitpicks.

gustingonzalez:
Hi @elshize, thanks for your review. All done!

elshize (approved):

elshize:
Thank you for the contribution @gustingonzalez :fire:

PR #625, opened 2025-12-28
Update `invert` tool documentation

### LINKED ISSUE #619: Tool invert argument `--term-count` should be a required argument
The documentation for invert mentions that the term count argument is required. But this is not the case in the actual tool arguments in `tools/invert.cpp`. Need to confirm which one it should be.

Docs.
https://github.com/pisa-engine/pisa/blob/07d63477462e68cfeb18f624278df545519284c4/docs/src/guide/inverting.md?plain=1#L12

tools/invert.
```
$ ./build/debug/bin/invert --help
Constructs an inverted index from a forward index.
Usage: ../../pisa/build/debug/bin/invert [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  -i,--input TEXT REQUIRED    Forward index basename
  -o,--output TEXT REQUIRED   Output inverted index basename
  --term-count UINT           Number of distinct terms in the forward index
  -j,--threads UINT           Number of threads
  --batch-size UINT [100000]  Number of documents to process at a time
  -L,--log-level TEXT:{critical,debug,err,info,off,trace,warn} [info] 
                              Log level
  --config                    Configuration .ini file
```

lgrz:
Update the documentation for `invert` tool to be inline with the output of the forward index from `parse_collection`.

Updates to CLI and Guide documentation for `invert` are provided.

Fixes #619

### DIFF
--- a/docs/src/guide/inverting.md
+++ b/docs/src/guide/inverting.md
@@ -6,12 +6,15 @@ existence of a forward index in the path `path/to/forward/cw09b`:
 
     $ mkdir -p path/to/inverted
     $ ./invert -i path/to/forward/cw09b \
-        -o path/to/inverted/cw09b \
-        --term-count `wc -w < path/to/forward/cw09b.terms`
+        -o path/to/inverted/cw09b
 
-Note that the script requires as parameter the number of terms to be
-indexed, which is obtained by embedding the
-`wc -w < path/to/forward/cw09b.terms` instruction.
+Inverting an index requires the knowledge of the number of terms in
+the lexicon ahead of time. In the above example, the `invert` command
+assumes that a `cw09b.termlex' file exists from the output of
+`parse_collection` which is used to lookup the term count.
+
+Note that the number of terms can be provided using `--term-count` in
+case the lexicon is not available or on a different path.
 
 ## Inverted index format
 
--- a/tools/app.cpp
+++ b/tools/app.cpp
@@ -171,7 +171,13 @@ auto Threads::print_args(std::ostream& os) const -> std::ostream& {
 Invert::Invert(CLI::App* app) {
     app->add_option("-i,--input", m_input_basename, "Forward index basename")->required();
     app->add_option("-o,--output", m_output_basename, "Output inverted index basename")->required();
-    app->add_option("--term-count", m_term_count, "Number of distinct terms in the forward index");
+    app->add_option(
+        "--term-count",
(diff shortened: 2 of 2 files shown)


docs/src/guide/inverting.md
```diff
@@ -6,12 +6,16 @@ existence of a forward index in the path `path/to/forward/cw09b`:
...
     $ mkdir -p path/to/inverted
     $ ./invert -i path/to/forward/cw09b \
-        -o path/to/inverted/cw09b \
-        --term-count `wc -w < path/to/forward/cw09b.terms`
+        -o path/to/inverted/cw09b
 
-Note that the script requires as parameter the number of terms to be
-indexed, which is obtained by embedding the
-`wc -w < path/to/forward/cw09b.terms` instruction.
+Note that the `invert` command requires the number of terms in the lexicon
```
elshize: I would reword this first sentence now because it reads as if `invert` required the num of terms as argument.

Example (feel free to edit):

> Inverting an index requires the knowledge of the number of terms in the lexicon ahead of time. In the above example, [...]

docs/src/guide/inverting.md
```diff
@@ -6,12 +6,16 @@ existence of a forward index in the path `path/to/forward/cw09b`:
...
+(or vocabulary) when building an inverted index. In the above example, the
+`invert` program assumes that a `cw09b.termlex' file exists from the
+output of `parse_collection` which is used to lookup the term count.
+
+The `invert` command also has a legacy option where the term count can be
```
elshize: Instead of calling it a legacy option, I would simply mention that the number of terms can be provided using `--term-count` in case the lexicon is not available or on a different path.

In the future, we should try to standardize paths but for now, lots of our CLI tools don't make assumptions on what paths are. It's a bit of a mess unfortunately.

Not sure if you've seen it, but we wrote this [CLI wrapper](https://github.com/pisa-engine/pisa/tree/main/pisactl) for the last SIGIR tutorial, which is an POC of how this may look, but it's not necessarily the final solution (written in Python for now but ultimately we would like to improve the native binaries instead).

lgrz:
Thanks for the feedback! Have updated based the comments above.

docs/src/guide/inverting.md
lgrz: Thanks for the pointer on the CLI wrapper :)

elshize (approved):

PR #626, opened 2026-01-03
Build docs

lgrz:
Rebuild the docs from HEAD which was overlooked in #619


elshize (approved):

elshize:
Thanks, my bad for not catching it. I think it would be nice to have a CI check for that, so I opened https://github.com/pisa-engine/pisa/issues/627 if it's something you'd like to work on perhaps, shouldn't be too difficult.

PR #630, opened 2026-01-31
fix: Use backward-compatible default CLI value

elshize:
Use `capture_default_str()` instead of `default_value()` for backward compatibility.

Using `default_value()` failed on one of the tests on `main` that used system dependency for CLI11 that was in a lower version.

---

Also deleting operators that are already implicitly deleted.

### DIFF
--- a/include/pisa/util/util.hpp
+++ b/include/pisa/util/util.hpp
@@ -113,24 +113,23 @@ function_iterator<State, AdvanceFunctor, ValueFunctor> make_function_iterator(
 }
 
 struct stats_line {
-    stats_line() : m_out(std::cout) { m_out << "{"; }
-    explicit stats_line(std::ostream& out) : m_out(out) { m_out << "{"; }
+    stats_line() { std::cout << "{"; }
     stats_line(stats_line const&) = default;
     stats_line(stats_line&&) noexcept = default;
     stats_line& operator=(stats_line const&) = default;
     stats_line& operator=(stats_line&&) noexcept = default;
-    ~stats_line() { m_out << "}" << std::endl; }
+    ~stats_line() { std::cout << "}" << std::endl; }
 
     template <typename K, typename T>
     stats_line& operator()(K const& key, T const& value) {
         if (!first) {
-            m_out << ", ";
+            std::cout << ", ";
         } else {
             first = false;
         }
 
         emit(key);
-        m_out << ": ";
+        std::cout << ": ";
         emit(value);
         return *this;
     }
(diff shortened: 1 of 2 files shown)


include/pisa/util/util.hpp
```diff
@@ -115,10 +115,10 @@ function_iterator<State, AdvanceFunctor, ValueFunctor> make_function_iterator(
 struct stats_line {
     stats_line() : m_out(std::cout) { m_out << "{"; }
     explicit stats_line(std::ostream& out) : m_out(out) { m_out << "{"; }
-    stats_line(stats_line const&) = default;
-    stats_line(stats_line&&) noexcept = default;
-    stats_line& operator=(stats_line const&) = default;
-    stats_line& operator=(stats_line&&) noexcept = default;
+    stats_line(stats_line const&) = delete;
```
gustingonzalez: Why is it better in this use case to delete these operations directly instead of rely on the compiler implicit deletion?

elshize: Good question. First, let me set up the scene: I'm switching from `default` to `delete` because once we have a reference, the compiler will implicitly delete these operators but nevertheless successfully compile. It still works because we don't use these, but it is confusing that you see `= default` when it is not true. Clang-Tidy detects that, however, and fails on `main`.

Now coming back to your question, which, if I understand correctly, is: why not just let the compiler do its job and remove those altogether? This is we're following the [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three.html). I encourage you to read the article (many more articles available online as well), but the short of it is: if you define a constructor, it's good practice to always define copy constructor/operator as well (and move constructor/operator if you want move semantics). The implicitly defined constructors and operators that can be surprising.

In my opinion, explicitly deleting one or many of them is an extension of this rule. I think it comes down to how easily you can reason about a class when looking at its definition. The rules on generating implicit definitions (or deleting them) are complex, so why not be explicit about it, so that the reader can quickly realize it, even if they don't know the rules?

---

Side note: I noticed that passing the reference is no longer needed because we didn't end up using this class this way. I'll revert it to what it was before because I don't like how it owns a reference. We could improve it but (read below)...

---

I think `stats_line` is terrible in general, so I might just want to rewrite this. There are several reasons why this is bad. If you read the linked article to the end, you will notice that this actually violates the rule of zero, because we don't manage any resources in the destructor. Also the way it is called with `operator()` is really weird and confusing. It's not thread safe as well. In general, unnecessarily complicated, I would say, especially for something we use in like maybe 3-4 places.

elshize: Made the change to revert `stats_line` to its previous state.

gustingonzalez: Great explanation, thanks! This change in particular now makes sense for me, so the PR looks good to me.

gustingonzalez (approved):

PR #631, opened 2026-02-01
chore: Refactor stats_line into StatsBuilder

elshize:


include/pisa/util/stats_builder.hpp
```diff
@@ -0,0 +1,77 @@
...
+concept Streamable = requires(std::ostream& os, T value) {
+    { os << value } -> std::convertible_to<std::ostream&>;
+};
+
+/**
+ * Builds a simple key-value JSON for printing statistics.
+ */
+class StatsBuilder {
+  private:
+    std::unique_ptr<::pisa::detail::StatsBuilderInterface> m_impl;
```
elshize: Separating implementation details to the source file so that we can keep nlohmann dependency private.

src/util/stats_builder.cpp
```diff
@@ -0,0 +1,81 @@
...
+    nlohmann::json m_json;
+
+  public:
+    virtual void add(std::string const& key, bool value) { m_json[key] = value; }
+    virtual void add(std::string const& key, std::int64_t value) { m_json[key] = value; }
+    virtual void add(std::string const& key, std::uint64_t value) { m_json[key] = value; }
+    virtual void add(std::string const& key, double value) { m_json[key] = value; }
+    virtual void add(std::string const& key, std::string value) { m_json[key] = std::move(value); }
+    virtual void add(std::string const& key, const char* value) { m_json[key] = value; }
+    [[nodiscard]] virtual auto build() const -> std::string { return m_json.dump(2); }
```
gustingonzalez: I think `build` sounds as something that becomes immutable (or finalized) once it is "built". But in this case, I can build it and then continue adding elements to the structure. Maybe a better name could be `str`/`to_str`?

As a follow-up, if we make this change, we could also consider renaming `pisa::stats_builder()` by `pisa::json_stats`.

src/util/stats_builder.cpp
```diff
@@ -0,0 +1,81 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "pisa/util/stats_builder.hpp"
+#include "nlohmann/json.hpp"
+
+namespace pisa {
+
+class JsonStatsBuilder: public ::pisa::detail::StatsBuilderInterface {
```
gustingonzalez: I think `JsonStatsBuilder` sounds verbose (very Java-style). Maybe `JsonStats` could be enough? That would make sense, since we could expose a `str` method instead of having a `build` method (comment in line 31).

src/util/stats_builder.cpp
elshize: Thanks, I think this makes sense, I'll make the suggested changes once I find a moment.

gustingonzalez:
@elshize, looks good to me!

PR #633, opened 2026-03-21
Fix query algorithm WAND validation and duplicate TSV header in `queries`

gustingonzalez:
Key changes in this pull request: 
1. Fix `ranked_and`, `ranked_or`, `ranked_or_taat`, `ranked_or_taat_lazy` incorrectly marked as not requiring WAND data in `VALID_ALGORITHMS`. This did not cause issues because the check was correctly enforced at runtime.
2. Centralize WAND data validation using `VALID_ALGORITHMS` instead of duplicating check in `queries`, reducing the need for double checks when modifying the logic. 
3. Fix duplicate TSV header when running `queries` using multiple algorithms (previously printed for each algorithm).

### DIFF
--- a/tools/app.cpp
+++ b/tools/app.cpp
@@ -85,12 +85,12 @@ const std::map<std::string, bool> Algorithm::VALID_ALGORITHMS = {
     {"wand", true},
     {"block_max_wand", true},
     {"block_max_maxscore", true},
-    {"ranked_and", false},
+    {"ranked_and", true},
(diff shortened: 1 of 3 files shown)


elshize (approved):
Nice 🚀 I'll merge this. FYI I filed a follow-up https://github.com/pisa-engine/pisa/issues/634 to improve `VALID_ALGORITHMS` mapping, you can have a look if you're interested.

PR #636, opened 2026-05-06
CI: job to build docs from HEAD and compare to `docs/book`

### LINKED ISSUE #627: Check docs generation in CI
Documentation needs to be rebuilt to take effect (see https://github.com/pisa-engine/pisa/tree/main/docs) and it's easy to miss.

We could add a CI step to verify that docs are up to date.

Steps could be:
1. Generate docs to a temp directory.
2. Compare if the files in that directory are the same as in our generated directory.

Caveat: would this only fail when docs are modified? I think so, as long as we consistently use the same version of mdbook in CI. It may be a good idea to modify the instructions to install a pinned version locally, so that it matches what we do in CI.

lgrz:
Fixes #627 -- Check docs generation in CI.

- Add docs job to commit.yml workflow which reuses the binaries built from the `test_cli` job
- Build `docs/book`


elshize (approved):

elshize:
Thanks @lgrz!

PR #641, opened 2026-09-07
Use enum for algorithm WAND requirements

### LINKED ISSUE #634: Use enum for valid algorithms
Instead of using booleans to say whether an algorithm needs wand data, use a descriptive enum:

```cpp
enum WandDataRequired { YES, NO };
```

Or a variation on that.

This makes it easier to understand what the mapping means without needing an explanation in a comment.

appleweiping:
Fixes #634.

- add the scoped `WandDataRequired` enum with `Yes` and `No` values
- replace boolean entries and checks without changing runtime behavior
- update the CLI mapping and validation tests

Validation:

- clang-format 20.1.6 passed
- git diff check passed
- the complete build and CLI suite will run in CI because CMake is not available in my local environment

Developed with OpenAI Codex assistance. I reviewed the change and remain responsible for it.


appleweiping:
The Commit and Pull Request workflows are both held at action_required before any jobs start because the branch comes from a fork. Codacy passed. I could not run the CMake suite locally, so approving the workflows is the remaining verification step.

elshize:
> the complete build and CLI suite will run in CI because CMake is not available in my local environment

FYI it's possible -- in theory -- to run these tests using docker, similar to how the CI is set up. But I'm also seeing that the clang docker seems to be broken, which is also tripping the CI.

I see that all tests in the coverage job passed, so I'll merge it. I'll try to find some time later in the week to fix the cert issues.

PR #643, opened 2026-09-13
Fix CI-build, dockerfiles

lgrz:
Recent builds from PR submissions are broken. These changes should enable builds to run again.

The gcc images have moved on from bullseye and for clang, `install_clang.sh` is already tied to apt sources for bookworm. Note the gcc images for version 14 onwards are on trixie [1].

This is a minimum fix to get builds working. Future work/discussion should consider moving to trixie and more recent gcc/clang images. This would also be a good time to consolidate the build pipeline.

```
[1]: https://hub.docker.com/_/gcc
```


lgrz:
Ah ok, I see now that what the cert problem from discussion #641 was. The logs below are from my own build of `trunk.yml` workflow. Will look into it and update this PR

```
#8 7.759 W: http://apt.llvm.org/bookworm/dists/llvm-toolchain-bookworm-19/InRelease: No system certificates available. Try installing ca-certificates.
#8 7.759 W: Failed to fetch http://apt.llvm.org/bookworm/dists/llvm-toolchain-bookworm/InRelease  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 199.232.154.49 443]
#8 7.759 W: Failed to fetch http://apt.llvm.org/bookworm/dists/llvm-toolchain-bookworm-19/InRelease  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 199.232.154.49 443]
```

lgrz:
A few more issues have been fixed which did not show up until running on this repo.

- Remote build cannot download scripts and execute them
- Remote build refuses to download gpg keys
- Update actions yaml files to use the same version of the checkout action

I think this should be ready for review now. Thanks

/cc @elshize

amallia (approved):

amallia:
Thanks Luke. Could you fix the warnings so we can merge this PR?

lgrz:
Thanks Antonio, the static analysis errors should be fixed now.

amallia (approved):

lgrz:
Sorry, I spoke too soon. The suggestions from static analysis do not work for `apt`. It seems we actually do want word splitting in that case. For example (using the `show` subcommand to show the behavior),

No quote:
```
$ apt show lsb-release software-properties-common && echo $?
Package: lsb-release
... [snip]
Package: software-properties-common
... [snip]

0
```

With quote:
```
$ apt show "lsb-release software-properties-common"; echo $?
N: Unable to locate package lsb-release software-properties-common
N: Unable to locate package lsb-release software-properties-common
E: No packages found

100
```

lgrz:
Have reverted the `apt` line in llvm.sh. Not sure if we should roll back to the original ./install-clang.sh script and fix the gpg key, etc from there?

.github/workflows/trunk.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        compiler: [gcc10, gcc11, gcc12, gcc13, gcc14, clang16, clang17, clang18, clang19]
+        compiler: [gcc12, gcc13, gcc14, clang17, clang18, clang19]
```
elshize, line 12: Not necessarily to fix in the same PR, but should we move on to new versions?

I think we could just do the last 3 versions for each. So say, gcc 14/15/16 and clang 20/21/22?

elshize:
There's something off with Codacy, got stuck on this commit. Also, generally, codacy tends to be very nitpicky so if we ever disagree with it, we can dismiss such warnings.

The PR looks good to merge for me. Unless either of you feel like we need to wait for Codacy, we can bypass the checks and merge it.

.github/workflows/trunk.yml
lgrz: Sounds good, will create a new issue/PR for this one.

lgrz:
> The PR looks good to merge for me. Unless either of you feel like we need to wait for Codacy, we can bypass the checks and merge it.

Thanks, I am happy with the PR as is.

lgrz:
That last update has no changes; it was to trigger codacy again as it did not finish last time.

elshize:
Thanks @lgrz. I suppressed Codacy analysis from `llvm.sh` in the future.

lgrz:
Good idea, thanks @elshize

